PR #8, opened 2020-08-23
Set up frontend structure, ESLint, and styled components

alankbi:
Proposed frontend structure: 

`api`: all the fetch/rest calls to the backend
`components`: all reusable UI pieces, such as headers, buttons, code editors, forms, or profile cards
`views`: all pages, such as landing, join, game, and results


frontend/src/App.tsx
```diff
@@ -1,36 +1,93 @@
...
-import './App.css';
- 
-class App extends Component {
- 
-    state:any = {};
- 
-    componentDidMount() {
-        setInterval(this.hello, 250);
+import LandingPage from './views/Landing';
+import fetchHello from './api/hello';
```
zpChris, line 5: In my internship, we ordered imports alphabetically -- is that something we should do? That might pertain more to Java anyway, but I wanted to bring this up anyway.

frontend/src/App.tsx
```diff
@@ -1,36 +1,93 @@
...
+
+const StyledApp = styled.div`
```
zpChris, line 7: Not a suggested change, just a note: If @alankbi or @helenli8181 haven't done so yet, if either of you are using Visual Studio Code, I recommend adding the [vscode-styled-components](https://marketplace.visualstudio.com/items?itemName=jpoissonnier.vscode-styled-components) extension to help with syntax highlighting -- this makes the code much more readable.

frontend/src/api/hello.ts
```diff
@@ -0,0 +1,4 @@
+const fetchHello = (): Promise<string> => fetch('/api/v1/hello')
+  .then((response) => response.text());
```
zpChris, line 2: The `fetch('/api/v1/hello')` returns a `SampleObject`, so `response` should be turned into `response.json()` (not `response.text()`. This also means the `Promise` should hold a different type, other than a **string**.

(Some similar changes will need to be made on the frontend file, `App.tsx`, as well.)

frontend/src/App.tsx
```diff
@@ -1,36 +1,93 @@
...
+
+  hello = () => {
+    fetchHello()
+      .then((message) => {
+        this.setState({ message });
+      });
+  };
+
+  render() {
+    const { message } = this.state;
```
zpChris, line 72: Why is `message` surrounded by brackets?

frontend/src/App.tsx
```diff
@@ -1,36 +1,93 @@
...
+  }
+
+  componentDidMount() {
+    setInterval(this.hello, 250);
+  }
+
+  hello = () => {
+    fetchHello()
+      .then((message) => {
+        this.setState({ message });
```
zpChris, line 67: Right now, in the current setup, this actually holds a message within a message (see screenshot attachment). 
Screen Shot 2020-08-23 at 5 41 07 PM

This should be taking the actual message from the SampleObject JSON object, and displaying only the message to the frontend. (See the second screenshot below.)
Screen Shot 2020-08-23 at 5 43 12 PM

frontend/src/App.tsx
```diff
@@ -1,36 +1,93 @@
...
+
+  constructor(props: MyProps) {
+    super(props);
+
+    this.state = { message: 'Loading' };
+  }
+
+  componentDidMount() {
+    setInterval(this.hello, 250);
+  }
```
zpChris, line 62: nit: Set `this.state` with the 'Loading' message as a class field, and replace `this.state = { message: 'Loading' };` with `this.hello()`.

frontend/src/components/navigation/Header/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+function Header() {
```
zpChris, line 3: How would this be incorporated into the actual app? (If you have any tutorial / link to refer to, that'd work.)

frontend/src/views/Landing/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import Header from '../../components/navigation/Header';
+
+const Content = styled.div`
+  width: 80%;
+  margin: 10%;
+  text-align: center;
+  font-size: large;
+`;
+
+function LandingPage() {
```
zpChris, line 12: Is this used anywhere? If not, how do we incorporate it?

src/main/java/com/rocketden/main/controller/v1/BaseRestController.java
```diff
@@ -0,0 +1,8 @@
+package com.rocketden.main.controller.v1;
+
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RequestMapping("/api/v1")
+public class BaseRestController {
+    // empty
```
zpChris, line 7: nit: Remove comment, or edit it to explain how the **BaseRestController** works.

src/main/java/com/rocketden/main/controller/v1/SampleController.java
```diff
@@ -1,12 +1,12 @@
...
+package com.rocketden.main.controller.v1;
 
 import com.rocketden.main.model.SampleObject;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
-public class SampleController {
+public class SampleController extends BaseRestController {
```
zpChris, line 9: Ah, I see how this works -- that's clever! Nice job 👍

frontend/src/index.tsx
```diff
@@ -1,14 +1,31 @@
...
+  code {
+    font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
+      monospace;
+  }
+`;
+
 ReactDOM.render(
   <React.StrictMode>
+    <GlobalStyle />
     <App />
```
zpChris, line 26: I know this works correctly, I just have a question on how this works: Should you be editing the **App** style rather than adding a new HTML element with the appropriate font settings?

frontend/src/index.tsx
```diff
@@ -1,14 +1,31 @@
...
   </React.StrictMode>,
-  document.getElementById('root')
+  document.getElementById('root'),
```
zpChris, line 28: nit: Remove comment. (It's the last element of the list.)

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="Eslint" enabled="true" level="WARNING" enabled_by_default="true" />
+  </profile>
+</component>
```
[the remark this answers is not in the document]
alankbi: Tried adding a new line but apparently it was reset before committing, I think since this is an auto-generated IDE settings file we might not be able to really modify it.

frontend/src/App.tsx
alankbi: Fixed, although I have two notes/questions:

1. ESLint recommends that you have package imports first (e.g. 'react') and then relative imports (e.g. './logo.svg'), so I left them separate but sorted each group. I think can modify ESLint rules though if that's something you think should change.
2. When sorting alphabetically, do you prefer sorting by the thing you're importing (e.g. LandingPage) or by the file it's coming from (e.g. './api/hello')? In this scenario both lead to the same result so it doesn't matter, but in future situations it could be good to clarify (my personal preference is by file but don't feel too strongly about this).

frontend/src/App.tsx
alankbi: Similarly, there's a styled components plugin for IntelliJ too. And while we're on the topic of IDE settings, I added ESLint in this PR but I think you'll still need to enable linting in VSCode settings.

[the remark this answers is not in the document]
alankbi: I think you're right about App being a component so I moved that to the components folder. As for StyledApp, since it's applying styling to the App component I kept it in the same file kind of like how Spectrum does it [here](https://github.com/withspectrum/spectrum/blob/alpha/src/components/card/index.js). Although it'd also make sense to put it in a separate file like they do [here](https://github.com/withspectrum/spectrum/tree/alpha/src/components/button). Lmk if you have a preference on that.

frontend/src/App.tsx
```diff
@@ -1,36 +1,93 @@
...
-    componentDidMount() {
-        setInterval(this.hello, 250);
+import LandingPage from './views/Landing';
+import fetchHello from './api/hello';
+
+const StyledApp = styled.div`
+  text-align: center;
+  
+  .App-logo {
+    height: 40vmin;
```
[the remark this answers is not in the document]
alankbi: Yeah that's a good point, I added a card in the backlog to discuss responsive styling. I left this as is though cause we're eventually gonna remove it anyways (I just copied over the CSS from App.css before deleting it).

frontend/src/api/hello.ts
alankbi: Thanks for catching that! This should be fixed now.

frontend/src/App.tsx
alankbi: It's called destructuring according to [this site](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), basically it'd be equivalent to `const message = this.state.message`. This is also an ESLint suggestion.

frontend/src/App.tsx
alankbi: Good catch, fixed.

frontend/src/App.tsx
alankbi: I might want to push back on this change for the following two reasons:

1. ESLint recommends initializing state in the constructor instead of as class fields, and similarly, based on a lot of online examples, it seems like the convention is to initialize state in the constructor (although to be fair I think both are equivalent). 
2. In general, it's recommended that API calls are put in the `componentDidMount` function rather than the constructor ([see here](https://stackoverflow.com/questions/44811983/calling-action-from-constructor-vs-a-life-cycle-method/44812019)). If the concern was making sure 'Loading' doesn't stay shown for so long, I added an initial call to `this.hello` in `componentDidMount` to hopefully mitigate that.

Definitely open to discussion on this though if you feel strongly the other way.

frontend/src/components/navigation/Header/index.tsx
alankbi: Right now it's incorporated into LandingPage (left more details under that comment).

frontend/src/views/Landing/index.tsx
alankbi: In the future, I'd imagine it being incorporated using React Router (see [this example](https://github.com/withspectrum/spectrum/blob/alpha/admin/src/routes.js#L29)) which basically will allow us to have routes such as /landing, /login, /game, /results, etc. Right now we don't have that though, so I just included it at the bottom of the App component (you should see it if you scroll down on localhost:8080).

src/main/java/com/rocketden/main/controller/v1/BaseRestController.java
alankbi: Added a brief comment describing the class.

frontend/src/index.tsx
alankbi: These are global styles (i.e. things applied on tags such as 'body' or 'html') meaning I don't think it will work if you apply the style on App (a div). The current code comes from [this post](https://stackoverflow.com/questions/46760861/styled-components-how-to-set-styles-on-html-or-body-tag).

frontend/src/index.tsx
alankbi: That's another ESLint style recommendation to include trailing commas, though it can be manually disabled. How strong is your preference against them?

zpChris: Ah I didn't realize that was the ESLint style recommendation -- I have no problem with this then. Thanks for checking that 👍

frontend/src/App.tsx
zpChris: 1. That sounds good, thanks for the clarification there.
2. I don't have any personal preference, so let's go by file moving forward.

frontend/src/App.tsx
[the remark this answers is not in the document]
zpChris: I think maintaining the same structure Spectrum has makes sense, so your change is good IMO.

frontend/src/App.tsx
zpChris: Ah gotcha, thanks for the explanation.

frontend/src/views/Landing/index.tsx
zpChris: Thanks for including that -- looks good.

frontend/src/App.tsx
zpChris: That makes perfect sense, thanks for the response -- sounds great by me.

zpChris:
I turned off the "Ensure there are no sensitive filenames (ex: `id_rsa`, `.env`)" rule on GitGuardian and re-ran the security checks, so it's now passed. I don't think this is a long-term solution, but I don't think `application.properties` should've been flagged in this case anyway as nothing was leaked there, so this should be OK to merge.

alankbi:
> I turned off the "Ensure there are no sensitive filenames (ex: `id_rsa`, `.env`)" rule on GitGuardian and re-ran the security checks, so it's now passed. I don't think this is a long-term solution, but I don't think `application.properties` should've been flagged in this case anyway as nothing was leaked there, so this should be OK to merge.

Thanks, I'll merge. I agree we should re-enable that rule at some point while maybe just disabling the flag specifically for application.properties (I saw an option for that when I was messing around with GitGuardian).

PR #9, opened 2020-08-25
Add ClearDB MySQL database, create basic DAO setup, add tests, and clean up old files.

zpChris:
### List of changes:
- Add the ClearDB MySQL database.
  - Use the GitHub secret to encrypt the ClearDB database password.
- Create the basic DAO (Data Access Object) setup to store, update, and retrieve programming problems in the database.
  - See this [StackOverflow post](https://stackoverflow.com/questions/35078383/what-are-the-dao-dto-and-service-layers-in-spring-framework) for more information.
- Add testing for the GET and POST requests that retrieve and add problems.
  - The h2 database is used for local testing -- see this [Baeldung article](https://www.baeldung.com/spring-testing-separate-data-source) for more information.
- Clean up old files and filenames (however, I didn't remove them completely). 

#### How to Allow the Project to Run Locally
In order to run the project and tests locally, just follow one step:
1. Run the command `export JASYPT_ENCRYPTOR_PASSWORD=XXXX`, with `XXXX` representing our password.

It may also be useful to login to Heroku, and that way you can view all of the project details:
1. Navigate to your project through Terminal, and run `heroku login` to log in to Heroku through your project. (See login details on the Google Drive.)
2. Confirm you are set up by checking you get the same results when running the terminal commands in the image (in no particular order). 
GitHub Commands to Check ClearDB

#### How to View the Database
If you use a Mac, follow these steps:
1. Download [Sequel Pro](https://sequelpro.com/).
2. Connect to the database using the ClearDB credentials (see image, then add the password).
    1. You may have to start the localhost site with the `./mvnw spring-boot:run` command before connecting to the database.
3. Choose the database on the top-left after you connect, then you can view the database contents with the "Content" tab. You're should be all connected to the local database now.
Sequel Pro Login Details

### Notes:
- I used part of the instructions from both the [ClearDB Setup Tutorial](https://devcenter.heroku.com/articles/cleardb) and the [Spring MySQL Tutorial](https://spring.io/guides/gs/accessing-data-mysql/) to set up access to the database.
- I used the instructions from this [StackOverflow post](https://stackoverflow.com/questions/37404703/spring-boot-how-to-hide-passwords-in-properties-file) to encrypt the database password. Previous versions of the password are on GitHub, but I have since reset the password. I used this [online tool](https://www.devglan.com/online-tools/jasypt-online-encryption-decryption) to retrieve the encrypted password (using the "Two Way Encryption (With Secret Text)" option).
- I used `@DirtiesContext` to re-create the database after each test method; see this [StackOverflow post](https://stackoverflow.com/questions/34617152/how-to-re-create-database-before-each-test-in-spring) for more information.
- I removed the `containsString(...)` method from the test methods, and instead verified that the string matches exactly through the `content().string(...)` method.
- I added `JASYPT_ENCRYPTOR_PASSWORD: ${{secrets.JASYPT_ENCRYPTOR_PASSWORD}}` as an environment variable to the `workflow.yml` file to de-encrypt the database password in the `application.properties` file (non-testing version).
- There are lots of additions / deletions to the `pom.xml` file that may seem superfluous -- it was the only way I could get the database to work, so I've kept them in the CL.
- I'm not sure how the database will interact remotely -- if it's separate from the local database or not. As of now, I had planned to simply merge and deploy to Heroku, and see what happens from there.

### Screencast and Images:
There are no screencasts or images in this CL because there are no changes to the UI, just setting up the database and backend.


src/main/java/com/rocketden/main/controller/v1/ProblemController.java
```diff
@@ -0,0 +1,36 @@
...
+
+@RestController
+public class ProblemController extends BaseRestController {
+
+	@Autowired
+	private ProblemRepository problemRepository;
+  
+  @PostMapping("/addProblem")
+  public @ResponseBody Problem addNewProblem (@RequestParam String name, 
+    @RequestParam String description) {
```
alankbi, line 21: This might be a nit, but is it possible to change these from RequestParam to RequestBody annotations? This gets a bit deep into REST API design where you have things like query variables, path variables, and body variables (these two sites [[1]](https://stackoverflow.com/questions/14551194/how-are-parameters-sent-in-an-http-post-request) [[2]](https://softwareengineering.stackexchange.com/questions/332637/rest-webservices-post-put-which-type-of-parameter-should-i-choose-and-whe) give an overview of the differences between them and [this](https://stackoverflow.com/questions/28039709/what-is-difference-between-requestbody-and-requestparam) details how to use them with Spring), but the gist from what I've read is that GET tends to use query params whereas POST tends to use body params. However, this change would affect a lot of the tests so we can always revisit this later (esp since you already spent a ton of time on this PR lol).

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class ProblemController extends BaseRestController {
+
+	@Autowired
+	private ProblemRepository problemRepository;
+  
+  @PostMapping("/addProblem")
```
alankbi, line 19: This also may be a nit but could we change the URL paths to follow the REST patterns listed in [[1]](https://nordicapis.com/10-best-practices-for-naming-api-endpoints/) and [[2]](https://restfulapi.net/resource-naming/)? I'm not sure what this would look like exactly for our scenario, maybe something like /problems, but we can discuss this in a call.

alankbi:
Beautiful changelist ❤️ 

Leaving some general questions here:

1. Could you copy the part about setting up the export command locally to the README?
2. Do we need to add that variable to Heroku or is it already taken care of by adding it to GitHub secrets?
3. I get the following warning when running GET and POST requests locally:

``` 
HikariPool-1 - Failed to validate connection com.mysql.cj.jdbc.ConnectionImpl@15badc4b (No operations allowed after connection closed.). Possibly consider using a shorter maxLifetime value.
```

Any ideas? On a related note, the requests each take about a second which seems long; idk if this has anything to do with that or not.

src/main/resources/application.properties
```diff
@@ -0,0 +1,4 @@
+spring.jpa.hibernate.ddl-auto=update
+spring.datasource.url=jdbc:mysql://us-cdbr-east-02.cleardb.com/heroku_bf61bac08e7d5ce
+spring.datasource.username=be954c4707bc32
+spring.datasource.password=ENC(epj2Y3gphq6164k15apQphtgq8aj12qZ)
```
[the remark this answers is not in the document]
zpChris: Thank you very much indeed 🙂

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
```diff
@@ -0,0 +1,36 @@
+package com.rocketden.main.controller.v1;
```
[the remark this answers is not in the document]
zpChris: What indentation? I don't see any here?

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class ProblemController extends BaseRestController {
+
+	@Autowired
```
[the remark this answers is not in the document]
zpChris: That makes sense, thanks for pointing that out -- done!

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
zpChris: Absolutely, that makes sense -- changed to `"/problems"` for both the GET and POST requests.

src/main/java/com/rocketden/main/dto/problem/Problem.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Entity // This tells Hibernate to make a table out of this class
+@Getter
+@Setter
+public class Problem {
+	@Id
+	@GeneratedValue(strategy=GenerationType.AUTO)
+	private Integer id;
+
+  private String name;
```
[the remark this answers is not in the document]
zpChris: Done!

src/main/java/com/rocketden/main/dto/problem/Problem.java
```diff
@@ -0,0 +1,21 @@
+package com.rocketden.main.dto.problem;
```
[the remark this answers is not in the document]
zpChris: Sure, I can do that -- thanks for the clarification.

src/test/java/com/rocketden/main/ProblemTests.java
```diff
@@ -0,0 +1,87 @@
...
+				.andExpect(content().string(containsString(getProblemsResult)));
+	}
+
+	@Test
+	public void restCallGetProblemsEmptyList() throws Exception {
+		assertThat(this.restTemplate.getForObject("http://localhost:" + port + "/api/v1/getProblems",
+				String.class)).contains("[]");
+  }
+
+}
```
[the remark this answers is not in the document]
zpChris: To the best I can confirm, yes -- the `@DirtiesContext(...)` tag should take care of that. See this [StackOverflow post](https://stackoverflow.com/questions/34617152/how-to-re-create-database-before-each-test-in-spring) for more information.

src/test/java/com/rocketden/main/ProblemTests.java
```diff
@@ -0,0 +1,87 @@
...
+	private TestRestTemplate restTemplate;
+
+	@LocalServerPort
+  private int port;
+  
+  // Constants for common POST requests.
+  private final String postAddProblemRequestOne = "/api/v1/addProblem?name=" +
+    "Sort an Array&description=Sort an array from lowest to highest value.";
+  private final String postAddProblemRequestTwo = "/api/v1/addProblem?name=" +
+    "Find Maximum&description=Find the maximum value in an array.";
```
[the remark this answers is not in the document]
zpChris: I think that was copy/pasting messing this up -- sorry about that. It should be fixed now, thanks for pointing this out.

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
```diff
@@ -0,0 +1,36 @@
...
+    // Add the problem to the database, and return the problem JSON.
+    Problem problem = new Problem();
+    problem.setName(name);
+		problem.setDescription(description);
+    problemRepository.save(problem);
+		return problem;
+	}
+
+	@GetMapping("/getProblems")
+	public @ResponseBody Iterable<Problem> getAllProblems() {
```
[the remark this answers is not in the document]
zpChris: That makes sense, thanks for linking the article. Done!

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
zpChris: This is a good point. It was a lot of changes, but I thought it was important so I replaced the two `@RequestParam` parameters with a single `@RequestBody` variable. I had to add a function in `util/Utility.java` that converted the `x-www-form-urlencoded` body format to a `Map<String, String>` mapping the keys to values, but I managed to do so. I also had to reformat the testing as you mentioned, and putting the body in a String turned out to be the easiest way to do this, but I got this to work successfully as well.

zpChris: Note: There may be other `replace(...)` things that we need to include later; right now I only have the following two:
```
pair[1].replace("+", " ");
pair[1].replace("%0A", "\n");
```
However, I think we can add them as they come up. I thought there would be a more elegant solution to convert the body into a proper format, but this is the only attempt that worked.

alankbi: I feel bad cause I think one of the links I shared did mention `x-www-form-urlencoded` which might not be best suited for our use case; there should be an easier way where you accept JSON by just putting `@ResponseBody Problem` as the parameter instead of String, similar to the POST example in this tutorial: https://spring.io/guides/tutorials/bookmarks/

That's definitely my bad and we can deal with the details of it later, don't want to keep burdening you on this PR haha

zpChris:
> Beautiful changelist ❤️
> 
> Leaving some general questions here:
> 
> 1. Could you copy the part about setting up the export command locally to the README?
> 2. Do we need to add that variable to Heroku or is it already taken care of by adding it to GitHub secrets?
> 3. I get the following warning when running GET and POST requests locally:
> 
> ```
> HikariPool-1 - Failed to validate connection com.mysql.cj.jdbc.ConnectionImpl@15badc4b (No operations allowed after connection closed.). Possibly consider using a shorter maxLifetime value.
> ```
> 
> Any ideas? On a related note, the requests each take about a second which seems long; idk if this has anything to do with that or not.

Thank you! I'll address the questions here:
1. Yes -- done! Thanks for this note.
2. That variable _should_ already be taken care of through GitHub secrets. I don't know for sure what the deploy functionality will look like, but if something goes wrong I can try to add it.
3. Fixed! This was quite a pain to be honest, but I added HikariCP config with some new dependencies to Maven and set the `maxLifetime` value less than `connectionTimeout` and `idleTimeout` in the `application.properties` file. I also added a test to `ProblemTests.java` to ensure that the Hikari Connection Pool (HikariCP) was configured. Some relevant links include (1)[https://www.baeldung.com/spring-boot-hikari] (2)[https://stackoverflow.com/questions/59894368/spring-boot-application-is-ignoring-hikaricp-config] (3)[https://stackoverflow.com/questions/60310858/possibly-consider-using-a-shorter-maxlifetime-value-hikari-connection-pool-spr].

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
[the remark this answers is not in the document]
alankbi: My bad, was referring to later down, but I think it comes down to an issue of tabs vs. 2 spaces vs. 4 spaces which we can visit some other time.

src/main/java/com/rocketden/main/Main.java
```diff
@@ -1,9 +1,12 @@
 package com.rocketden.main;
 
+import com.ulisesbocchio.jasyptspringboot.annotation.EnableEncryptableProperties;
+
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@SpringBootApplication
+@SpringBootApplication()
```
alankbi, line 8: Out of curiosity, why the parentheses? Not an issue.

src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -0,0 +1,33 @@
+package com.rocketden.main.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Utility {
+
+  /**
+   * Convert the Request Body in 'x-www-form-urlencoded' format to a HashMap
```
alankbi, line 9: In the future this can be removed if we decide to change request body format to JSON, in which we can use DTO objects instead of Strings. But I think that's beyond the scope of this PR :)

src/main/resources/application.properties
```diff
@@ -0,0 +1,8 @@
...
+spring.jpa.open-in-view=true
+spring.datasource.hikari.connectionTimeout=45000
```
alankbi, line 6: Could you explain what these values are and what they do? Nice work on figuring that out btw :)

src/main/java/com/rocketden/main/Main.java
zpChris: Sorry, this was leftover from other changes I was trying to make -- removed 👍

src/main/java/com/rocketden/main/util/Utility.java
zpChris: Yes I think you're likely right  -- I'll look into this, and if `application/json` seems better, I'll implement that in a separate CL.

src/main/resources/application.properties
zpChris: I added comments to the `application.properties` file to explain each of these values.

The first four lines (copied below) set up the database connection.
```
spring.jpa.hibernate.ddl-auto=update
spring.datasource.url=jdbc:mysql://us-cdbr-east-02.cleardb.com/heroku_bf61bac08e7d5ce
spring.datasource.username=be954c4707bc32
spring.datasource.password=ENC(epj2Y3gphq6164k15apQphtgq8aj12qZ)
```
The next line (copied below) enables Open Session in View. You can see more details in this [StackOverflow post](https://stackoverflow.com/questions/30549489/what-is-this-spring-jpa-open-in-view-true-property-in-spring-boot); OSIV is enabled by default, and a warning is provided when a value is not set. The GitHub response actually says to set this to `false`, but I think extra work needs to be done if that's the case, and since `true` is the default, I wanted to set it to that at first. See more info: [[1]](https://www.yawintutor.com/warn-3840-spring-jpa-open-in-view-is-enabled-default/).
```
spring.jpa.open-in-view=true
```
The final lines (copied below) set configuration parameters for HikariCP; read more [here](https://www.baeldung.com/spring-boot-hikari#hikari-configuration). The `maxLifetime` value has to be less than `idleTimeout` (and, if I understand correctly, `connectionTimeout` as well) as explained in this [StackOverflow post](https://stackoverflow.com/questions/46830159/hikari-failed-to-validate-connection-because-connection-is-closed).
```
spring.datasource.hikari.connectionTimeout=45000 
spring.datasource.hikari.idleTimeout=600000 
spring.datasource.hikari.maxLifetime=30000
```

src/main/java/com/rocketden/main/controller/v1/ProblemController.java
zpChris: Ah, I tried doing `@ResponseBody Problem`, but it didn't work -- possibly because I didn't have `application/json`. I'll try this on another branch.

alankbi: It might have to do with the way you send in the POST request using Postman, I think under Body you have to select raw -> JSON in order to match the format on the client end.

PR #12, opened 2020-08-28
Change "/problems" POST request to "application/json" format, and update relevant tests.

zpChris:
### List of Changes:
- Change the `"/problems"` POST request format setting from `application/x-www-form-urlencoded` to `application/json`.
  - This was decided after @alankbi brought this up in [this discussion](https://github.com/rocketden/main/pull/9#discussion_r478665757) or PR #9, and since I had a method and was specially converting the format to JSON, it would be more efficient to accept JSON from the start.
- Add **ProblemDTO** in order to directly accept the JSON from the POST request. See [[1]](https://rules.sonarsource.com/java/tag/spring/RSPEC-4684) for an explanation of why this was necessary, and for a basis of how I formatted the `addNewProblem(...)` method.
- Fix the testing to reflect the updated format of accepting input in the `"/problems"` POST request.
- Removed the `Utility.java` class in the `main` folder as I do not need to convert from `application/x-www-form-urlencoded` to JSON anymore.
- Added the `Utility.java` class in the `test` folder to add the `convertObjectToJsonString(...)` method; this was used in creating the test inputs for the `"/problems"` POST request.

### Notes:
- If you didn't see it above, you should look at [this page from SonarSource](https://rules.sonarsource.com/java/tag/spring/RSPEC-4684); this details how I set up the Problem DTO and `addNewProblems(...)` POST method.
- I added a constructor in the `ProblemTests()` class to set up the JSON input for the `"/problems"` POST request.
  - I also added a `Utility.java` class in the `test` folder that uses [Gson](https://github.com/google/gson) to convert Java objects to JSON strings. I then converted the `ProblemDto` object to a JSON string through this method to build the JSON input. While I could've written the JSON string directly, I felt this was the cleanest procedure.

### Screencast and Images:
There are no screencasts or images in this CL because there are no changes to the UI, just setting up the database and backend.


src/test/java/com/rocketden/main/ProblemTests.java
```diff
@@ -51,6 +53,26 @@
...
+    sortArrayProblem.setName("Sort an Array");
+    sortArrayProblem.setDescription
+      ("Sort an array from lowest to highest value.");
+    sortArrayProblemJson = Utility.convertObjectToJsonString(sortArrayProblem);
+
+    // Create "Find Maximum" problem JSON string.
+    findMaxProblem = new ProblemDto();
+    findMaxProblem.setName("Find Maximum");
+    findMaxProblem.setDescription
+      ("Find the maximum value in an array.");
```
alankbi, line 72: Is there a reason why the parentheses are on a new line, like to do with max line length? If so, I get that, though imo it does look a bit funky.

zpChris: Yes, it's for max line length (I try to keep it less than 80 characters). I agree it looks funky, it doesn't bother me too much but if you have a fix / suggestion we can change this.

src/main/java/com/rocketden/main/dto/problem/ProblemDto.java
```diff
@@ -0,0 +1,11 @@
+package com.rocketden.main.dto.problem;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class ProblemDto {
+  private String name;
+  private String description;
```
helenli8181, line 10: Is this where we will add the fields indicating problem difficulty and type too?

zpChris: Yeah that's right! Any of the problem details; name, description, example I/O, and maybe test cases (or at least their IDs) would go here. I kept it simple for now because we aren't going to be using all that for a while. 👍

PR #16, opened 2020-09-01
Add base for constructing sockets, enable connection and disconnection from frontend.

zpChris:
### List of Changes:
- Add the fundamental code for constructing sockets, and set up a predefined endpoint at `"/join-room-endpoint"`.
  - Eventually, these endpoints will be dynamically populated after the room code is created or inputted.
- Add the necessary dependencies in `pom.xml`. 
  - I had to put these at the top, and may have re-arranged some of the dependencies in that file, as it was throwing errors before. I also added comments describing the different dependencies, as the list was getting very large and difficult to understand.
  - Some of the added dependencies may not be needed for _this_ PR, but they will be needed for adding all future socket development and the next few PRs.
- Add a button with an `onClick` component that triggers a function and add a `socket.js` file that holds the `connect(endpoint)` and `disconnect()` functions.

### Notes:
- I am largely moving the code from the [Spring WebSocket Tutorial](https://spring.io/guides/gs/messaging-stomp-websocket/) into this repository, but changing it slightly to fit our codebase.
- The "Connect" and "Disconnect" buttons are just for testing; eventually, the user will not have to click these in order to connect or disconnect to the socket.
- I placed the connection / disconnection functions for the frontend in a `Utility.js` file; I'm open to moving this directly into the "Join Room" file, or elsewhere.
- I am not adding testing yet, as there are no messages being sent over the sockets, which is first what is tested with sockets.
- The console output after clicking "Connect" then "Disconnect" should match what's shown below. You can click "Disconnect" and get a console output even if you aren't connected; same with "Connect", you can get the same console output multiple times in a row.
  - Ignore the 404 Error, that's another issue: [Link to Trello Bug Task](https://trello.com/c/CdXOmN1q/88-404-errors-on-non-homepage-refresh).
Console Output for Socket Connection
- It is OK that the server says it is `undefined` in the console output; as described in [this GitHub Discussion](https://github.com/stomp-js/stompjs/issues/4), that will not affect anything in the app.

### Screencast and Images:
Join Room Page


frontend/src/util/Utility.ts
```diff
@@ -0,0 +1,19 @@
+import SockJS from "sockjs-client";
+import Stomp from "stompjs";
+
+let stompClient : any = null;
+
+export const Connect = (endpoint : any) => {
```
helenli8181, line 6: Are endpoints strings?

zpChris: Yes; this has since been updated!

frontend/src/views/Join.tsx
```diff
@@ -1,9 +1,16 @@
 import React from 'react';
+import { SocketButtonConnection } from '../components/core/Link';
+import { Connect, Disconnect } from '../util/Utility';
 
 function JoinGamePage() {
   return (
     <div>
-      Join game page
+      <SocketButtonConnection onClick={() => Connect('/join-room-endpoint')}>
```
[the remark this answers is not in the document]
zpChris: Maybe try again now? I'm not sure how you're seeing a Heroku issue if you're testing it locally -- can you try this again, and if it still shows this, maybe we can call and screenshare?

frontend/package.json
```diff
@@ -12,17 +13,20 @@
...
     "react-dom": "^16.13.1",
     "react-router-dom": "^5.2.0",
     "react-scripts": "3.4.3",
+    "stompjs": "^2.3.3",
     "styled-components": "^5.1.1",
     "typeface-roboto": "0.0.75",
     "typescript": "~3.7.2"
   },
   "devDependencies": {
+    "@types/sockjs-client": "^1.1.1",
```
alankbi, line 29: Could we move this to dependencies to be consistent with the other `@types` dependencies? Or move all of them to dev dependencies if they're not actually necessary for deployment.

frontend/src/components/core/Link.tsx
```diff
@@ -15,8 +15,22 @@ const createButtonLink = (Button:any, props:any) => {
...
+    </Button>
+  );
+
+  return button;
+};
+
 export const PrimaryButtonLink = (props:any) => createButtonLink(PrimaryButton, props);
 
+export const SocketButtonConnection = (props:any) => createButtonOnClick(PrimaryButton, props);
+
```
alankbi, line 33: nit: I think you can get away without this and just do something like this in the `components/core/Button.tsx` folder, since there isn't actually any Link wrapping involved:

```js
export const SocketButton = styled.button`
  // styles go here
`;
```

frontend/src/util/Utility.ts
```diff
@@ -0,0 +1,19 @@
+import SockJS from "sockjs-client";
```
alankbi, line 1: Could this maybe go in the `api` folder, since this [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API) page refers to it as an API? Or maybe we could rename it to `sockets` or something, mainly cause to me this file seems more important than just a utility class.

frontend/src/util/Utility.ts
```diff
@@ -0,0 +1,19 @@
...
+import Stomp from "stompjs";
+
+let stompClient:any = null;
+
+export const Connect = (endpoint:string) => {
```
alankbi, line 6: Miniscule nit: rename non-component functions to lower-case.

frontend/src/util/Utility.ts
```diff
@@ -0,0 +1,19 @@
...
+    console.log('Connected: ' + frame);
+  });
+}
+
+export const Disconnect = () => {
+  if (stompClient !== null) {
+    stompClient.disconnect();
+  }
+  console.log("Disconnected");
+}
```
alankbi, line 19: nit: Enable and fix ESLint errors in this file

src/main/java/com/rocketden/main/config/WebSocketConfig.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.web.socket.config.annotation.WebSocketMessageBrokerConfigurer;
+
+@Configuration
+@EnableWebSocketMessageBroker
+public class WebSocketConfig implements WebSocketMessageBrokerConfigurer {
+
+	@Override
+	public void configureMessageBroker(MessageBrokerRegistry config) {
+		config.enableSimpleBroker("/topic");
+		config.setApplicationDestinationPrefixes("/app");
```
alankbi, line 16: What do these two lines do?

alankbi (changes requested):
Overall looks good 👍 - all I have are some nits over style before approving.

helenli8181 (approved):
I was (finally!) able to run this without max_user_connections and it worked exactly like the description. Everything looks good with me.

frontend/package.json
zpChris: Yeah no problem, I moved it to the `"dependencies"` list.

frontend/src/components/core/Link.tsx
zpChris: I realized that if I'm just using the `PrimaryButton` style, there wasn't really a need for me to create all those other button types and such -- I just changed the PR to use a `PrimaryButton` directly in the `Join.tsx` page. (This is especially true because eventually these buttons won't be there, so creating an extra style for this task is superfluous work.)

frontend/src/util/Utility.ts
zpChris: Sure, that sounds good -- moved it to the `api` folder in a `Socket.ts` file.

I also renamed `problem.ts` to `Problem.ts` to match -- we've used capitalized files in the frontend throughout this repository, I think that change keeps this uniform.

frontend/src/util/Utility.ts
zpChris: Done!

src/main/java/com/rocketden/main/config/WebSocketConfig.java
zpChris: Good question -- nothing important right now, I just removed them! They have to do with setting preferences and prefixes in paths for sending and receiving messages from the server, but since I'm not doing any of that yet in this PR, I just removed them. You can read more about them at [this Spring tutorial](https://spring.io/guides/gs/messaging-stomp-websocket/) -- do `Command + f` for "configureMessageBroker".

frontend/src/util/Utility.ts
zpChris: I found the issue; the following had to be added to the `.vscode/settings.json` file:

```
"eslint.workingDirectories": [
  "./frontend"
]
```

I fixed the errors in this file (now `Socket.ts`), and there was also a ["prefer-default-export" error](https://github.com/benmosher/eslint-plugin-import/blob/v2.22.0/docs/rules/prefer-default-export.md) in the `Problem.ts` file that I fixed.

alankbi (approved):
Looks good, thanks for making those changes!

PR #17, opened 2020-09-02
Add create and join room functionality

alankbi:
### List of Changes:

* Create POST endpoint to create rooms
* Create PUT endpoint to join rooms (room-player relationship has yet to be added)
* Introduce service package for holding business logic 
* Add [ModelMapper](http://modelmapper.org/getting-started/) dependency for converting between DTOs and Entities 
* Write new integration tests and unit tests

### Notes:

* PR became bigger and more complicated than intended, and as a result, I think there are places where the code structure can be improved upon (specifically the request/response/entity mappings)
  * See [here](https://stackoverflow.com/questions/44349199/reusing-dto-for-various-request-response-types-vs-explicitness-of-what-is-requir) for why there are four almost identical request/response classes
* Hesitantly made a PUT request for joining rooms since neither GET nor POST seemed to perfectly fit the situation
* Returned response object for joining a nonexistent room is the same as the successful one (just with a 404 error code and custom error string), which might also need improvement
  * Should be simpler, more extendable methods of returning errors
* Both mapper classes have static methods for converting between DTOs/entities, which might lead to extendability problems in the future

### Future Improvements:

* Expire rooms after a set amount of time (either delete or move to a history table in the database)
* Add input validation (valid name, valid user, etc.)
* Prevent spam creation of rooms (limit against IP or require authentication maybe)
* Ensure unique room ids (against other active rooms or all past rooms as well)
* Potentially improve code structure with regards to DTO/request/response/entity mappings and error responses

### Screencast and Images:

**Create Room:**
Create Room

<br>

**Join Room:**
Join Room


src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+public class RoomServiceTests {
+
+    @Mock
+    private RoomRepository repository;
+
+    @Spy
+    @InjectMocks
```
alankbi, line 30: Two things happening here:
1. Mock and inject the RoomRepository field variable into RoomService so no database is actually touched
2. Create a spy of the RoomService allowing you to specify a set return value when `generateRoomId` is called

More on spies vs. mocks here: https://www.baeldung.com/mockito-spy#mock-vs-spy-in-mockito

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -0,0 +1,102 @@
...
+
+        JoinRoomRequest request = new JoinRoomRequest();
+        request.setPlayerName("rocket");
+        request.setRoomId(roomId);
+
+        Room room = new Room();
+        room.setRoomId(roomId);
+
+        // Mock repository to return room when called
+        Mockito.doReturn(room).when(repository).findRoomByRoomId(eq(roomId));
```
alankbi, line 65: Note on the following line: 

```java
Mockito.doReturn(room).when(repository).findRoomByRoomId(eq(roomId));
```
This makes it so that when `repository.findRoomByRoomId` is called with a string equal to `roomId`, it will return the `room` object that you defined a couple lines above

src/test/java/com/rocketden/main/util/Utility.java
```diff
@@ -9,4 +9,8 @@ public static String convertObjectToJsonString(Object o) {
         return gson.toJson(o);
     }
 
+    public static <T> T toObject(String json, Class<T> c) {
```
alankbi, line 12: Used the Gson codebase as a reference for this; essentially, the generics allow you to specify a class as the second parameter, and this method will then convert the JSON string into that class and return it.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    public void createRoomSuccess() {
+        // Verify create room request succeeds and returns correct response
+        CreateRoomRequest request = new CreateRoomRequest();
+
+        // Mock generateRoomId to return a custom room id
+        Mockito.doReturn("012345").when(service).generateRoomId();
+        CreateRoomResponse response = service.createRoom(request);
+
+        verify(repository).save(Mockito.any());
```
alankbi, line 47: Syntax: this verifies that the `repository` object's `save` method was called once with any argument. 

See this article: https://www.baeldung.com/mockito-verify

src/main/java/com/rocketden/main/dto/room/CreateRoomResponse.java
```diff
@@ -0,0 +1,14 @@
+package com.rocketden.main.dto.room;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class CreateRoomResponse {
+    public static final String SUCCESS = "Successfully created room.";
```
zpChris, line 9: Do you think we should also include some sort of `CANNOT_CREATE_ROOM` object? Or maybe we only include that later when that error could predictably arise (like when the host is already in another room or something). I'm fine with either, especially since I mentioned earlier to not include additions that are not used in the PR.

src/main/java/com/rocketden/main/dto/room/JoinRoomMapper.java
```diff
@@ -0,0 +1,13 @@
+package com.rocketden.main.dto.room;
```
zpChris, line 1: Note on the whole file: I don't think we need a separate `Mapper` class for each class (`JoinRoomResponse` and `CreateRoomResponse`). I think these could be methods in a `Utility.java` file within a `util` folder, or a separate `Mapper` class that has all of these separate methods.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -0,0 +1,69 @@
...
+    public static final int ROOM_ID_LENGTH = 6;
+
+    private final RoomRepository repository;
+
+    @Autowired
+    public RoomService(RoomRepository repository) {
+        this.repository = repository;
+    }
+
+    public JoinRoomResponse joinRoom(JoinRoomRequest request) {
```
zpChris, line 28: To be clear, this isn't actually adding anybody to the room? If you wanted, you could add a `List<Player>` (`Player` could be its own object or a `String` -- I'd recommend the `Player` object but if that's too many changes no problem if not) to the `Room` class.

If you didn't add this because you wanted to keep the PR smaller, that's also OK, no need to add that to this PR.

src/test/java/com/rocketden/main/util/Utility.java
zpChris: That's awesome, nice job on finding this 👍

src/test/java/com/rocketden/main/RoomTests.java
```diff
@@ -0,0 +1,121 @@
...
+                .andReturn();
+
+        String jsonResponse = result.getResponse().getContentAsString();
+        CreateRoomResponse actual = Utility.toObject(jsonResponse, CreateRoomResponse.class);
+
+        assertEquals(expected.getMessage(), actual.getMessage());
+    }
+
+    @Test
+    public void createAndJoinRoom() throws Exception {
```
zpChris, line 79: Could you add more general comments to this? I feel like I can follow it, but it's a really long test so that addition would be helpful.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Thanks so much for the explanation, very helpful!

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: This looks great, thanks for adding this.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Could you make it specific to check that this is called on a `Room` object? If that's difficult no problem, just let me know.

zpChris (changes requested):
I requested changes because there are a fair amount of nits, as well as some potential structural changes within this PR -- as a whole though, the PR looks fantastic. The testing especially was very rigorous, thanks for adding all of that.

I do want to note that I think our PRs should be smaller moving forward -- I know we've already talked about this, but it took me a couple hours to review and understand all of this code, as there are a lot of interacting parts here. Otherwise, awesome job! 😃 

One other change: In the description, can you add that the `joinRoom(...)` function doesn't actually add the user to the room? I feel like that may not be clear in this PR.

src/main/java/com/rocketden/main/dao/RoomRepository.java
```diff
@@ -0,0 +1,10 @@
+package com.rocketden.main.dao;
```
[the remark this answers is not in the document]
alankbi: I ended up moving `ProblemRepository.java` up a folder and deleting the `problem` folder, since I think there's no need to go too modularized (similar to what we decided for the frontend).

src/main/java/com/rocketden/main/dao/RoomRepository.java
```diff
@@ -0,0 +1,10 @@
...
+
+import com.rocketden.main.model.Room;
+import org.springframework.data.repository.CrudRepository;
+
+public interface RoomRepository extends CrudRepository<Room, Integer> {
```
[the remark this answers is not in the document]
alankbi: I don't know how `CrudRepository` actually does it internally, but this is the doc explaining a little bit about query creation: https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods.query-creation

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.persistence.Id;
+import java.time.LocalDateTime;
+
+@Entity
+@Getter
+@Setter
+public class Room {
+
+    @Id
+    @GeneratedValue(strategy= GenerationType.AUTO)
```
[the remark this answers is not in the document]
alankbi: Done, although will note that in the `controller` package we have lots of times where we do `value = ...` (spaces around the equals sign), so we might want to go back and standardize everything at some point.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        JoinRoomResponse response = JoinRoomMapper.entityToResponse(room);
+        response.setMessage(JoinRoomResponse.SUCCESS);
+        response.setPlayerName(request.getPlayerName());
+
+        return response;
+    }
+
+    public CreateRoomResponse createRoom(CreateRoomRequest request) {
```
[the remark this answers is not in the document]
alankbi: Removed it for now 👍

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -0,0 +1,24 @@
...
+    private Integer id;
+
+    private String roomId;
```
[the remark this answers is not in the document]
alankbi: I didn't want to make this PR any bigger, so I made a [card](https://trello.com/c/yrjzo9Ry/92-handle-unique-room-ids-keep-only-one-of-id-roomid-and-or-add-letters) in the Trello backlog so we can revisit all of changing to one of id/roomId, switching to GenerationType, and/or using alphanumeric rather than just numeric. Also, I originally had both `id` and `roomId` for this [security](https://stackoverflow.com/questions/396164/exposing-database-ids-security-risk) reason, although it may or may not really apply to us. Overall though, I think there's a decent amount to discuss about improving the room id design that it could warrant it's own task.

src/main/java/com/rocketden/main/dto/room/JoinRoomResponse.java
```diff
@@ -0,0 +1,17 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class JoinRoomResponse {
+
+    public static final String SUCCESS = "Sucessfully joined room.";
+    public static final String ERROR_NOT_FOUND = "A room could not be found with the given id.";
```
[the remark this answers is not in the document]
alankbi: I was actually thinking of removing these all together in favor of the error handling design [proposed here](https://stackoverflow.com/a/50053782/7386515); thoughts? That's probably not gonna make it in this PR though, so lmk and I can implement this in the meantime.

src/main/java/com/rocketden/main/dto/room/CreateRoomResponse.java
alankbi: I was thinking that too but then decided against it for exactly that reason you mentioned. In the future though I think we'll definitely need to have that.

src/main/java/com/rocketden/main/dto/room/JoinRoomMapper.java
alankbi: Merged the two files into a single `RoomMapper` class. As we add more mappings I think it might make sense to move it all into a single `utils` folder as you suggested.

src/main/java/com/rocketden/main/service/RoomService.java
alankbi: I tried adding a list of strings but then got [this error](https://stackoverflow.com/questions/21059451/cannot-declare-list-property-in-the-jpa-entity-class-it-says-basic-attribute), so I think the only way would be to create a new Player entity, which might be beyond the scope of the PR. However I added a card to the backlog for this [here](https://trello.com/c/sj0GCL37/97-actually-add-players-to-rooms).

src/main/java/com/rocketden/main/dto/room/JoinRoomResponse.java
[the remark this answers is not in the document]
zpChris: Based on a brief look through, this looks fantastic -- definitely implement this in another PR. I'd recommend those changes would be their own PR (to keep the size manageable).

src/main/java/com/rocketden/main/model/Room.java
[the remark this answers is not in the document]
zpChris: Ah, I didn't think about any security problems -- that's a good point. Thanks for making the card, and I agree with adding all of those tasks and addressing them in another PR. Great response 🙂

src/main/java/com/rocketden/main/service/RoomService.java
zpChris: Ok, that sounds OK -- thank you again for adding the card to the Trello board, much appreciated (good name too haha, "**Actually** add players to rooms").

src/test/java/com/rocketden/main/RoomTests.java
zpChris: Thanks for the additions 👍

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Just wanted to note that you left a 👍 on my comment, but didn't change this -- was that on purpose? If yes, sounds good as no change was required, but wanted to let you know in case.

src/main/java/com/rocketden/main/model/Room.java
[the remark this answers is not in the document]
zpChris: What are your thoughts about doing that in this PR? Seems like as good a time as any (either standardize by adding spacing or removing it, either is fine by me).

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -0,0 +1,102 @@
...
+public class RoomServiceTests {
+
+    @Mock
+    private RoomRepository repository;
+
+    @Spy
+    @InjectMocks
+    private RoomService service;
+
+    @Before
```
zpChris, line 32: I have something called [SonarLint](https://www.sonarlint.org/) installed that is showing the following error (see image). 
SonarLint Issue on `@Before` Tag
For my PR in `ProblemTests.java`, I simply put these in a constructor because I didn't want to do `@BeforeEach` since it just needed to be run once. I'm assumming that's the same with yours -- I don't have a preference for what you want to do here, but I'm interested in your thoughts.

zpChris (approved):
LGTM! Fantastic and thorough changes, much appreciated. I left a few nits but nothing major at all. Looks great, nice job! 👍

One note: You did keep the **Future: ...** comments in the repository. I think I'm ok with that, my preference is to leave those out though so they don't clutter the repository. Maybe we can have a quick discussion on this to make sure the codebase still stays clean and up-to-date.

alankbi:
Addressed some of your final comments (summarized below):

* Get rid of future comments
* Checked `Mockito.any()` - in the code it's been fixed but the code block in GitHub shows the outdated version
* Added spaces around the `=` in the annotation parameters, mainly to be consistent with how the [Spring tutorials](https://github.com/spring-projects/spring-petclinic) do it
* Changed the `@Before` import statement - turns out I didn't even need that method so I deleted it 
  * Worth also noting that I think the constructors and before methods are (almost) equal; see [here](https://stackoverflow.com/questions/6094081/junit-using-constructor-instead-of-before)

src/test/java/com/rocketden/main/service/RoomServiceTests.java
alankbi: Thanks, fixed this (by just deleting the method since the tests passed without it). More details below.

zpChris (approved):
LGTM! Thanks for making all of those changes. 👍

src/main/java/com/rocketden/main/dto/room/JoinRoomRequest.java
```diff
@@ -0,0 +1,11 @@
+package com.rocketden.main.dto.room;
+
+import lombok.Getter;
+import lombok.Setter;
```
helenli8181, line 4: Not a critique but I didn't know you could make getters and setters this way! note to self to try this in the future

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -0,0 +1,24 @@
...
+@Setter
+public class Room {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Integer id;
+
+    private String roomId;
+
+    private LocalDateTime createdDate = LocalDateTime.now();
```
helenli8181, line 23: Minor detail but would this field make more sense to be named as `createdDateTime` since it also includes time and not just date?

helenli8181 (approved):
This all looks great! 👌After the changes you made I only had a nitpick left.

src/main/java/com/rocketden/main/model/Room.java
alankbi: Yeah that makes sense, just updated it and will merge now 👍

PR #22, opened 2020-09-07
Display problems and game page redesign

helenli8181:
### List of Changes:

- `problem.ts`: Made type `Problem` into an export, and changed `getProblems` from a default to named export.
- `Game.tsx`: I added interfaces, changed the `GamePage` function into a class, and added a constructor and a render function which retrieve the problem data from `problem.ts`. Styling is contained within this class as style components.
- I used FlexBox to arrange the problem and code editor panels. 

### Notes:

- For now, it just automatically displays the first problem in the database. Of course, later we will need to be able to specify which problem, probably through random selection.
- If there is an error with finding the problem to display the error will be printed in the console. In the future it should display an error message on the webpage.  

### Screencast and Images:

Screen Shot 2020-09-11 at 5 19 07 PM


frontend/src/views/Game.tsx
```diff
@@ -1,13 +1,44 @@
...
-    <div>
-      Game page
-      <Editor height="90vh" language="javascript" />
-    </div>
-  );
+interface GameState {
+  problems?: Problem[];
+}
+
+interface GameProps {}
```
alankbi, line 10: What's the difference between interface and type? If we use interface here should we be doing the same elsewhere, or vice versa?

frontend/src/views/Game.tsx
```diff
@@ -1,13 +1,44 @@
...
+    return (
+      <div>
+        <div className={styles.header}><h3>Game page</h3></div>
+        <div className={styles.problemPanel}>
+          <h3>{ firstProblem?.name }</h3>
+          <p>{ firstProblem?.description}</p>
+        </div>
+        <div className={styles.editorPanel}>
+          <Editor height="90vh" language="javascript" />
+        </div>
```
alankbi, line 38: nit: Could you replace `<p>` with `Text` from `src/components/core/Text.tsx` and create a new component for `<h3>` along the lines of the other components there? That way these font sizes can be reused on other pages too.

frontend/src/views/Game.tsx
```diff
@@ -1,13 +1,44 @@
...
+  constructor(props: GameProps) {
+    super(props);
+    this.state = {};
+  }
+
+  componentDidMount() {
+    getProblems().then((problems) => {
+      this.setState({ problems });
+    }).catch((error) => {
+      console.log(error);
```
alankbi, line 22: We may want to add some error validation to display an error component if something goes wrong, although that might come later since it'd apply to all the different game pages

frontend/src/views/styles.module.css
```diff
@@ -0,0 +1,28 @@
+.problemPanel {
```
alankbi, line 1: nit: Move this into `Game.tsx` using [styled components](https://styled-components.com/), similar to how it's done [here](https://github.com/rocketden/main/blob/master/frontend/src/components/core/Button.tsx).

frontend/src/views/styles.module.css
```diff
@@ -0,0 +1,28 @@
...
+    position: absolute;
+    top: 80px;
+    left: 0;
+    bottom: 0;
+    width: 50%;
+    display: inline-block;
+    background-color: #e3e3e3;
```
alankbi, line 8: Note for future reference: these colors might move into the Theme file once we come back and properly design everything

alankbi (changes requested):
Page design looks great to me! Before merging, I requested a couple of changes and also had two notes/questions:

* What happens if no problems exist in the database, and do we need to worry about that (eventually the way problems are fetched will be changed, but idk if we should add an error check for that now or have a separate story to implement error checks for all pages)? 
* When I drag the screen width to below 640px, the components keep shrinking even though there's a global style setting min-width to 640px. Not sure why that happens, but I think it might have to do with using `%` vs. `vw`.

frontend/src/views/styles.module.css
```diff
@@ -0,0 +1,28 @@
...
+
+.header {
+    position: absolute;
+    top: 0;
+    width: 100%;
+    height: 80px;
+    display: block;
+    background-color: #85d2ff;
+    text-align: center;
+}
```
zpChris, line 28: nit: Add endline to the end of the file.

This is also an **eslint** suggestion, though that suggestion may not pop up on `css` files.

zpChris, line 28: We already have a `header.tsx` file; I think we should use that instead of creating a new custom Header.

frontend/src/views/Game.tsx
```diff
@@ -1,13 +1,44 @@
...
-      <Editor height="90vh" language="javascript" />
-    </div>
-  );
+interface GameState {
+  problems?: Problem[];
+}
+
+interface GameProps {}
+
+class GamePage extends React.Component<GameProps, GameState> {
```
zpChris, line 12: I'm curious on @alankbi's thoughts on this; I was under the impression that we were going to use functions, and incorporate the `state` variable similar to [this React tutorial](https://reactjs.org/docs/hooks-state.html). I'm assuming you changed it to a class when seeing [the "State and Lifecycle" React tutorial](https://reactjs.org/docs/state-and-lifecycle.html) -- this may be better I'm just not sure, do you guys know what the pros / cons are?

frontend/src/api/problem.ts
```diff
@@ -1,4 +1,4 @@
-type Problem = {
+export type Problem = {
```
zpChris, line 1: Comment on the whole `problem.ts` file: I think the merge conflicts was likely because I changed this file name to `Problem.ts` instead of `problem.ts`. That matches with the file naming pattern -- can you change this file name to `Problem.ts`?

frontend/src/api/problem.ts
```diff
@@ -10,7 +10,5 @@ const routes = {
   postProblem: `${basePath}/`,
 };
 
-const getProblems = (): Promise<Problem[]> => fetch(routes.getProblems)
+export const getProblems = (): Promise<Problem[]> => fetch(routes.getProblems)
```
zpChris, line 13: Note: No action required, but I'd like to get input on this; I think we determined last time that the `Problem[]` type doesn't actually type-check the function correctly. If this is the case, should we try to fix it, or is it not a big deal?

frontend/src/views/styles.module.css
zpChris: Thoughts on also changing the styling to use **flexbox**? A guide on CSS-tricks is [here](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#flexbox-background); I think this would be easier to design with in the future. I'd also like to get @alankbi's thoughts on this.

frontend/src/views/Game.tsx
```diff
@@ -1,13 +1,44 @@
...
+    const firstProblem = problems?.[0];
+    return (
+      <div>
+        <div className={styles.header}><h3>Game page</h3></div>
+        <div className={styles.problemPanel}>
+          <h3>{ firstProblem?.name }</h3>
+          <p>{ firstProblem?.description}</p>
+        </div>
+        <div className={styles.editorPanel}>
+          <Editor height="90vh" language="javascript" />
```
zpChris, line 37: Note: A scrollbar appears next to the `Editor` when you add new lines, even when it doesn't go past the bottom of the page. Is this fixable?

frontend/src/views/styles.module.css
```diff
@@ -0,0 +1,28 @@
+.problemPanel {
+    position: absolute;
+    top: 80px;
+    left: 0;
+    bottom: 0;
+    width: 50%;
+    display: inline-block;
```
zpChris, line 7: nit: Add `padding` to the problem panel?

zpChris (changes requested):
I have a couple requested changes on design layout, primarily on the `styles.module.css` file; I also left a few comments bringing up topics with no required changes, just choices I think we should discuss (for both @helenli8181 and @alankbi).

frontend/src/views/Game.tsx
helenli8181: I used interfaces here because they are more ideal for defining the props & states required by React.Component. Generally types and interfaces in Typescript are becoming more similar to each other but there are specific cases where one is better than the other, i.e. types rather than interfaces should be used as arguments or return types of a function. I think it's fine to have both in our program.

helenli8181:
> Page design looks great to me! Before merging, I requested a couple of changes and also had two notes/questions:
> 
> * What happens if no problems exist in the database, and do we need to worry about that (eventually the way problems are fetched will be changed, but idk if we should add an error check for that now or have a separate story to implement error checks for all pages)?
> * When I drag the screen width to below 640px, the components keep shrinking even though there's a global style setting min-width to 640px. Not sure why that happens, but I think it might have to do with using `%` vs. `vw`.

@alankbi On the second point, I tried switching to vw instead of % but the window behaved the same way. In both cases the components continued to shrink as long as the browser window could.

frontend/src/api/problem.ts
helenli8181: I think it's fine to leave as is

frontend/src/views/Game.tsx
helenli8181: It looks like there's no way to make it disappear completely even while turning its visibility to "hidden". There doesn't seem to be an option to remove the large chunk of white space under the last line either.

frontend/src/views/Game.tsx
helenli8181: I'll add this to the trello board!

frontend/src/views/Game.tsx
helenli8181: The main difference between `useState` in a function and `setState` in a class is that `setState` allows you to define multiple state variables at once while `useState` can only have one. This purpose isn't really expressed in my code as it is because I'm using only one variable which specifies whether problems exist to be displayed. The `componentDidMount()` function in a class can be incredibly useful -- it allows you to wait until other components are rendered before running a certain piece of code. I think I can transition this `GamePage` class into a function if you guys think it's better, I can probably even get rid of `componentDidMount()`,  however I don't think we should necessarily try to avoid using classes for the sake of consistency.

alankbi: Yeah so I think React Hooks (the functional components + useXXX functions) are newer and the ["future"](https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both) of React, but were only released recently and class components are still more common in existing apps/tutorials. For the sake of consistency and long-term prospects, it might be good to stick with functional components, but I think the difference isn't too important and we can always convert them later.

frontend/src/api/problem.ts
alankbi: To update on this, I think [this link](https://stackoverflow.com/questions/55960602/how-to-cast-fetch-api-response-to-a-specific-type-in-typescript) explains the behavior well; basically, TypeScript can't actually validate the API responses match the defined types (cause it comes in at runtime), so it's just up to us to make sure it matches properly (so that at compile time in your IDE TypeScript can work properly when you use those types elsewhere in the app).

frontend/src/views/Game.tsx
alankbi: Maybe that's part of the desired behavior? It seems like it makes it so that you can always scroll down just enough so that you don't see any of the lines you typed (IIRC VSCode does the same). Idk if there's a way to toggle that or not.

alankbi:
> @alankbi On the second point, I tried switching to vw instead of % but the window behaved the same way. In both cases the components continued to shrink as long as the browser window could.

I think it's working now actually, once you decrease the width below 640px it stops shrinking so nice work on that :) 

I think most of my changes were addressed but I'll wait till @zpChris takes a look before doing doing a final run-through/approval if that sounds good

frontend/src/views/Game.tsx
zpChris: I also want to add that I think `useState` allows you to set multiple state variables; I did so in PR #26. As for converting them later @alankbi, personally I don't think it would make sense to change it now just to change it back later -- whichever we choose, I think we should use now, as it'd make development much easier.

zpChris:
[Display problems on game page](https://trello.com/c/oA6wA3WY/83-display-problems-on-game-page)

frontend/src/views/Game.tsx
helenli8181: Sounds good, I am switching the class into a function.

frontend/src/views/Game.tsx
helenli8181: I'll make a note to look more into this issue when we're working more on the design! I think what's happening is there's an empty block under the code editor that stays the same size even when the editor expands.

helenli8181:
New screenshot after merging with Master: 
Screen Shot 2020-09-14 at 2 09 20 PM

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
+  top: 50px;
+`;
+
+const FlexPanel = styled.div`
+  flex: 1;
+  padding: 1rem;
+  background-color: #e3e3e3;
+  overflow: none;
+  height: 100vh;
+  width: 50vw;
```
zpChris, line 25: I'm having trouble understanding if some of these attributes are actually contributing anything? Notably, the following don't seem to be doing anything:
```
flex: 1;
width 50vw;
```
Since you already have the `flex: 1` value, I don't know that the others are necessary?

I'd also consider looking into how you can update this to make the height of "Header + Problem & Code Panel" equal to the screen height (I also referenced this in a comment below).

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
+`;
+
+export default function GamePage() {
   const location = useLocation<LocationState>();
   const [room, setRoom] = useState<Room | null>(null);
 
+  const [problems, setProblems] = useState<Problem[] | null>(null)
```
zpChris, line 32: Add ending semicolon to the line.

Note: You should be seeing this (also ending newline) through ESLint.

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
+const FlexPanel = styled.div`
+  flex: 1;
+  padding: 1rem;
+  background-color: #e3e3e3;
+  overflow: none;
+  height: 100vh;
+  width: 50vw;
+`;
+
+export default function GamePage() {
```
zpChris, line 28: nit: Either keep this and change the other `views` files to this format, or change this to `function GamePage()` and keep the other files as-is. Personally, consistency of structure across files is a personal preference for me, and I think will help with readability moving forward.

frontend/src/views/Game.tsx
zpChris: This is more of an open discussion, both for @alankbi and @helenli8181, as I'm open to either. I wanted to propose resolving this immediately within this PR, because I think it may be good to say that when you introduce `catch` functions that are _going to stay_ (this isn't temporary code, like in PR #23), if the PR is not already bloated we add it to that same PR -- what are both of your thoughts?

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
       setRoom(location.state.room);
     }
+    getProblems().then((response) => {
+      setProblems(response);
+    }).catch((error) => {
+      console.log(error);
+    });
   }, [location]);
 
+  const firstProblem = problems?.[0];
```
zpChris, line 46: Add error handling if there are no problems?

zpChris (changes requested):
The design looks better, thank you! I'm still requesting changes because the code has some ESLint errors, you accidentally left the old merged code in some places which means the header is below the first code editor, and some of the styling seems superfluous.

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
 import { Room } from '../api/Room';
+import { Text, ProblemHeaderText } from '../components/core/Text';
+import Header from '../components/navigation/Header';
 
 type LocationState = {
   room: Room,
 }
 
-function GamePage() {
+const FlexContainer = styled.div`
```
zpChris, line 14: Forgot to add this during the review, so leaving it as a single comment -- could these Container components go in the `Container.tsx` `core` file?

frontend/src/components/core/Text.tsx
```diff
@@ -15,3 +15,7 @@ export const LargeText = styled.h3`
 export const LandingHeaderText = styled.h1`
   font-size: ${({ theme }) => theme.fontSize.xxLarge};
 `;
+
+export const ProblemHeaderText = styled.h3`
+  font-size: ${({ theme }) => theme.fontSize.Medium};
```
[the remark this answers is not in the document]
helenli8181: Medium is a font size, but it's the default value so technically line 20 is not necessary. I kept it there just to have a `ProblemHeaderText` because we'll probably need more styling on it later.

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
+  display: flex;
+  top: 50px;
```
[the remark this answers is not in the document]
helenli8181: You're right, it doesn't do anything anymore. Thanks for pointing it out!

frontend/src/views/Game.tsx
helenli8181: Ok I got rid of those two lines, they may have been from an old commit that got mixed in

frontend/src/views/Game.tsx
helenli8181: The error handling is supposed to be represented by line 42. Here the question mark accounts for the possibility of problems not existing. When I test this with only `npm start` so that it can't access the problem data, it prints an error to the console. Other than changing the error validation as in the above comment, are there any other ways you'd recommend doing this?

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
+    getProblems().then((response) => {
+      setProblems(response);
+    }).catch((error) => {
+      console.log(error);
+    });
   }, [location]);
 
+  const firstProblem = problems?.[0];
+
   return (
```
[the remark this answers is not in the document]
helenli8181: Sorry about that, I had rearranged the components but didn't notice that this version was not updated.

frontend/src/views/Game.tsx
zpChris: I'd recommend displaying error text to show that an error occurred when trying to get something to load: see what @alankbi did in his PR [here](https://github.com/rocketden/main/pull/24/files#diff-c47bfa9c3f1b80608df1f7a1dc057e30R48), for example.

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
zpChris: What are your thoughts on styling it now?

helenli8181:
Error message displayed when problem cannot load:

Screen Shot 2020-09-14 at 7 55 45 PM

frontend/src/views/Game.tsx
```diff
@@ -1,31 +1,68 @@
...
       <Editor height="90vh" language="javascript" />
+      <div>
+        <Header />
+        <FlexContainer>
+          <FlexPanel>
+            <ProblemHeaderText>{ firstProblem?.name }</ProblemHeaderText>
+            <Text>{ firstProblem?.description }</Text>
+          </FlexPanel>
+          <FlexPanel>
+            <Editor height="100vh" language="javascript" />
```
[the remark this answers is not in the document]
helenli8181: The Editor only fills the space within its `FlexPanel` component; this works even with other FlexBox components occupying the page as shown by the new `RoomInfoBar` component.

frontend/src/views/Game.tsx
helenli8181: Just put in a proper error handling system; I tried to make it as consistent as possible with Alan's in `Landing.tsx` though there are a few small differences.

frontend/src/components/core/Text.tsx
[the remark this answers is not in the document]
zpChris: Where do you see Medium is a font size? The below image shows the font sizes I can see -- where is Medium? If it is equal to `default`, why not just use `default` like we have elsewhere?
Medium Font Size Not Found

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
zpChris: Sorry, I think I didn't clarify -- IMO the page design should match [LeetCode](https://leetcode.com/problems/single-number-ii/), where you cannot scroll any further on the page (just within the divs, but the bottom of the page remains the footer with things like "Console" and "Submit"). I was asking if you could construct this so it would match that layout; essentially, so you cannot scroll on the whole page.

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,45 @@ type LocationState = {
...
-      {' '}
-      {room ? room.roomId : 'No room joined'}
-      <Editor height="90vh" language="javascript" />
+      <Header />
+      <FlexContainer>
+        <FlexInfoBar>
+          Room:
+          {' '}
+          {room ? room.roomId : 'No room joined'}
+        </FlexInfoBar>
```
zpChris, line 48: nit: Reduce distance between this and the header? The below image seems like a big gap to me.
Info Bar and Header

frontend/src/views/Game.tsx
zpChris: Bump @helenli8181 -- I think you forgot to respond to this before re-request?

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,45 @@ type LocationState = {
...
 
   // Called every time location changes
   useEffect(() => {
     if (location && location.state && location.state.room) {
       setRoom(location.state.room);
     }
+    getProblems().then((res) => {
+      if (isError(res)) {
+        setError((res as ErrorResponse).message);
+        setProblems(null);
```
zpChris, line 30: I believe this is superfluous, as it is already set above.

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,45 @@ type LocationState = {
...
+      } else {
+        setProblems(res as Problem[]);
+        setError(null);
```
zpChris, line 33: Same here; not needed I believe.

zpChris (changes requested):
Note: Most of my changes were left as comments -- not sure how this happened, sorry about that. You should be able to see more comments above, let me know if you have any trouble addressing them.

Thanks for making those changes! I'm requesting a few more edits, mainly because I want to try to get the whole game site to all be on one page without scrolling (I think that was miscommunicated earlier, sorry if that wasn't clear in my last review). Also, I have some nits on error handling.

I think I'm going to wait for @alankbi to review before I review again. 👍

frontend/src/views/Game.tsx
helenli8181: Resetting to null is safety for when this `useEffect()` function is called multiple times in one session. Originally, my `componentsDidMount()` equivalent would only have been called once but after merging with Alan's code it's supposed to rerun with every location change.

frontend/src/views/Game.tsx
helenli8181: Same response as above

frontend/src/components/core/Text.tsx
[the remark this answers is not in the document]
helenli8181: Good point, I'll fix that.

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,45 @@ type LocationState = {
...
+      <FlexContainer>
+        <FlexInfoBar>
+          Room:
+          {' '}
+          {room ? room.roomId : 'No room joined'}
+        </FlexInfoBar>
+        <FlexPanel>
+          <ProblemHeaderText>{ firstProblem?.name }</ProblemHeaderText>
+          <Text>{ firstProblem?.description }</Text>
+          <ErrorText>{error ? <ErrorMessage message={error} /> : null}</ErrorText>
```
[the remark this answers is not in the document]
helenli8181: Got rid of the `<ErrorText>` tag; I didn't realize that the red style was a part of `<ErrorMessage>` already.

frontend/src/views/Game.tsx
helenli8181: This gap is mainly due to the Header being very tall while the words are squished at the top. Since the Header affects every page, do you think I should change it?

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
helenli8181: Oh I see what you mean. The page should fit the browser window now.

frontend/src/views/Game.tsx
zpChris: I'll leave that largely up to your discretion -- if you do that, make sure the other pages look OK too. If not, I'd just modify the margin of the `FlexInfoBar` (you could also make this margin negative, but I'd be careful when doing that as I'm not sure if it's good design practice).

frontend/src/components/core/Container.tsx
```diff
@@ -10,6 +10,27 @@ const createContainer = (Container: any, props: MyProps) => {
   return <Container>{children}</Container>;
```
alankbi, line 10: Unrelated to your PR but I just realized that this code I wrote is probably completely unnecessary haha

frontend/src/components/core/Container.tsx
```diff
@@ -10,6 +10,27 @@ const createContainer = (Container: any, props: MyProps) => {
...
+  height: 30px;
+  line-height: 30px;
+  text-align: center;
+`;
+
+export const FlexPanel = styled.div`
+  flex: 1;
+  padding: 1rem;
+  background-color: #e3e3e3;
+`;
```
alankbi, line 32: I'm actually not sure if these two should go in the container file; at the same time, I think it does make sense to have the flex container in this file, and these three components should be grouped together. So idk really what the best situation is here. Can FlexContainer be reused in other classes (I'm not too familiar with flex)? If so, then maybe keep that in here and over flex info bar/panel back to Game.tsx. If it's pretty specific to the game page/only is useful with the info bar and panel, then I think maybe it can actually be put back in the Game.tsx file (maybe renaming it to Content that way it doesn't get flagged for refactoring later on).

I'm not really familiar or sure on what the best route forward is with this, so leaving it for now should be also fine.

frontend/src/components/core/Text.tsx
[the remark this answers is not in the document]
alankbi: Has this been fixed yet?

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,45 @@ type LocationState = {
 function GamePage() {
   const location = useLocation<LocationState>();
   const [room, setRoom] = useState<Room | null>(null);
+  const [problems, setProblems] = useState<Problem[] | null>(null);
+  const [error, setError] = useState<string | null>(null);
```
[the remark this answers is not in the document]
alankbi: Has this been fixed yet?

frontend/src/views/Game.tsx
alankbi: Same as my previous comment about whether this has been changed to `setError('')`. I may also just be looking at an outdated version so sorry if that's the case

frontend/src/views/Game.tsx
alankbi: As discussed in the call, I think we want to have some error check in case zero problems are returned (an empty list). As for how to go about that, not too sure, but a proposed solution would maybe be in the `getProblems` then block, add another else-if to see if `problems[]` is empty (if so call setError with a custom message), and if it's not, then just save the first problem in state (rather than the whole list).

frontend/src/views/Game.tsx
alankbi: Not sure what's causing the space being so large (probably partly header but also maybe has something to do with padding/margin of flex container/info bar), but if you want to customize the header bar, I'd recommend creating a new, say, GameHeader file in `navigation`, which can have custom styling and text. I'm actually not sure anymore whether that should be put in the GameLayout file or whether it can be put directly in this file, because if we want to store the room id in the header (as in the mockup), then we'll need to pass that info as props to the header, which might be hard if it's in a separate file.

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
alankbi: Not sure if this has been updated already, but if not, this can simply be `{error ? <ErrorMessage message={error} /> : null}` (without the ErrorText wrapped around it).

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,45 @@ type LocationState = {
...
+          {' '}
+          {room ? room.roomId : 'No room joined'}
+        </FlexInfoBar>
+        <FlexPanel>
+          <ProblemHeaderText>{ firstProblem?.name }</ProblemHeaderText>
+          <Text>{ firstProblem?.description }</Text>
+          <ErrorText>{error ? <ErrorMessage message={error} /> : null}</ErrorText>
+        </FlexPanel>
+        <FlexPanel>
+          <Editor height="100vh" language="javascript" />
```
alankbi, line 55: Not sure if the full-height thing has been fixed, but if not, I think it may have to do with the `100vh` here. Not sure if the solution would be to get rid of it (making it default back to `100%`) or something else though.

alankbi (changes requested):
Mainly minor nits and suggestions/questions building on Chris' comments, but otherwise it's almost there I think!

frontend/src/views/Game.tsx
helenli8181: Working on it now :)

frontend/src/views/Game.tsx
helenli8181: I just made a change where if the problem set is an empty array in `Game.tsx` then it displays a different error message. It doesn't seem like good practice to put if/else statements in the then block in getProblems(), if that's even possible.

frontend/src/views/Game.tsx
helenli8181: Yep, it got fixed --  it was actually changing % to vh that did it haha

frontend/src/views/Game.tsx
helenli8181: I reduced the height of the room info bar; it looks slightly better but a header redesign is probably required later.

frontend/src/components/core/Container.tsx
```diff
@@ -10,6 +10,27 @@ const createContainer = (Container: any, props: MyProps) => {
...
+  padding: 0.5rem;
+  height: 15px;
+  line-height: 15px;
+  text-align: center;
+`;
+
+export const FlexPanel = styled.div`
+  flex: 1;
+  padding: 1rem;
+  background-color: #e3e3e3;
```
alankbi, line 31: Minor nit: Move color to Theme

frontend/src/views/Game.tsx
alankbi: Sounds good; also, I deleted all problems in the database to see how `problems?.[0]` would behave and everything seems to be good 👍

alankbi (approved):
Thanks for addressing the changes! I left another minor nit to move one of the colors to the Theme file, but after that, looks good to me 👍

frontend/src/components/core/Container.tsx
zpChris: Bump @helenli8181

frontend/src/views/Game.tsx
```diff
@@ -10,20 +16,47 @@ type LocationState = {
...
+  const firstProblem = problems?.[0];
+
   return (
     <div>
-      Room:
-      {' '}
-      {room ? room.roomId : 'No room joined'}
-      <Editor height="90vh" language="javascript" />
+      <FlexContainer>
+        <Header />
```
zpChris, line 45: Thoughts on adding this `Header` to the `Game.tsx` page, similar to `Main.tsx`?

zpChris (approved):
LGTM! I want to note that there are changes requested before you merged, including one from @alankbi you still need to address. Please do those and make sure all comments are resolved before merging.

After that, let's merge! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

frontend/src/components/core/Container.tsx
helenli8181: I changed the game page to use the Theme background color so it matches our other pages now. The only drawback of this is the header and problem panel are now the same color.

PR #23, opened 2020-09-08
Add socket subscription and messaging functionality.

zpChris:
### List of Changes:
- Add backend functionality to allow users to subscribe to a URL, and send messages / information to that URL. The other users who are subscribed to that URL will then receive those sent messages.
- Add a `UserIntegrationTests.java` file that tests the connection / subscription / messaging functionality.
- Create a `User.java` class that holds the nickname with which users connect to sockets.
- Add checks to confirm that the User cannot connect if already connected, and cannot disconnect or send messages if disconnected.
- Add an **Input** element to let the user enter a nickname.
- Remove the `Message.java`, `MessageController.java`, and `MessageTests.java` classes.
- Change `export` functions and variables across many `React.js` files to **default exports** (an **eslint** warning).

### Notes:
- I continued to use [this Spring socket tutorial](https://spring.io/guides/gs/messaging-stomp-websocket/) to build the base for this repository -- especially the testing.
  - I initially had a `Color` object within the `User.java` class, but this caused testing errors, so I removed this (at least for now). I may reintroduce this as a `String` with the relevant hex values, or a custom `Color` object (I previously used `java.awt.Color`).
- When the user connects to the socket, nothing is stored in the database, so none of the user data is permanent.
- If the checks to prevent the user from connecting to the socket twice do not exist, future received messages are duplicated multiple times.
- If the user connects or sends user information without any nickname in the input field, the nickname "Anonymous" is provided.

### Screencast and Images:
[Socket Subscription and Send Message Screencast](https://www.loom.com/share/52a79a191468410cbb4e98862d1e913f)

Image of Socket Connection and Messages:
Image of Socket Connection and Messages


src/main/java/com/rocketden/main/controller/v1/BaseRestController.java
```diff
@@ -2,7 +2,11 @@
...
 import org.springframework.web.bind.annotation.RequestMapping;
 
+// Sets the base path for API endpoints. Custom controllers should extend this class.
 @RequestMapping("/api/v1")
 public class BaseRestController {
-    // Sets the base path for API endpoints. Custom controllers should extend this class.
+
+    // Constants to hold custom paths.
+    public static final String BASE_URL = "/api/v1";
+    public static final String BASE_SOCKET_URL = "/api/v1/socket";
```
zpChris, line 11: There is a **sonarlint** warning (see image) on this class as it sees it as a Utility class that should thus have a private constructor; however, when I added this, it created errors when running the application.

BaseRestController Error

Thus, I didn't address the error; if there is any action I should take to address the warning, please let me know.

frontend/src/api/Socket.ts
```diff
@@ -3,14 +3,52 @@ import Stomp from 'stompjs';
...
+// Variable to hold the current connected state.
+let connected = false;
+
+// Create constants for the subscription and send message URLs.
+const SUBSCRIBE_URL:string = '/api/v1/socket/subscribe-user';
+const SEND_USER_URL:string = '/api/v1/socket/user';
+
+export const sendUser = (nickname:string) => {
+  if (connected) {
+    const nicknameInput:string = (nickname !== '') ? nickname : 'Anonymous';
```
alankbi, line 19: nit: Idk if this actually works or not but maybe try simplifying this to `nickname || 'Anonymous'`

frontend/src/api/Socket.ts
```diff
@@ -3,14 +3,52 @@ import Stomp from 'stompjs';
...
+    stompClient.send(SEND_USER_URL, {}, nicknameInput);
+  } else {
+    console.error('You must be connected to a socket before sending user information.');
```
alankbi, line 22: Note for future: these might eventually have to be converted to some kind of UI error message

frontend/src/api/Socket.ts
```diff
@@ -3,14 +3,52 @@ import Stomp from 'stompjs';
...
+};
+
+export const connect = (endpoint:string, nickname:string) => {
+  if (!connected) {
+    // Connect to given endpoint, subscribe to future messages, and send user message.
+    const socket:any = new SockJS(endpoint);
+    stompClient = Stomp.over(socket);
+    stompClient.connect({}, () => {
+      stompClient.subscribe(SUBSCRIBE_URL, (user:any) => {
+        const userObject:User = JSON.parse(user.body);
```
alankbi, line 33: Question: Is it possible to get stomp to return `user` as an object rather than a string?

frontend/src/api/Socket.ts
```diff
@@ -3,14 +3,52 @@ import Stomp from 'stompjs';
...
-  if (stompClient !== null) {
+  if (connected) {
     stompClient.disconnect();
+    stompClient = null;
+    // Reassign connected variable.
+    connected = false;
+  } else {
+    console.error('You cannot disconnect because you are not connected to any socket.');
   }
 };
```
alankbi, line 54: nit: Remove or replace `any` tags, if possible

frontend/src/components/config/App.tsx
```diff
@@ -3,7 +3,7 @@ import { Switch } from 'react-router-dom';
 import MainLayout from '../layout/Main';
 import LandingPage from '../../views/Landing';
 import NotFound from '../../views/NotFound';
-import { CustomRoute } from './Route';
```
alankbi, line 6: Thanks for taking care of these ESLint warnings :)

frontend/src/views/Join.tsx
```diff
@@ -1,16 +1,27 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { PrimaryButton } from '../components/core/Button';
-import { connect, disconnect } from '../api/Socket';
+import Input from '../components/core/Input';
+import { connect, disconnect, sendUser } from '../api/Socket';
+
+// Create constants for the subscription and send message URLs.
+const SOCKET_ENDPOINT:string = '/api/v1/socket/join-room-endpoint';
```
alankbi, line 7: Could this go in the `Socket.ts` file to have a centralized hub for URL paths? That'd mean an extra import and possible style questions for exporting constants so we can also revisit later.

frontend/src/views/Join.tsx
```diff
@@ -1,16 +1,27 @@
...
-      <PrimaryButton onClick={() => connect('/join-room-endpoint')}>
+      <Input placeholder="Enter your nickname" onChange={(event) => setNickname(event.target.value)} />
+      <PrimaryButton onClick={() => connect(SOCKET_ENDPOINT, nickname)}>
         Connect
       </PrimaryButton>
       <PrimaryButton onClick={disconnect}>
         Disconnect
       </PrimaryButton>
+      <PrimaryButton onClick={() => sendUser(nickname)}>
+        Send User Information
```
alankbi, line 23: This all looks great on the UI 👍

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.springframework.messaging.handler.annotation.MessageMapping;
+import org.springframework.messaging.handler.annotation.SendTo;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class UserSocketController extends BaseRestController {
+
+    @MessageMapping("/user")
+    @SendTo(BaseRestController.BASE_SOCKET_URL + "/subscribe-user")
```
alankbi, line 13: Question: I thought the config in `WebSocketConfig` meant you didn't need to specify the base path, but it seems like it's still necessary for the `@SendTo` annotations? I noticed the same thing in the `registerStompEndpoints` method as well; could you clarify when and where the effects of `configureMessageBroker` take place?

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
```diff
@@ -0,0 +1,19 @@
...
+    public User user(String nickname) {
+        User user = new User();
+        user.setNickname(nickname);
+    	return user;
+    }
```
alankbi, line 18: Optional nit: Fix indent, and create a DTO/response/request class(es) for User. That comes with a lot of boilerplate though (extra classes, mappers, service class, tests) so I think we can also hold off on this until it gets a bit more complex.

src/test/java/com/rocketden/main/UserIntegrationTests.java
```diff
@@ -0,0 +1,134 @@
...
+        // Create variables to model message transport, the SockJS client, and the STOMP client.
+        List<Transport> transports = new ArrayList<>();
+        transports.add(new WebSocketTransport(new StandardWebSocketClient()));
+        this.sockJsClient = new SockJsClient(transports);
+        this.stompClient = new WebSocketStompClient(sockJsClient);
+        this.stompClient.setMessageConverter(new MappingJackson2MessageConverter());
+    }
+
+    @Test
+    public void getUser() throws Exception {
```
alankbi, line 56: It is possible to extract some of these overridden classes into separate methods/utility files (bonus if they can then be reused for other socket tests)? If not, could you add a summary comment at the top explaining what this test does?

alankbi (changes requested):
Looks great, I mainly just have a couple of nits, optional changes, and questions.

frontend/src/api/Socket.ts
zpChris: Yeah that makes sense -- I added these `console.error(...)` messages because the UI that is currently setup in this PR will be removed (we won't have "Connect", "Disconnect", and "Send User Information" buttons). I'll be sure to update these to show the relevant information in the UI when or if necessary; I will note that after I update the UI, the user will not be able to connect to the socket when already connected or vice versa, so in that case I would leave this as-is as no UI error message would be needed. I'll address that again when the time comes, though.

frontend/src/components/config/App.tsx
zpChris: No problem 🙂

frontend/src/api/Socket.ts
zpChris: Ah, yeah that's smart -- it worked, thank you!

frontend/src/api/Socket.ts
zpChris: In the line `stompClient.subscribe(SUBSCRIBE_URL, (user:any) => {`, `user` **is** an object; however, this object is generated by STOMP protocol. You may be referring to `user.body`, which **is** a string; as far as I can tell, the `user` object is generated by the STOMP protocol, and it places the string JSON of the object send from the backend into `user.body` -- so I'm not sure if it is possible to return this as an object. Is there a reason you'd want this to be an object (is it just so we don't have to have the `JSON.parse(user.body)` method, or is there another reason)?

Here is a screenshot of the `user` object in the console with the nickname `"Anonymous"`:
User Object in Console

I'd also note that this is the format that the Spring tutorial used, so I am basing this off of something.

src/main/java/com/rocketden/main/model/User.java
```diff
@@ -0,0 +1,10 @@
+package com.rocketden.main.model;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class User {
+    private String nickname;
```
helenli8181, line 9: Is a nickname the only thing we need to know about a user? Should we also save some other unique identifier like a machine IP address or a randomly generated ID?

frontend/src/api/Socket.ts
zpChris: I didn't realize how to do that before -- I was able to find the relevant types for looking at the files from the `import` statements.

I also added this to the rules in `.eslintrc.json`:
```
  "no-unused-vars": "off",
  "@typescript-eslint/no-unused-vars": [
    "error"
  ],
```
This (found at [this Stack Overflow post](https://stackoverflow.com/a/61555310/7517518)) change fixes the false positives for the `no-unused-vars` warnings.

frontend/src/views/Join.tsx
zpChris: Thank you! I'll be removing it in the next PR, but much appreciated 👍

frontend/src/views/Join.tsx
zpChris: No, I think that makes sense; I added the following line to `Socket.ts`:
```
const SOCKET_ENDPOINT:string = '/api/v1/socket/join-room-endpoint';
```
I also added the relevant `import` to `Join.tsx` (this file).

src/main/java/com/rocketden/main/model/User.java
zpChris: There is more information we will want in the future; namely, we will include an auto-generated **ID** when the User is placed into the database, and we will associate a **color** with each User in the backend. I didn't include these things in this PR because I wouldn't have used them, and I didn't want to include unused fields / variables.

For now, I don't see any need for storing IP addresses -- putting aside potential legal issues, when we want to identify users we can add accounts rather than tracking them through IP addresses (especially for the MVP, this is not needed).

src/main/java/com/rocketden/main/controller/v1/BaseRestController.java
zpChris: Making it a `final` class is a good idea, but because other classes extend `BaseRestController` (see `UserSocketController extends BaseRestController`), so if this class was `final` then the rest of the classes would have to be `final` as well (which is not correct).

As for the constructor, I think you're right that an empty constructor actually works great! I didn't realize this the first time because I had included an error that is thrown in the constructor as the **sonarlint** description / example had shown, but this prevented `./mvnw spring-boot:run` from running properly. I had to make the constructor `protected`, as otherwise the classes that extended `BaseRestController` threw the following error:
Private Constructor Error
All in all, I added the lines:
```
/** 
  * Protected empty constructor to allow visability from extended classes,
  * but also remove warning on Utility class having no constructor to hide
  * the implicit public one.
  */ 
protected BaseRestController() {}
```
Thanks for the comment!

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
zpChris: I fixed the indentation, thanks for the note. 👍

As for creating the DTO/response/request classes for the User, I think I'll add that on future PRs, as it's not as necessary for this one IMO.

src/test/java/com/rocketden/main/UserIntegrationTests.java
zpChris: Right now, I'm hesitant to separate those overridden classes into separate files because this is the only test that uses them. If it's OK, I'm thinking we can move these methods into utility files when multiple tests use them; in the meantime however, I added a summary comment at the top explaining what the test does 🙂

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
zpChris: That's a good question; the Spring tutorial gives the best example, I've copied and pasted the relevant material below:

Code:
```
@MessageMapping("/hello")
@SendTo("/topic/greetings")
public Greeting greeting(HelloMessage message) throws Exception {
    Thread.sleep(1000); // simulated delay
    return new Greeting("Hello, " + HtmlUtils.htmlEscape(message.getName()) + "!");
}
```
```
@Override
public void configureMessageBroker(MessageBrokerRegistry config) {
    config.enableSimpleBroker("/topic");
    config.setApplicationDestinationPrefixes("/app");
}
```

Explanation:
> The `configureMessageBroker()` method implements the default method in `WebSocketMessageBrokerConfigurer` to configure the message broker. It starts by calling `enableSimpleBroker()` to enable a simple memory-based message broker to carry the greeting messages back to the client on destinations prefixed with `/topic`. It also designates the `/app` prefix for messages that are bound for methods annotated with `@MessageMapping`. This prefix will be used to define all the message mappings. For example, `/app/hello` is the endpoint that the `GreetingController.greeting()` method is mapped to handle.

All of this is to say that I think, as far as I can understand, it's just syntax that the prefix for the `@MessageMapping(...)` value _doesn't_ include the prefix while the `@SendTo(...)` value _does_ include the prefix.

zpChris:
I addressed the comments and nits above; when I did, the remote tests started failing. I fixed this by pulling the master branch locally, and when I did, I saw errors when running locally (the same that I saw remote). I had to revert the `FrontendRedirectController` back to its original code before PR #25; this fixed all errors.

In the future, I think the way to fix these errors is to pull the updates from master and solve whatever error shows up locally.

frontend/.eslintrc.json
```diff
@@ -20,6 +20,10 @@
         "@typescript-eslint"
     ],
     "rules": {
+        "no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": [
+          "error"
+        ],
```
helenli8181, line 26: I'm just curious, what's the difference between these two rules, since you set them differently?

helenli8181 (approved):
I think this looks really great! Works as expected :)

frontend/.eslintrc.json
alankbi: Before turning off no unused vars completely, I'd suggest looking at [this](https://stackoverflow.com/questions/57802057/eslint-configuring-no-unused-vars-for-typescript) to see if that might fix it without disabling the rule

alankbi (approved):
Looks good! One suggestion but otherwise approved

frontend/.eslintrc.json
zpChris: This doesn't disable it, it actually just makes it work as expected (see image).
Test Variable to Check No-Unused-Vars
Based on [this Stack Overflow post](https://stackoverflow.com/a/61555310/7517518) and [this referenced Github post](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md), it disables the base rules and sets the ESLint rule instead.

PR #24, opened 2020-09-09
Allow room creation on frontend

alankbi:
### List of Changes:

* Enable sending POST requests to create a room from frontend
* Add generic error handling and loading components
* Provide example code for getting the room from location on the game page

### Notes:

* Since the API calls can return both successes and errors, I had to implement an error [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), which is used on the create page with `isError`
* On the game page, `useEffect` is triggered only when location changes; see [here](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects) for more info on how that works
* For more info on `useHistory` and `useLocation`, see [here](https://reactrouter.com/web/api/Hooks)

### Screencast and Images:

Create new room (when error occurs):
Create page

Game page showing room id (if it exists):

Game page


frontend/src/views/Game.tsx
```diff
@@ -1,10 +1,28 @@
...
+    if (location && location.state && location.state.room) {
+      setRoom(location.state.room);
+    }
+  }, [location]);
+
   return (
     <div>
       Game page
+      Room:
+      {room ? room.roomId : 'No room joined'}
```
zpChris, line 25: nit: Add space to `'No room joined'`, making it `' No room joined'`.

zpChris, line 25: nit: Put these text blocks in some HTML object, like a centered `h1` or something similar. (Also, do you need the "`Game page`" line?)

frontend/src/views/Game.tsx
```diff
@@ -1,10 +1,28 @@
...
+  room: Room,
+}
 
 function GamePage() {
+  const location = useLocation<LocationState>();
+  const [room, setRoom] = useState<Room | null>(null);
+
+  // Called every time location changes
+  useEffect(() => {
+    if (location && location.state && location.state.room) {
```
zpChris, line 16: What causes the `location` to lose its value? I saw that if you refresh the page (incognito, cache-clearing, or regular), it retains the value of `location.state.room`. If you instead load it on a different tab or change URLs then return, the `location.state` value is `null`.

This is fantastic, and likely exactly the behavior we want; however, since I don't understand this variable that well, I wanted to ask -- are there are any edge cases with this we should watch out for?

zpChris: Building on this, I was going to ask whether we should store URLs on the game page as `/game?roomId=XXXXXX`. It doesn't seem like that's needed here, but it may help with sharing the URL / code easily, as otherwise the other users have no other option but to type it out. Thoughts?

frontend/src/views/Create.tsx
```diff
@@ -1,9 +1,40 @@
...
+  }, [history]);
+
   return (
     <div>
-      Create game page
+      {error ? <ErrorMessage message={error} /> : null}
+      {loading ? <Loading /> : null}
+      <PrimaryButton onClick={createNewRoom}>
+        Create New Room
+      </PrimaryButton>
```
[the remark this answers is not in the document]
zpChris: If you did this, you would likely update the design slightly on this page to make the loading or error text bigger (as there may be nothing else on this page). Or, you could just not have this page at all, and display loading and error messages on the homepage. I'm open to any of these ideas -- thoughts?

frontend/src/views/Create.tsx
```diff
@@ -1,9 +1,40 @@
...
+    setLoading(true);
+    createRoom()
+      .then((res) => {
+        setLoading(false);
+        if (isError(res)) {
+          setError(res.message);
+        } else {
+          redirectToGame(res as Room);
+        }
+      });
```
zpChris, line 28: Just to clarify: This is the type guard, where an error will be set if the `error` field is present on `res`?

A comment may be helpful here.

frontend/src/api/Room.ts
```diff
@@ -0,0 +1,17 @@
...
+  roomId: string;
+};
+
+const basePath = '/api/v1/rooms';
+const routes = {
+  createRoom: `${basePath}/`,
+  joinRoom: `${basePath}/`,
+};
+
+export const createRoom = (): Promise<Room | ErrorResponse> => axios.post<Room>(routes.createRoom)
```
zpChris, line 15: Should we be using `axios` instead of `fetch(...)` in the `Problem.ts` file? I know you didn't change that here, but it may be good to note to change that in the future if we want to maintain consistency.

frontend/src/api/Error.ts
```diff
@@ -0,0 +1,30 @@
...
+    status: string,
+    data: any,
+  },
+};
+
+export const errorHandler = (err: AxiosError): ErrorResponse => {
+  if (!err.response) {
+    return {
+      error: true,
+      status: 'XXX',
```
zpChris, line 18: Does this `XXX` value hold any meaning -- if initiating the REST request failed, is there any specific error status associated with that? If not, no problem.

zpChris (changes requested):
I left lots of comments, but almost all of them are nits or small questions. The only real change that made me "Request Changes" rather than "Approve" right away is that I think the `/game/create` page may be largely unnecessary, as you could put the code on `Create.tsx` essentially directly on the landing page. Let me know what your thoughts are on that; otherwise, this looks great!

frontend/src/views/Game.tsx
helenli8181: I'm not sure having unique URLs for rooms would provide tangible benefit. To share it you could send just the code, and others can copy/paste that into the "join room" box without typing it.

frontend/src/views/Game.tsx
helenli8181: I deleted `Game Page` and replaced it with the header bar in my version of this file. If we want to delete it, we could do that when we have to merge the two versions.

frontend/src/views/Game.tsx
```diff
@@ -1,10 +1,28 @@
 import Editor from '@monaco-editor/react';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
+import { Room } from '../api/Room';
```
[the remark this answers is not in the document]
zpChris: I think I may have fixed this in my PR #23 -- see [this discussion](https://github.com/rocketden/main/pull/23#discussion_r486071349).

helenli8181 (approved):
Generally this looks great to me, I just had some minor comments!

frontend/src/views/Game.tsx
zpChris: That's true, I still think it would be easy though to just share the URL -- that's copying only one thing vs copying two. Plus, in this setup I think it would help to provide both options (there isn't any choice overload that happens here, so whichever method people prefer can be used). I guess overall, it's up to @alankbi.

alankbi: I think that's something that might make sense to implement on the join page; the way I imagined it, the host on the waiting room page can copy a share link such as "/game/join?id=104020" and share that with others, and then when people navigate to that link, it'd basically be the join page but skipping past the enter room id step straight to the enter nickname step. But with all that, I added a [card](https://trello.com/c/SlNjdzCe/101-join-page-takes-query-param-with-roomid-for-quick-sharing) to the backlog so that we can discuss this further.

frontend/src/views/Create.tsx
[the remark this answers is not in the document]
alankbi: I got rid of the page so now create new room on the landing page just takes you directly to the game page. At some point though we might need to re-add the create page though, or maybe not; I can also imagine that creating a new game would just take you to the waiting room and that's where hosts might be able to customize their rooms.

frontend/src/views/Game.tsx
alankbi: I deleted it anyways since it only takes a second :)

frontend/src/views/Game.tsx
alankbi: To answer the first question, this is a quote from the documentation: 

>The useLocation hook returns the location object that represents the current URL. You can think about it like a useState that returns a new location whenever the URL changes.

So I guess refreshing doesn't change the URL but if you go forward/back and all that then it does?

frontend/src/api/Error.ts
alankbi: I think it's when the request doesn't initiate in the first place so no status nor data is returned, hence the 'XXX' as a dummy value. I added a comment for that but can also replace it w/ something else like '000' or '0' or make it an optional param; open to ideas.

frontend/src/api/Room.ts
alankbi: Pushed a commit to change that since it should be a pretty simple fix.

frontend/src/views/Create.tsx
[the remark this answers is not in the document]
zpChris: > I can also imagine that creating a new game would just take you to the waiting room and that's where hosts might be able to customize their rooms.
That's what I was thinking; I think it would reduce the barrier to starting the game, which is what we want. This could also allow participants to see or edit the game settings themselves.

alankbi:
Addressed changes, and also wanted to note that when testing the room POST request locally, I came across the 45s timeout error that we got back when we were trying to figure the whole `max_user_connections` thing (although it went away on second try).

frontend/src/views/Game.tsx
zpChris: Thanks for the answer to the `useLocation` hook. As for this:
> it'd basically be the join page but skipping past the enter room id step straight to the enter nickname step.

That's how I imagined it too, so it let's users join the game more quickly.

frontend/src/api/Error.ts
zpChris: I think that's fine, 'XXX' as a dummy value makes sense.

zpChris:
[Frontend room creation process for games](https://trello.com/c/u2xdh6Dh/86-frontend-room-creation-process-for-games)

frontend/src/api/Room.ts
zpChris: Awesome, thank you!

frontend/src/views/Game.tsx
```diff
@@ -1,10 +1,28 @@
...
+      setRoom(location.state.room);
+    }
+  }, [location]);
+
   return (
     <div>
-      Game page
+      Room:
+      {' '}
+      {room ? room.roomId : 'No room joined'}
```
zpChris, line 25: When this is combined with @helenli8181's PR #22 in `master`, this will look weird; imo, whoever's PR is merged second should try to account for this. This likely means @helenli8181 will make another change in her PR to account for making this an OK design.

frontend/src/views/Landing.tsx
```diff
@@ -12,9 +42,11 @@ function LandingPage() {
         Join a Game
       </PrimaryButtonLink>
       <br />
-      <TextLink to="/game/create">
+      <TextButton onClick={createNewRoom}>
         Or create a room &#8594;
-      </TextLink>
+      </TextButton>
+      {error ? <ErrorMessage message={error} /> : null}
+      {loading ? <Loading /> : null}
```
zpChris, line 49: Quick note on this -- I tested what would happen if I turned the wifi off and tried to **create a room**. It loaded for 45 seconds and then displayed the error message (which is expected). I think this is a good setup, but I wanted to note it down in case you would prefer to do anything else (such as a shorter "end request" time than 45 seconds).

zpChris (approved):
LGTM! Just left two comments, but neither require you to change anything. Thanks for making those changes, looks great!

frontend/src/views/Landing.tsx
alankbi: Personally I think that should be fine based on testing what LeetCode does in response to no internet (also shows loading for a long time), but I'll add a card to the backlog.

PR #29, opened 2020-09-14
Manage waiting room users

zpChris:
### List of Changes:
- Add methods in `UserSocketController.java` to add and delete user from a `Set<User>` object (not connected to the database, reset after relaunching the app); the subscription URL now sends a `Set<User>` of the current users in the waiting room.
  - Add and update the relevant tests.
- Add methods in `Socket.ts` to call those in `UserSocketController.java`.
- Create a `UserNicknameText` styled component to hold each user in the frontend. Click this to delete the user.
- Add frontend error handling.

### Notes:
- This branch is set to merge into `master`, not `UserRoomDatabase`. I decided to create a faux backend where the users are simply stored in a `Set<User>` object, despite the fact that my previous PR #27 created the basic setup for adding users to the database.
- I removed the `console.error(...)` messages, and added frontend UI error handling (right now, just for the `connect(...)` method).
- Things to handle in the future:
  - Not allow user to join if the nickname is already taken.
  - Remove the user from the room if they delete their tab.
  - Preserve the current page state if the user refreshes the page.
  - Not allow users to remove themselves by deletion and still stay on the tab. In the future, I think I should only allow the host to remove users.
  - Possibly show the host a different page layout than the other users.
- Later in this sprint, I will work to combine this PR with #27, so that the User is added to a predefined Room in the database after joining. All updates to the Room's users could then be in that database. **Important related note:** @alankbi, we should determine what is going in the backend vs. database soon, IMO.
- The current testing setup only allows testing one "send" method at a time. I tried to use [this tutorial's setup](http://rafaelhz.github.io/testing-websockets/) which allows testing the response from specific "send" methods, but I ran into issues. I posted a [question on Stack Overflow](https://stackoverflow.com/questions/63885922/java-spring-testing-multiple-send-messages-websocket) -- I'm not sure what to do here if nobody responds.

### Screencast and Images:

[Screencast of the Waiting Room UI](https://www.loom.com/share/c31ad73e70cc477dbf48dea24b371699)

Waiting Room with Two Users (Chris and John):
Waiting Room with Two Users

Error with Connecting to Socket on UI:
Error with Connecting to Socket on UI


zpChris:
[Add user to backend on joining waiting room](https://trello.com/c/MWWj8Jdo/120-add-user-to-backend-on-joining-waiting-room)

frontend/src/api/Socket.ts
```diff
@@ -1,54 +1,92 @@
...
 const SUBSCRIBE_URL:string = '/api/v1/socket/subscribe-user';
-const SEND_USER_URL:string = '/api/v1/socket/user';
+const ADD_USER_URL:string = '/api/v1/socket/add-user';
+const DELETE_USER_URL:string = '/api/v1/socket/delete-user';
 
-export const sendUser = (nickname:string) => {
-  if (connected) {
-    stompClient.send(SEND_USER_URL, {}, nickname || 'Anonymous');
+export const isValidNickname = (nickname: string) => nickname.length > 0
+  && !nickname.includes(' ') && nickname.length <= 16;
```
helenli8181, line 21: Should there be a check to make sure multiple people don't share the same username in a game? You used a HashSet to store the users but it doesn't appear to do anything if the nickname already exists.

frontend/src/views/Join.tsx
```diff
@@ -80,6 +96,18 @@ function JoinGamePage() {
             {nickname}
             &quot;.
           </LargeText>
+          <div>
+            {
+              users.map((user) => (
+                <UserNicknameText onClick={(event) => {
+                  deleteUser((event.target as HTMLElement).innerText);
+                }}
+                >
```
helenli8181, line 105: Nit: is it normal to put the angle bracket on a separate line when it has the same indentation as the closing brackets above it?

helenli8181 (commented):
I just have a few questions but otherwise this works great!

frontend/src/api/Socket.ts
zpChris: See the Notes section:
> - Things to handle in the future:
>   - Not allow user to join if the nickname is already taken.

Since it would require restructuring, I think I was planning on waiting on this, especially because we don't have separate rooms set up right now.

frontend/src/api/Socket.ts
```diff
@@ -1,54 +1,92 @@
...
+        });
+        // Reassign connected variable.
+        connected = true;
+        addUser(nickname);
+        resolve();
+      }, () => {
+        reject(new Error('The socket failed to connect.'));
+      });
+    } else {
+      reject(new Error('The socket is already connected.'));
```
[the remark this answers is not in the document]
zpChris: Two notes:
- I no longer have "connect" or "disconnect" buttons -- those are from an old PR. The "Enter" button on the "Enter Nickname" page will trigger the `connect(...)` function, but there will be no button that triggers `disconnect(...)` -- this will be triggered when the user leaves the page or is booted from the game or whatnot.
- This error handling is used even to check for cases that currently don't seem possible. Part of the point of error handling is to check for unexpected cases -- for example, so if something goes wrong where the user connects but they aren't transferred the the "Waiting Room" page, they won't re-connect, which will let them know to refresh the page.

frontend/src/views/Join.tsx
zpChris: Yes -- this is actually a required ESLint styling when an inner attribute function is called. (See image below for an example of this error being thrown.)

ESLint Styling Errors

helenli8181 (approved):
Thanks for the clarifications, I think this PR looks good!

frontend/src/api/Socket.ts
```diff
@@ -1,54 +1,92 @@
...
+export const addUser = (nickname:string): Error | undefined => {
+  if (connected && isValidNickname(nickname)) {
+    stompClient.send(ADD_USER_URL, {}, nickname);
+  } else if (!connected) {
+    return new Error('The socket is not connected.');
   } else {
-    console.error('You must be connected to a socket before sending user information.');
+    return new Error('The provided nickname is invalid.');
   }
 };
```
alankbi, line 35: Suggestion: For addUser and deleteUser, I'm not sure if you actually need nickname error validation since you already validate it in `Join.tsx`. Although assuming that in the future we add server-side verification on inputs, then maybe we can keep the error validation here but instead of validating and returning a custom Error, just returning the ErrorResponse received from the server.

frontend/src/api/Socket.ts
```diff
@@ -1,54 +1,92 @@
...
 
-export const disconnect = () => {
+/**
+ * Connect and subscribe the user via socket.
+ * @returns error, if present
+*/
+export const connect = (endpoint:string, nickname:string,
+  setUsers: React.Dispatch<React.SetStateAction<User[]>>):
+  Promise<Error | undefined> => new Promise<Error | undefined>((resolve, reject) => {
```
alankbi, line 57: For consistency with how you do it in the above functions, are you able to just resolve the Error object instead of rejecting? That way we don't have to deal with catch blocks in our React files, and then you might also be able to simplify this line to `Promise<Error | undefined>`

frontend/src/api/Socket.ts
```diff
@@ -1,54 +1,92 @@
...
   }
 };
 
-export const disconnect = () => {
+/**
+ * Connect and subscribe the user via socket.
+ * @returns error, if present
+*/
+export const connect = (endpoint:string, nickname:string,
+  setUsers: React.Dispatch<React.SetStateAction<User[]>>):
```
alankbi, line 56: Not too sure about the recommended practice of passing a React set state function. Instead of this, is it possible to return the users and call setUsers in Join.tsx in a then block?

frontend/src/views/Join.tsx
```diff
@@ -49,15 +58,21 @@ function JoinGamePage() {
             }}
             onKeyPress={(event) => {
               if (event.key === 'Enter' && validNickname) {
-                connect(SOCKET_ENDPOINT, nickname);
-                setPageState(2);
+                connect(SOCKET_ENDPOINT, nickname, setUsers).then(() => {
+                  setPageState(2);
+                }).catch((response) => {
+                  setError(response.message);
```
alankbi, line 64: Building on my previous comment, if you decide to implement that change, then this might look something like `if isError(res) then setError((res as ErrorResponse).message) else setPageState(2)`

frontend/src/views/Join.tsx
```diff
@@ -49,15 +58,21 @@ function JoinGamePage() {
...
           />
           <LargeInputButton
             onClick={() => {
-              connect(SOCKET_ENDPOINT, nickname);
-              setPageState(2);
+              connect(SOCKET_ENDPOINT, nickname, setUsers).then(() => {
+                setPageState(2);
+              }).catch((response) => {
+                setError(response.message);
+              });
```
alankbi, line 75: nit: Extract into a helper function since it seems like this is a duplicated code block

src/test/java/com/rocketden/main/UserIntegrationTests.java
```diff
@@ -96,7 +98,62 @@ public void handleFrame(StompHeaders headers, Object payload) {
...
+                        } catch (Throwable t) {
+                            failure.set(t);
+                        } finally {
+                            session.disconnect();
+                            latch.countDown();
+                        }
+                    }
+                });
+                try {
+                    session.send(BaseRestController.BASE_SOCKET_URL + "/delete-user", "Chris");
```
alankbi, line 156: Is Chris already in the HashSet when you run this or is this just making sure removing a nonexistent user doesn't cause an error?

alankbi (changes requested):
UI looks very pretty as always :) Left some suggestions here and there, lmk what your thoughts are about those!

frontend/src/api/Socket.ts
zpChris: Yes, my thinking was that somebody could enable the frontend "Enter" button through inspect element even though the current nickname is invalid; in that case, we would still want to catch the case and return an error.

I'm happy to add validation on the server side as well, that makes sense; however, if I keep the validation here as well, what should I do if the nickname is not valid? Should I return some sort of error in that case, or should I just not have this validation at all?

frontend/src/api/Socket.ts
zpChris: Hmm ok, I'm sorry I didn't realize that before; I'd want to discuss maybe using `catch` blocks instead? Wouldn't it be smarter to use that built-in block rather than calling `isError(...)`?

I'll also note that I used `catch(...)` in a branch on top of this as well -- not that it matters, but just I think it might be smart to switch to that? (See [catch block from next PR](https://github.com/rocketden/main/blob/ed5b3253a252e58548e0bcfb8c9ed50c55feedbc/frontend/src/views/Join.tsx#L48).)

frontend/src/api/Socket.ts
zpChris: I actually do this in the next PR! So I'm happy to do it in this as well, I agree that is better practice. I will note though that in my next PR I add more `setX` functions, and I'm not sure how to structure it otherwise -- see [here](https://github.com/rocketden/main/blob/ed5b3253a252e58548e0bcfb8c9ed50c55feedbc/frontend/src/views/Join.tsx#L42). We can talk about this during our standup.

src/test/java/com/rocketden/main/UserIntegrationTests.java
zpChris: I'm removing a nonexistent error. I noted in the "Notes" section of the PR description that I had trouble creating two "send" functions on the same `@Test`, so I'm not sure how to test actually removing a user (or any sequence of socket send operations).

frontend/src/api/Socket.ts
zpChris: After our conversation, I removed the secondary check on the frontend -- done!

frontend/src/api/Socket.ts
zpChris: After our conversation, I decided to switch the codebase to use `catch(...)`. I am updating the old functions as well right now.

frontend/src/api/Socket.ts
zpChris: After our standup, I have replaced this with a callback that is triggered every time the user's subscription receives a message; thus, I don't need to pass the any `setX` function.

frontend/src/api/Socket.ts
zpChris: The old functions are updated!

frontend/src/views/Join.tsx
zpChris: Good idea -- done!

zpChris:
Re-requested review from @alankbi. A couple notes:
- We may want to consider adding more documentation to these methods (explaining what each of the parameters are and whatnot). I think we may be reaching the point where, since I'm the one building on top of this code, it becomes increasingly incomprehensible as only I understand the mechanics. We may want to try to clean it up if it seems convoluted, or else one of you take up stories on this as well.
- We are reaching a "Pyramid of Doom" situation with all of the `connect`, `subscribe`, and `addUser` functions (I separated these from the `Socket.ts` class as they seem like separate functions that shouldn't necessarily be grouped).
- I considered using `async` / `await` for the `Socket.ts` functions to prevent requiring a bunch of `catch` calls (and just replace it with a single `try` / `catch`), but I figured the current setup is fine.

alankbi (approved):
Approved! Some comments:

* I agree on distributing some of the socket work, and I also think async/await is something worthwhile to look into (I added it to the backlog). 
* Pyramid of doom - love that term haha, maybe this is something that async/await can help with? Not too sure
* Finally, I think the documentation, separation of functions in Socket.ts, and the error handling looks _amazing_ :) Makes it really easy to follow the flow and understand what's going on so major props

src/test/java/com/rocketden/main/UserIntegrationTests.java
zpChris: I added issue #33 for this. I'll merge this PR, but we should address this at some point.

PR #30, opened 2020-09-14
Backend error handling

alankbi:
### List of Changes:

* Add exception folder containing API error classes
* Move messages from DTO classes into API error classes
* Remove if statements handling errors in controller classes
* Change error handling functionality to throwing exceptions
* Update tests to handle new error functionality

### Notes:

* See [here](https://stackoverflow.com/questions/38117717/what-is-the-best-way-to-return-different-types-of-responseentity-in-spring-mvc-o) for details about this design
* See [here](https://github.com/rocketden/main/commit/b76759c4bda597db83deb58238f649899b8dd906) for our discussion on proposed design


alankbi:
[Improve error response process](https://trello.com/c/IAncsU5Z/94-improve-error-response-process)

src/main/java/com/rocketden/main/service/UserService.java
```diff
@@ -24,35 +26,26 @@ public CreateUserResponse createUser(CreateUserRequest request) {
...
-     * 1. Non-empty
+     * 1. Non-null and non-empty
      * 2. Less than or equal to sixteen characters
      * 3. Contains no spaces
      */
     public static boolean validNickname(String nickname) {
-        return nickname.length() > 0 && nickname.length() <= 16 
-            && !nickname.contains(" ");
+        return nickname != null && nickname.length() > 0
+                && nickname.length() <= 16 && !nickname.contains(" ");
```
helenli8181, line 48: Maybe this has been brought up before, but should valid nicknames be able to include non-alphanumeric characters?

src/main/java/com/rocketden/main/exception/RoomErrors.java
```diff
@@ -0,0 +1,16 @@
+package com.rocketden.main.exception;
+
+import com.rocketden.main.exception.api.ApiError;
+import org.springframework.http.HttpStatus;
+
+public class RoomErrors {
```
helenli8181, line 6: What do you think about combining the RoomErrors and UserErrors? Right now we only have these two types of errors but in the future there could be more categories, and I think it would be more efficient to keep all of them together as a "catalogue" of sorts.

src/main/java/com/rocketden/main/exception/UserErrors.java
```diff
@@ -0,0 +1,10 @@
+package com.rocketden.main.exception;
+
+import com.rocketden.main.exception.api.ApiError;
+import org.springframework.http.HttpStatus;
+
+public class UserErrors {
+
+    public final static ApiError INVALID_USER = new ApiError(HttpStatus.BAD_REQUEST,
```
helenli8181, line 8: I think it might be a good idea to create our own custom error codes, because neither HTTP status codes nor the string message make good unique identifiers. This doesn't make a difference when we are just displaying the error message to the web page, but it would be necessary if we want the client side to do something in response to the error. 
Maybe something like this:
`ApiError INVALID_USER = new ApiError(HttpStatus.BadRequest, "INVALID_USER", "Please provide a user...")`

src/main/java/com/rocketden/main/exception/UserErrors.java
```diff
@@ -0,0 +1,10 @@
...
+            "Please provide a user with a valid nickname (1-16 characters without spaces).");
```
helenli8181, line 9: Related to my previous comments, alternatively we could move these error messages out to some sort of reference table or dictionary keyed with custom unique codes instead of having them all hard coded in various places. There's a couple of error messages on the front end files that could benefit from more organization too.

helenli8181 (changes requested):
Overall I think this new organization is great; it was very satisfying to see large chunks of code being moved out to new classes haha. I'm just wondering what you guys think about a slightly different way to organize the error fields.

src/main/java/com/rocketden/main/exception/RoomErrors.java
alankbi: Hm, I had the same thought but different conclusion: as we get more types of errors, I was thinking it'd be helpful to have these separate classes to better distinguish errors (e.g. `RoomErrors.NOT_FOUND` and `UserErrors.NOT_FOUND`) and make it easier to refactor if we ever want to split/change the codebase. However, I can also be flexible on this, especially since right now we only have a few errors so it's not too hard to change. @zpChris any thoughts?

src/main/java/com/rocketden/main/exception/UserErrors.java
alankbi: I think that's a good idea - any thoughts on what error codes we would have for some of the existing errors? For example, some errors don't require any action on the client besides displaying it to the user (e.g. room not found, invalid user) so I'm not sure what error code they should have (maybe like a default "NONE" or "DISPLAY" or something). Basically want to make sure that the labels convey useful info to the client and don't end up being redundant.

Another thought I just had which is kind of a variant of yours is if instead of error codes, it returns the type of error to be displayed, so for example right now we only show errors as red text, but maybe different types of errors warrant a popup or a page refresh. This might be kinda awkward to implement on the frontend though.

src/main/java/com/rocketden/main/exception/RoomErrors.java
zpChris: I think that separate classes is preferable; it let's us easily know which error is attributable to which class, and enables more modularization in the future.

Note: I tried to find if Spectrum's GitHub had any examples to base this decision off of, but I didn't find anything. If you search "exception" on [this Spring GitHub project](https://github.com/spring-projects/greenhouse/find/master), they use separate classes (I know exception and error are different though, so my if you think this is irrelevant that's fine).

src/main/java/com/rocketden/main/exception/UserErrors.java
alankbi: I talked about this with Chris in an above comment where a HashMap was brought up (although not sure if that's what you were referring to when you said dictionary), but I think we went against it because that would mean extra overhead plus it would decentralize the errors (forcing the programmer to manually look up a message from the map when throwing an error). So I guess in a sense this file is our centralized "reference table" for errors, although there definitely could be a better way of doing it - did you have anything in mind for this, like Java enums or some other useful APIs?

src/main/java/com/rocketden/main/exception/UserErrors.java
alankbi: [This](https://www.baeldung.com/rest-api-error-handling-best-practices) provides some possible other fields we could return as part of our errors in case we think any of them are useful.

src/main/java/com/rocketden/main/exception/UserErrors.java
alankbi: Maybe like moving the errors into a JSON or other data file of some sort? Especially as we start to get more and more errors, then maybe that might make it easier to add and edit them.

alankbi:
Ended up going the Enum route as we discussed during our calls. This has the added benefit of passing the type of error (i.e simply the variable name) to the client.

src/main/java/com/rocketden/main/exception/api/ApiErrorResponse.java
```diff
@@ -3,13 +3,15 @@
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 
-// Response object returned to client containing message
+// Response object returned to client containing message and error type (name of enum var)
 @Getter
 @EqualsAndHashCode
 public class ApiErrorResponse {
     private final String message;
+    private final String type;
```
helenli8181: Nit: should error type go before message?

helenli8181: Nit: should error type go before the message?

helenli8181 (approved):
This looks great to me!

src/main/java/com/rocketden/main/exception/UserError.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Getter
+public enum UserError implements ApiError {
+
+    INVALID_USER(HttpStatus.BAD_REQUEST, "Please provide a user with a valid nickname (1-16 characters without spaces).");
+
+    private final HttpStatus status;
+    private final ApiErrorResponse response;
+
+    UserError(HttpStatus status, String message) {
```
zpChris, line 16: Can you add an access level (`private`, `protected`, or `public`) here? (The same applies for `RoomError`.)

src/main/java/com/rocketden/main/exception/api/ApiError.java
```diff
@@ -1,22 +1,13 @@
...
-    public ApiError(HttpStatus status, String message) {
-        this.status = status;
-        this.response = new ApiErrorResponse(message);
-    }
+    HttpStatus getStatus();
 
-    public String getMessage() {
-        return this.response.getMessage();
-    }
+    ApiErrorResponse getResponse();
```
zpChris, line 10: I don't think this method is ever used? Thoughts on implementing it in `RoomError` or `UserError`, or else removing it?

zpChris: If you do this, could you test it somewhere as well?

zpChris (changes requested):
This looks amazing! I have just three comments, essentially nits, but I requested changes just because I want to confirm the new implementation works for me as well.

I love the new setup 💯

src/main/java/com/rocketden/main/exception/UserError.java
alankbi: Tried adding an access level but public/protected is not allowed and private gives this message

Enum private is redundant

src/main/java/com/rocketden/main/exception/api/ApiError.java
alankbi: Yeah it's kinda unclear since RoomError and UserError don't actually seem to "implement" it, but rather it's auto-generated by the `@Getter` notation (it won't compile if `@Getter` is removed as expected if you inherit an interface without implementing all the methods). This is where the getResponse method is used:

Where getResponse is used in codebase

src/main/java/com/rocketden/main/exception/UserError.java
zpChris: Ah ok sorry about that, still learning about `enum`s -- thanks for the clarification there.

src/main/java/com/rocketden/main/exception/api/ApiError.java
zpChris: Ok that makes sense, thanks for the explanation. 👍

zpChris (approved):
LGTM! Thanks for the clarifications 👍 😃

src/main/java/com/rocketden/main/exception/UserError.java
alankbi: Yeah same haha, when I first was implementing it I tried to make it public which obviously didn't work out

src/main/java/com/rocketden/main/exception/api/ApiErrorResponse.java
alankbi: Sorry, just realized I totally forgot about this. It looks like by default, the JSON responses from these types of classes are unordered; however, there is a way to make them ordered using the annotation explained [here](https://stackoverflow.com/questions/9610948/json-fields-unordered). I added a backlog card in case we want to do this for all of our response objects.

PR #37, opened 2020-09-21
Refactor to use RoomDto and UserDto

alankbi:
### List of Changes:

* Convert CreateRoomResponse and JoinRoomResponse objects to RoomDto
* Convert usages of User and CreateUserResponse in client-related classes to UserDto
* Refactor mapper methods
* Fix test cases to use new DTO objects

### Notes:

* I'm kind of following [this](https://stackoverflow.com/questions/53966427/should-service-layer-accept-a-dto-or-a-custom-request-object-from-the-controller) post in terms of design (having a separate Request object but the same DTO for returns)
* Lots of files changed, but it's mainly just refactoring


alankbi:
[Add RoomDto and UserDto objects](https://trello.com/c/VZVV9MyK/118-add-roomdto-and-userdto-objects)

src/main/java/com/rocketden/main/dto/room/RoomMapper.java
```diff
@@ -9,12 +9,11 @@ protected RoomMapper() {}
 
     private static final ModelMapper mapper = new ModelMapper();
 
-    public static CreateRoomResponse entityToCreateResponse(Room entity) {
-        return mapper.map(entity, CreateRoomResponse.class);
-    }
-
-    public static JoinRoomResponse entityToJoinResponse(Room entity) {
-        return mapper.map(entity, JoinRoomResponse.class);
+    public static RoomDto toDto(Room entity) {
```
zpChris, line 12: Question: Could we have just one "Mapper" class for all by passing in generics? Would that even be helpful, or is that unnecessary?

src/main/java/com/rocketden/main/dto/room/RoomMapper.java
```diff
@@ -9,12 +9,11 @@ protected RoomMapper() {}
...
+        if (entity == null) {
+            return null;
+        }
+        return mapper.map(entity, RoomDto.class);
     }
 
     public static GetRoomResponse entityToGetResponse(Room entity) {
```
zpChris, line 19: Do you still need this method? Or can you update it to use `RoomDto`?

zpChris: On that note, did you leave in the `GetRoomResponse` file on purpose?

src/test/java/com/rocketden/main/mapper/RoomMapperTests.java
```diff
@@ -11,46 +11,47 @@
...
+                .map(UserMapper::toEntity)
+                .collect(Collectors.toSet());
+        assertEquals(room.getUsers(), actualUsers);
+    }
 
-        assertNotNull(response);
-        assertEquals(room.getRoomId(), response.getRoomId());
-        assertEquals(room.getUsers(), response.getUsers());
+    @Test
+    public void nullRoomMappings() {
```
zpChris, line 54: Is there a reason why you have a `toEntity` method for the `User` class but not for the `Room` class? If that wasn't by design for some reason, thoughts on adding it there and in this test case?

zpChris (approved):
LGTM! I have a couple nits and questions in here, but all of them you should be able to address easily. If it's OK, ping me in the chat once you address all of the comments -- I wanted to approve but just check again so I can understand the PR.

src/main/java/com/rocketden/main/dto/room/RoomMapper.java
alankbi: I think we could (especially at this stage since we don't have that many), but I think the reason for keeping the separate mappers would be for flexibility as things grow more complex and the black-box ModelMapper isn't able to do all of it for us (such as if variable names don't match, or if conversion requires other info such as game data or something)

src/main/java/com/rocketden/main/dto/room/RoomMapper.java
alankbi: Yeah, figured that the tradeoffs were keeping the separate get response means that you don't need to return all the data (users, settings, etc.) if you're just validating that a room exists, while at the same time that would mean you need to have a separate class specifically for the GET endpoint. So I'm okay with doing this either way, just lmk

src/test/java/com/rocketden/main/mapper/RoomMapperTests.java
alankbi: The main reason was the need to convert UserDtos to Users within some of the service classes, whereas that functionality hasn't been needed yet for Rooms (since clients pass in a complete UserDto but never more than just a roomId for room requests). Again just lmk if you want me to add it for consistency as that should be pretty quick and easy to implement

zpChris: No need, thanks for the explanation.

src/main/java/com/rocketden/main/dto/room/RoomMapper.java
zpChris: Ok that makes sense, no need to update -- thanks for the explanation.

src/main/java/com/rocketden/main/dto/room/RoomMapper.java
zpChris: Ok I think that makes sense -- yeah I found that implementing a [generic version](https://github.com/modelmapper/modelmapper/issues/187) actually seems complicated, so this is OK.

helenli8181 (approved):
This all looks good to me!

PR #38, opened 2020-09-22
Create the Waiting Room page and Update the Create and Join pages.

zpChris:
### List of Changes:
- Create the Waiting Room page (this is essentially a variant of the old Join page, but without the `pageState`).
- Update the Create and Join pages as necessary (the Join page now simply redirects to the Waiting Room page on enter).

### Notes:
- The error that shows "The socket failed to connect." is now on the Waiting room page.
- Whenever you refresh on the Waiting room page, it will re-connect and re-add you as a user. In the future (maybe next PR?), we will be able to detect when somebody reloads or closes the tab, which will improve the user management.
- Eventually, if you go to the waiting room without a Room or Nickname, it will redirect the user to the appropriate Join or Create page.

### Screencast and Images:
There are no images as there are no major UI changes, and all of the page flow updates can be shown through the screencast below.

[Screencast of the Waiting Room Page](https://www.loom.com/share/8df757f482fa4400999adb439f028528)

### DIFF
--- a/frontend/src/components/config/App.tsx
+++ b/frontend/src/components/config/App.tsx
@@ -9,6 +9,7 @@ import GameLayout from '../layout/Game';
 import JoinGamePage from '../../views/Join';
 import CreateGamePage from '../../views/Create';
 import GameResultsPage from '../../views/Results';
+import WaitingRoomPage from '../../views/Waiting';
 
 function App() {
   return (
@@ -17,6 +18,7 @@ function App() {
       <CustomRoute path="/game" component={GamePage} layout={GameLayout} exact />
       <CustomRoute path="/game/join" component={JoinGamePage} layout={MainLayout} exact />
       <CustomRoute path="/game/create" component={CreateGamePage} layout={MainLayout} exact />
+      <CustomRoute path="/game/waiting" component={WaitingRoomPage} layout={MainLayout} exact />
       <CustomRoute path="/game/results" component={GameResultsPage} layout={GameLayout} exact />
       <CustomRoute path="*" component={NotFound} layout={MainLayout} />
     </Switch>
--- a/frontend/src/components/config/App.tsx
+++ b/frontend/src/components/config/App.tsx
@@ -9,7 +9,7 @@ import GameLayout from '../layout/Game';
 import JoinGamePage from '../../views/Join';
 import CreateGamePage from '../../views/Create';
 import GameResultsPage from '../../views/Results';
-import WaitingRoomPage from '../../views/Waiting';
+import LobbyPage from '../../views/Lobby';
 
 function App() {
   return (
@@ -18,7 +18,7 @@ function App() {
       <CustomRoute path="/game" component={GamePage} layout={GameLayout} exact />
       <CustomRoute path="/game/join" component={JoinGamePage} layout={MainLayout} exact />
       <CustomRoute path="/game/create" component={CreateGamePage} layout={MainLayout} exact />
-      <CustomRoute path="/game/waiting" component={WaitingRoomPage} layout={MainLayout} exact />
+      <CustomRoute path="/game/lobby" component={LobbyPage} layout={MainLayout} exact />
       <CustomRoute path="/game/results" component={GameResultsPage} layout={GameLayout} exact />
       <CustomRoute path="*" component={NotFound} layout={MainLayout} />
     </Switch>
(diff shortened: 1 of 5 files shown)


zpChris:
[Waiting page on game page](https://trello.com/c/YdXEksMw/121-waiting-page-on-game-page)

helenli8181 (approved):
This looks great! I'm not finding any problems with it :)

frontend/src/components/config/App.tsx
```diff
@@ -17,6 +18,7 @@ function App() {
       <CustomRoute path="/game" component={GamePage} layout={GameLayout} exact />
       <CustomRoute path="/game/join" component={JoinGamePage} layout={MainLayout} exact />
       <CustomRoute path="/game/create" component={CreateGamePage} layout={MainLayout} exact />
+      <CustomRoute path="/game/waiting" component={WaitingRoomPage} layout={MainLayout} exact />
```
alankbi, line 21: Personal preference and nit, but what do you think about calling this '/game/lobby' or something like that instead?

frontend/src/views/Waiting.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { useLocation } from 'react-router-dom';
+import { Message } from 'stompjs';
+import { LargeText, UserNicknameText } from '../components/core/Text';
+import {
+  addUser, SUBSCRIBE_URL, connect,
+  deleteUser, SOCKET_ENDPOINT, subscribe, User,
+} from '../api/Socket';
+import ErrorMessage from '../components/core/Error';
+
+type WaitingRoomPageProps = {
```
alankbi, line 11: nit: Maybe rename this to something involving Location, as I think it's being passed to useLocation as opposed to the component props

frontend/src/views/Waiting.tsx
```diff
@@ -0,0 +1,106 @@
...
+  // Variable to hold whether the user is connected to the socket.
+  const [socketConnected, setSocketConnected] = useState(false);
+
+  /**
+   * Nickname that is populated if the join page is on the waiting room stage.
+   * Set error if no nickname is passed in despite the waiting room stage.
+   */
+  if ((!location || !location.state || !location.state.nickname) && !error) {
+    setError('No nickname was provided for the user in the waiting room.');
+  }
```
alankbi, line 34: Thoughts on moving this and lines 77-79 into a useEffect hook that runs on page load (and with a [location] [dependency](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects))? That way we might not need the socketConnected variable and can also make sure it runs exactly once after the component mounts.

alankbi (approved):
Approved! Just left some small questions and suggestions, lmk what your thoughts are on those

frontend/src/views/Join.tsx
```diff
@@ -1,154 +1,27 @@
...
-
-  return joinPageContent;
+  const redirectToWaitingRoom = (nickname: string) => new Promise<undefined>((resolve) => {
+    history.push('/game/waiting', { nickname });
+    resolve();
+  });
+
+  // Render the "Enter nickname" state.
+  return (
+    <EnterNicknamePage
```
alankbi, line 19: I also got this error in the console from line 19 of this file: 
`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.`

frontend/src/components/config/App.tsx
zpChris: I was thinking that `waiting` wasn't the best name; `lobby` is much better, thanks I'll change it to that!

zpChris: Done!

frontend/src/views/Waiting.tsx
zpChris: Done -- updated to `LobbyPageLocation` 👍

frontend/src/views/Waiting.tsx
zpChris: I think that's a great idea -- thanks!

frontend/src/views/Join.tsx
zpChris: Thanks for pointing that out, I think I saw that earlier; it isn't popping up anymore, so I think it was fixed by adding `useEffect(...)`.

zpChris:
I re-requested review from @alankbi even though he approved because I changed a decent number of things. Namely, I included `useCallback` in the `connectUserToRoom` function so it does not update when included as a `useEffect` dependency -- see [this Medium article](https://medium.com/@infinitypaul/reactjs-useeffect-usecallback-simplified-91e69fb0e7a3).

alankbi (approved):
Looks good!

PR #39, opened 2020-09-23
Dynamic room endpoints

alankbi:
### List of Changes:

* Add dynamic socket endpoints that depend on the room id
* Make join room PUT request send update to socket subscribe endpoint
* Update frontend to use new routes
* Add unit and integration tests

### Notes:

* See [here](http://rafaelhz.github.io/testing-websockets/) for socket testing tutorial
* Used [this post](https://stackoverflow.com/questions/27047310/path-variables-in-spring-websockets-sendto-mapping) as reference
* The waiting/lobby page has a hardcoded room id as we haven't fully hooked up that functionality yet
* You won't be able to see a discernible difference in the waiting room behavior as it is now


alankbi:
[Dynamically create new custom socket endpoints based on rooms, and allow users to connect to them](https://trello.com/c/MUOPzKrz/99-dynamically-create-new-custom-socket-endpoints-based-on-rooms-and-allow-users-to-connect-to-them)

frontend/src/views/Lobby.tsx
```diff
@@ -27,6 +26,9 @@ function LobbyPage() {
   // Variable to hold whether the user is connected to the socket.
   const [socketConnected, setSocketConnected] = useState(false);
 
+  // Dummy room id for dynamic room endpoint purposes
+  const socketRoomId = '791894';
```
zpChris, line 30: nit: Add `: string` for the type.

frontend/src/api/Socket.ts
```diff
@@ -12,11 +12,18 @@ export type User = {
...
-// Create constants for the connection, subscription, and send message URLs.
-export const SOCKET_ENDPOINT:string = '/api/v1/socket/join-room-endpoint';
-export const SUBSCRIBE_URL:string = '/api/v1/socket/subscribe-user';
-const ADD_USER_URL:string = '/api/v1/socket/add-user';
-const DELETE_USER_URL:string = '/api/v1/socket/delete-user';
+// Dynamic route endpoints that depend on the room id
+const basePath = '/api/v1/socket';
+let socketRoomId: string;
+export const routes = (roomId: string) => {
+  socketRoomId = roomId;
```
zpChris, line 19: Why is this included here? The `socketRoomId` is reset every time the `routes(...)` is called, but `socketRoomId` is the parameter for this method often?

frontend/src/views/Lobby.tsx
zpChris: How would we get the actual `socketRoomId`? I assume this be set in the `location` object, but I just wanted to confirm this.

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
```diff
@@ -1,30 +1,34 @@
...
+import org.springframework.messaging.handler.annotation.DestinationVariable;
 import org.springframework.messaging.handler.annotation.MessageMapping;
 import org.springframework.messaging.handler.annotation.SendTo;
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
 public class UserSocketController extends BaseRestController {
 
-    Set<UserDto> users = new HashSet<>();
+    private static final String SEND_TO_PATH = BaseRestController.BASE_SOCKET_URL + "/{roomId}/subscribe-user";
```
zpChris, line 16: Where is the `roomId` variable? I don't see where it is actually set in this file?

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -16,23 +18,28 @@
...
+    private static final Random random = new Random();
+    private static final String SOCKET_PATH = BaseRestController.BASE_SOCKET_URL + "/%s/subscribe-user";
 
     private final RoomRepository repository;
-    private static final Random random = new Random();
+    private final SimpMessagingTemplate template;
 
     @Autowired
-    public RoomService(RoomRepository repository) {
+    public RoomService(RoomRepository repository, SimpMessagingTemplate template) {
```
zpChris, line 40: Is the `SimpMessagingTemplate template` initialized / included by default somehow?

zpChris (commented):
This looks good to me, but I'm commenting without explicit approval so I can get on a call with you first to better understand this PR.

frontend/src/api/Socket.ts
alankbi: Yeah so socketRoomId is meant to save the latest roomId passed in since sometimes the routes are used within `Socket.ts` and it needs to know what roomId to use. It's an imperfect solution so I'd be open to changing it if you had any suggestions

frontend/src/views/Lobby.tsx
alankbi: Yeah it will be passed using location (I do this in my next PR)

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
alankbi: It comes from the MessageMapping string and is mapped to a variable using `@DestinationVariable String roomId`, and then when the SendTo annotation occurs, it should replace `{roomId}` with the correct string.

src/main/java/com/rocketden/main/service/RoomService.java
alankbi: Yeah, it's autowired, so Spring should take care of initializing it for us and making sure a valid instance is passed to this class.

helenli8181 (approved):
I think this looks good to go!

frontend/src/api/Socket.ts
zpChris: If you think creating a React hook would be cleaner, you could do that, but otherwise I think it's alright.

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
```diff
@@ -1,30 +1,34 @@
...
+
+    private Set<UserDto> users = new HashSet<>();
 
-    @MessageMapping("/add-user")
-    @SendTo(BaseRestController.BASE_SOCKET_URL + "/subscribe-user")
-    public Set<UserDto> addUser(String nickname) {
+    @MessageMapping("/{roomId}/add-user")
+    @SendTo(SEND_TO_PATH)
+    public Set<UserDto> addUser(@DestinationVariable String roomId, String nickname) {
```
zpChris, line 22: This (and `deleteUser(...)`) won't be used anymore, right? If so, should we remove them in a separate PR, or this one?

zpChris (commented):
Left no explicit approval as I still have a question on whether you should remove the `addUser` and `deleteUser` methods. You may have a plan to update those in a future PR though; if so, let me know. I also just want to get a brief overview on the general testing setup during the standup tomorrow morning, if possible.

src/main/java/com/rocketden/main/controller/v1/UserSocketController.java
alankbi: I wasn't sure if we still needed them or not, but yeah I can delete them (I'll probably do it in #41 as this one still technically supports the addUser/deleteUser functionality whereas that one makes breaking changes).

zpChris (approved):
LGTM! Thanks for the explanations :)

PR #40, opened 2020-09-23
Create the Enter Room ID page and incorporate local storage.

zpChris:
### List of Changes:
- Create the Enter Room ID page on the Join Room page (add back in `pageState`).
- Include `localStorage` to save the currently-entered `roomId` and current `pageState`.

### Notes:
- I attempted to use the `location` React hook to save the current state, but it didn't work, so I decided to use `localStorage` instead. This means that the state and `roomId` is saved across different tabs.
- The user is unable to enter any character except for digits `0` through `9`. We can change this if we decide to create alphanumeric room IDs.

### Screencast and Images:
[Screencast of the Enter Room ID Page and Flow](https://www.loom.com/share/7936709fd9c04e278d927fa7b68ea599)
- I added the `roomId` to show in the URL (`/game/lobby?roomId=XXXXXX`) after recording this screencast.

Image of the Room ID Page
Room ID Page

### DIFF
--- a/frontend/src/views/Join.tsx
+++ b/frontend/src/views/Join.tsx
@@ -1,7 +1,6 @@
 import React, { ReactElement, useEffect, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import EnterNicknamePage from '../components/core/EnterNickname';
-import ErrorMessage from '../components/core/Error';
 import { LargeCenterInputText, LargeInputButton } from '../components/core/Input';
 import { LargeText, Text } from '../components/core/Text';
 
@@ -98,7 +97,6 @@ function JoinGamePage() {
               The room ID must have exactly six digits (numbers 0 through 9).
             </Text>
           ) : null}
-          { /* error ? <ErrorMessage message={error} /> : null */ }
         </div>
       );
       break;
--- a/frontend/src/views/Join.tsx
+++ b/frontend/src/views/Join.tsx
@@ -1,11 +1,27 @@
 import React, { ReactElement, useEffect, useState } from 'react';
-import { useHistory } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router-dom';
 import EnterNicknamePage from '../components/core/EnterNickname';
 import ErrorMessage from '../components/core/Error';
 import { LargeCenterInputText, LargeInputButton } from '../components/core/Input';
 import { LargeText, Text } from '../components/core/Text';
 
+type JoinPageLocation = {
+  pageState: number,
+  roomId?: string;
+}
+
 function JoinGamePage() {
+  // Get history object to be able to move between different pages
+  const location = useLocation<JoinPageLocation>();
+
+  // Function to update the location variable upon entered room ID.
+  const updateLocation = (roomIdParam: string) => {
+    location.state = {
+      pageState: 2,
+      roomId: roomIdParam,
+    };
+  };
+
   // Get history object to be able to move between different pages
   const history = useHistory();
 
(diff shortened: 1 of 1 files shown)


zpChris:
[Add Enter Room ID to the Join room page](https://trello.com/c/Pr6wVBYO/131-add-enter-room-id-to-the-join-room-page)

frontend/src/views/Join.tsx
```diff
@@ -1,27 +1,121 @@
...
+  // Variable to hold whether the user is focused on the text input field.
+  const [focusInput, setFocusInput] = useState(false);
+
+  /**
+   * The local storage is updated when the page state updates.
+   */
+  const updateLocalStorage = (pageStateParam: string, roomIdParam: string) => {
+    localStorage.setItem('pageState', pageStateParam);
+    localStorage.setItem('roomId', roomIdParam);
+  };
```
alankbi, line 41: This all looks great, although I'm also curious what your opinion would be if we decided not to save page state across refreshes? A couple of concerns I have now are 1. you can't leave the nickname state if you decide you want to switch rooms and 2. the page state persists in the browser, which might be unexpected if users wanted to join another room in a different tab or revisited the website after several days.

Not sure how Kahoot does things, but I'd honestly be fine with a page refresh simply resetting to state 1 (if users accidently refreshed they'd just need to re-enter the room id), or maybe when users enter a room id, it "redirects" them to `game/join/{id}` (using someting like [useParams](https://reactrouter.com/web/api/Hooks)) or `game/join?roomId=XXXXXX` (this is also a card in the sprint) which then shows the nickname display? That way the back button still works and it doesn't persist longer than a single session. 

My opinion might be the minority though, so I'd love to hear your thoughts as well as what @helenli8181 thinks, either here or possibly in a call as well.

alankbi (commented):
Nice work! I held off on reviewing for now as I thought it'd be good to first talk about one of the comments I left. Lmk what your thoughts are

frontend/src/views/Join.tsx
helenli8181: I think it would be fine to not save page state when refreshing. The number of possible errors we could avoid would be worth having to re-enter room code if you accidentally refresh imo.

helenli8181 (approved):
Looks awesome! I approved because I'm fine with whether you decide to use the code as is or make changes, though I'm following the convo about saving page state upon refresh.

frontend/src/views/Join.tsx
```diff
@@ -1,27 +1,121 @@
...
+            </Text>
+          ) : null}
+        </div>
+      );
+      break;
+    case 2:
+      // Render the "Enter nickname" state.
+      joinPageContent = (
+        <div>
+          <EnterNicknamePage
```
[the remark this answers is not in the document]
zpChris: I considered an `if` statement or something, but yeah I think ternary reduces readability IMO. I'll leave it as-is for now, lmk if you think it should change though.

frontend/src/views/Join.tsx
zpChris: Yeah for sure, I have no problem with that; I'll remove the functionality that saves the state in `localStorage`.

frontend/src/views/Join.tsx
```diff
@@ -1,27 +1,121 @@
...
+            enterNicknameHeaderText="Enter a nickname to join the game!"
```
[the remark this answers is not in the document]
zpChris: Love the suggestion -- updated!

frontend/src/views/Join.tsx
zpChris: Done!

alankbi (approved):
Looks good!

PR #41, opened 2020-09-23
Pass room info to lobby page after creating a room

alankbi:
### List of Changes:

* After creating a room, pass the room info to the lobby page using location
* Change socket return data to be the RoomDto
* Replace state to be user and room info on the lobby page
* Add redirect from lobby page to join page if room hasn't been joined

### Notes:

* I removed the addUser and deleteUser methods as we might not need them anymore
* Since the join page isn't hooked up, clicking enter on that page just redirects to itself for now

### Screencast and Images:

Minor UI updates:
Lobby page


alankbi:
[Connect create page with lobby page and REST endpoint](https://trello.com/c/P0C59QEU/133-connect-create-page-with-lobby-page-and-rest-endpoint)

frontend/src/views/Lobby.tsx
```diff
@@ -1,41 +1,47 @@
...
 
   // Set the nickname variable.
-  const [nickname, setNickname] = useState('');
+  const [user, setUser] = useState<User | null>(null);
 
   // Hold error text.
   const [error, setError] = useState('');
 
+  // Variable to determine whether to redirect back to join page
+  const [shouldRedirect, setShouldRedirect] = useState(false);
```
zpChris, line 25: Right now, there's a brief moment where you are on the lobby before you are redirected back. Am I correct in assuming that the next PR will take care of the actual room validation?

If that is true, then maybe you don't not need all this redirect logic, and instead you could just throw an error? Since you'll end up deleting this after merging #43, if I understand correctly.

frontend/src/views/Lobby.tsx
```diff
@@ -46,15 +52,10 @@ function LobbyPage() {
...
+  const connectUserToRoom = useCallback((roomId: string) => {
     connect(roomId).then(() => {
       subscribe(routes(roomId).subscribe, subscribeCallback).then(() => {
-        try {
-          addUser(nicknameParam);
-          setSocketConnected(true);
-        } catch (err) {
-          setError(err.message);
-        }
+        setSocketConnected(true);
```
zpChris, line 58: Doesn't this always redirect the user? Is that intended?

It always seems to redirect for me, even when I provide a correct room ID.

frontend/src/views/Lobby.tsx
```diff
@@ -65,36 +66,38 @@ function LobbyPage() {
 
   // Grab the nickname variable and add the user to the lobby.
   useEffect(() => {
-    // Grab the nickname variable; otherwise, set an error.
-    if (location && location.state && location.state.nickname) {
-      setNickname(location.state.nickname);
+    // Grab the user and room information; otherwise, redirect to the join page
+    if (location && location.state && location.state.user && location.state.room) {
+      setUser(location.state.user);
+      setRoom(location.state.room);
```
zpChris, line 72: What if the `user` is present but the `room` isn't? Or vice versa?

If you throw an error instead of redirecting, then you could throw an error depending on whether the `user` or `room` isn't provided; otherwise, no change is necessary.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -172,20 +172,14 @@ public void getRoomFailure() {
...
-        user.setNickname("test");
-        Set<User> users = new HashSet<>();
-        users.add(user);
-
-        UserDto userDto = new UserDto();
-        userDto.setNickname("test");
-        Set<UserDto> expected = new HashSet<>();
-        expected.add(userDto);
+        RoomDto roomDto = new RoomDto();
+        roomDto.setHost(new UserDto());
```
zpChris, line 176: Should you confirm that any `UserDto` nickname that is provided as well is also checked?

zpChris (changes requested):
Requesting changes because I have a few nits, but largely because I have questions on the current redirect setup you have constructed for the Lobby page.

frontend/src/views/Lobby.tsx
```diff
@@ -1,41 +1,47 @@
...
    * Update the users list.
    */
   const subscribeCallback = (result: Message) => {
-    const userObjects:User[] = JSON.parse(result.body);
-    setUsers(userObjects);
+    const newRoom:Room = JSON.parse(result.body);
+    setRoom(newRoom);
+  };
+
+  const deleteUser = (u: User) => {
```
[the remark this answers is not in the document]
alankbi: I agree that user is better than u; if I do this, I'll need to rename the user state variable, maybe to like currentUser? Lmk if you have any preference on the name

frontend/src/api/User.ts
```diff
@@ -0,0 +1,3 @@
+export type User = {
+  nickname: string,
```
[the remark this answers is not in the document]
alankbi: Right now it returns UserDto objects which don't include id, although that might be a good design decision we can talk about at the next meeting?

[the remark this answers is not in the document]
zpChris: Oh ok, if they don't include the `id` then I think this is OK as-is.

frontend/src/views/Lobby.tsx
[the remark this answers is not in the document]
zpChris: Ah I see -- `currentUser` or `userParam` are both good.

frontend/src/views/Lobby.tsx
```diff
@@ -65,36 +66,38 @@ function LobbyPage() {
...
-    // Grab the nickname variable; otherwise, set an error.
-    if (location && location.state && location.state.nickname) {
-      setNickname(location.state.nickname);
+    // Grab the user and room information; otherwise, redirect to the join page
+    if (location && location.state && location.state.user && location.state.room) {
+      setUser(location.state.user);
+      setRoom(location.state.room);
     } else {
-      setError('No nickname was provided for the user in the lobby.');
+      setShouldRedirect(true);
```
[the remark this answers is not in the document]
alankbi: I originally tried that but then realized doing so broke the back button functionality - basically, if you are pushed to the join page, then swipe back, you end up at the lobby page again, which immediately redirects you back to the join page. Idk how big of an issue that is for us though, so I can def change it back if cons outweigh the pros.

frontend/src/views/Lobby.tsx
alankbi: As for flashing, yeah I think that might be one of the drawbacks for using redirect (I'm not sure if `history.push` has the same issue or not). As for the second part, I'm not too sure - I was assuming that people who navigated to `/game/lobby` but hadn't actually joined/created a game (i.e. arrived at the lobby from a join/create page redirect) would be redirected back to the join page, however I'm totally open to changing the behavior if we all had something else in mind

frontend/src/views/Lobby.tsx
alankbi: Yeah if you navigate directly to `/game/lobby` it should always redirect you to the join page unless you already reached the page from the create/join pages (the current code looks only at the user/room location state rather than the roomId query param). My thinking behind this being that users need to enter a nickname then call the PUT endpoint to actually join the room, which currently all happens from the join page. Same thing as above tho - can easily change this if we don't like this design

frontend/src/views/Lobby.tsx
alankbi: Not sure, my thought was just that if any of those are missing, then the user hasn't joined the room yet and needs to re-join to be allowed back on the lobby page. However I can def look into that after discussisng/making the other changes you suggested!

src/test/java/com/rocketden/main/service/RoomServiceTests.java
alankbi: Yeah good point, just updated!

frontend/src/views/Lobby.tsx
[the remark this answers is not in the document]
alankbi: Sounds good, went with the `currentUser` option.

frontend/src/views/Lobby.tsx
```diff
@@ -65,35 +66,37 @@ function LobbyPage() {
...
       { error ? <ErrorMessage message={error} /> : null }
       <div>
         {
-          users.map((user) => (
-            <UserNicknameText onClick={(event) => {
-              deleteUser((event.target as HTMLElement).innerText);
-            }}
-            >
+          room?.users.map((user) => (
+            <UserNicknameText onClick={() => deleteUser(user)}>
```
helenli8181, line 99: This is probably better for a later PR -- should we have a pop up message or something asking the user to confirm they want to delete someone?

helenli8181 (commented):
This works well for me, but I'll hold off on approving for a little bit since I'm still trying to digest how some of the logic works.

frontend/src/views/Lobby.tsx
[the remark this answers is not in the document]
zpChris: Ok, that makes sense! I tried replacing this with `window.location.href = '/'`, but the issue you described occurred. I think your solution is good -- my only nit would be to put a comment _briefly_ noting why you are using this instead of the `history` hook.

frontend/src/views/Lobby.tsx
zpChris: I think in the future, there could be a brief "loading" state which the user sees, and then is redirected to a page with the message "you need to provide a valid room ID to enter the lobby". So this is definitely a fixable issue, and I think it is totally fine for now.

frontend/src/views/Lobby.tsx
zpChris: No, this makes sense -- thanks for the explanation!

frontend/src/views/Lobby.tsx
zpChris: Since it's the same action of redirect right now, I think this is fine. Later, I think it may make sense to redirect with a message on why they weren't connected to the lobby. If you wanted to do this in this PR, that works, but it's also fine to leave it as-is and get to that later.

zpChris (approved):
This PR looks great! I only left one nit on a past comment to add a brief comment on why you are using `<Redirect ... />` instead of the `history` hook.

Also, I left other comments noting that in the future, we may want to display a brief message explaining why the user was redirected (perhaps on the homepage) -- for example, it could be a brief floating message at the top that says "No user was provided." or "The provided room ID was invalid.". If you wanted to add that for this PR, you can, but no need to -- I think it already is great! :)

frontend/src/views/Lobby.tsx
alankbi: Yeah, right now it just does a simple `console.log` since we don't have that functionality yet. Lmk if you think I should change that to a pop-up or something tho

alankbi:
Sounds good, I added the comment and also changed the redirect to pass an error message to the join page. All that's left is to check location for an error on the join page, but since error handling on the join page is only added in #43, I'm thinking I might implement that over there to hopefully prevent merge conflicts.

helenli8181 (approved):
It looks great, thanks for the explanation a couple days ago! Sorry this approval came late

PR #42, opened 2020-09-24
Added delete user and tests

helenli8181:
### List of Changes:

- Added `deleteUser()` function in `UserService.java`
- Created `DeleteMapping` in UserController
- Added some tests in `UserTests` and `UserServiceTests`

### Notes:

- The variable `POST_USER` could be renamed. It sounds like a verb and not only post requests use it now. Any ideas?


src/main/java/com/rocketden/main/service/UserService.java
```diff
@@ -37,6 +38,17 @@ public UserDto createUser(CreateUserRequest request) {
...
+    public boolean deleteUser(DeleteUserRequest request) {
+        User user = repository.findUserByNickname(request.getNickname());
+
+        if (user == null) {
+            return false;
+        }
+
+        repository.delete(user);
+        return true;
+    }
```
zpChris, line 50: Thoughts on returning error handling similar to the other functions instead of `true` / `false`?

src/test/java/com/rocketden/main/UserTests.java
```diff
@@ -126,4 +130,35 @@ public void createNewUserNicknameContainsSpaces() throws Exception {
...
+
+        this.mockMvc.perform(post(POST_USER)
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(request)));
+
+        this.mockMvc.perform(delete(POST_USER)
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(request)))
+                .andExpect(status().isAccepted())
+                .andExpect(content().string("true"));
```
zpChris, line 150: I know `POST_USER` is the same value as `DELETE_USER`, but I think this makes it hard to follow from a readability standpoint. I'd recommend either renaming the variable or creating a duplicate variable named `DELETE_USER`.

zpChris (changes requested):
Looks good! Only left a nit and a change wondering if we should modify the return type of this, so it's not just `true` and `false` and instead provides more details.

alankbi:
This looks good, although do we also want to include some way to remove a user from a room?

src/test/java/com/rocketden/main/UserTests.java
helenli8181: How does just `USER` sound?

src/main/java/com/rocketden/main/service/UserService.java
helenli8181: Done, per our discussion yesterday!

src/main/java/com/rocketden/main/service/UserService.java
```diff
@@ -37,6 +38,19 @@ public UserDto createUser(CreateUserRequest request) {
         return UserMapper.toDto(user);
     }
 
+    public UserDto deleteUser(DeleteUserRequest request) {
+        User user = repository.findUserByNickname(request.getNickname());
```
alankbi, line 42: If there are multiple users with the given nickname, does this just get the first user?

src/main/java/com/rocketden/main/service/UserService.java
```diff
@@ -37,6 +38,19 @@ public UserDto createUser(CreateUserRequest request) {
...
+
+        if (user == null) {
+            throw new ApiException(UserError.NOT_FOUND);
+        }
+
+        UserDto ret = UserMapper.toDto(user);
```
alankbi, line 48: nit: I think you can delete this line and change the return to just be `return UserMapper.toDto(user)`

src/test/java/com/rocketden/main/UserTests.java
```diff
@@ -126,4 +130,33 @@ public void createNewUserNicknameContainsSpaces() throws Exception {
...
+        request.setNickname("rocket");
+
+        this.mockMvc.perform(post(POST_USER)
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(request)));
+
+        this.mockMvc.perform(delete(POST_USER)
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(request)))
+                .andExpect(status().isAccepted());
```
alankbi, line 149: nit: I would suggest doing something similar to the above tests, where it gets the returned objects and verifies that they match up correctly.

src/test/java/com/rocketden/main/UserTests.java
```diff
@@ -126,4 +130,33 @@ public void createNewUserNicknameContainsSpaces() throws Exception {
...
+
+    @Test
+    public void deleteNonExistingtUser() throws Exception {
+        DeleteUserRequest request = new DeleteUserRequest();
+        request.setNickname("rocket");
+
+        this.mockMvc.perform(delete(POST_USER)
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(request)))
+                .andExpect(status().isNotFound());
```
alankbi, line 160: Same as my other comment, I'd suggest adding a check that the returned error message is correct (see the other tests)

src/test/java/com/rocketden/main/service/UserServiceTests.java
```diff
@@ -13,8 +15,9 @@
 import org.mockito.Spy;
 import org.mockito.junit.jupiter.MockitoExtension;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
alankbi, line 18: nit: personal preference, but I think we should stick with importing each item individually (see [here](https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad)). Sometimes the IDE can auto-correct imports into the wildcard tho so usually no big deal if it slips through

src/test/java/com/rocketden/main/service/UserServiceTests.java
```diff
@@ -35,4 +38,40 @@ public void createUserSuccess() {
         verify(repository).save(Mockito.any(User.class));
         assertEquals("rocket", response.getNickname());
     }
+
+    @Test
+    public void createUserInvalidNickname() {
+        CreateUserRequest request = new CreateUserRequest();
+        request.setNickname("rocket rocket");
+
+        assertThrows(ApiException.class, () -> {
```
alankbi, line 47: nit: Sorry if this is overkill, but I'd also suggest verifying in these two tests that the exception responses contain the correct error messages (see `RoomServiceTests.java` for examples)

alankbi (changes requested):
The new design looks good! Sorry I kind of left a lot of nits haha, but all of them should be pretty quick fixes and then I think this should be good to merge 👍

src/test/java/com/rocketden/main/UserTests.java
```diff
@@ -126,4 +130,33 @@ public void createNewUserNicknameContainsSpaces() throws Exception {
...
+                .content(Utility.convertObjectToJsonString(request)));
+
+        this.mockMvc.perform(delete(POST_USER)
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(request)))
+                .andExpect(status().isAccepted());
+    }
+
+    @Test
+    public void deleteNonExistingtUser() throws Exception {
```
alankbi, line 153: nit: Fix spelling (`deleteNonExistentUser`)

zpChris:
I'll wait until @alankbi's comments are addressed until I review, since there are a decent number of nits. 👍

If I should review now instead, please just let me know! :)

src/main/java/com/rocketden/main/service/UserService.java
helenli8181: Yes, since right now UserDto doesn't contain another ID. This depends on how findUserByNickname works though which I'm not super clear about.

alankbi (approved):
Approved!

src/main/java/com/rocketden/main/controller/v1/UserController.java
```diff
@@ -25,4 +27,9 @@ public UserController(UserService service) {
     public ResponseEntity<UserDto> createUser(@RequestBody CreateUserRequest request) {
         return new ResponseEntity<>(service.createUser(request), HttpStatus.CREATED);
     }
+
+    @DeleteMapping("/user")
+    public ResponseEntity<UserDto> deleteUser(@RequestBody DeleteUserRequest request) {
+        return new ResponseEntity<>(service.deleteUser(request), HttpStatus.ACCEPTED);
```
zpChris, line 33: Am I correct in assuming that this code only goes through if the user is successfully deleted, and otherwise, it does not return the `HttpStatus.ACCEPTED` but instead throws an exception? If that's the case, should this return `HttpStatus.OK` based on [this article](https://briancaos.wordpress.com/2019/06/19/200-ok-vs-202-accepted-whats-the-difference/#:~:text=200%20OK%20means%20that%20the,the%20service%20will%20now%20start.)? (If so, you'll want to update the tests accordingly as well.)

src/main/java/com/rocketden/main/service/UserService.java
```diff
@@ -37,6 +38,18 @@ public UserDto createUser(CreateUserRequest request) {
         return UserMapper.toDto(user);
     }
 
+    public UserDto deleteUser(DeleteUserRequest request) {
+        User user = repository.findUserByNickname(request.getNickname());
+
+        if (user == null) {
```
zpChris, line 44: nit, personal preference: Add a comment above this `if` statement, similar to the `createUser(...)` method above. (No problem if you don't do this, just a personal preference.)

src/test/java/com/rocketden/main/UserTests.java
```diff
@@ -126,4 +130,50 @@ public void createNewUserNicknameContainsSpaces() throws Exception {
...
+        CreateUserRequest createRequest = new CreateUserRequest();
+        createRequest.setNickname("rocket");
+
+        UserDto expected = new UserDto();
+        expected.setNickname("rocket");
+
+        DeleteUserRequest request = new DeleteUserRequest();
+        request.setNickname("rocket");
+
+        this.mockMvc.perform(post(POST_USER)
```
zpChris, line 145: I think we talked about this before, but thoughts on changing this to just `USER` or `USER_URI`? (Same for the tests below.)

src/test/java/com/rocketden/main/service/UserServiceTests.java
```diff
@@ -35,4 +41,44 @@ public void createUserSuccess() {
         verify(repository).save(Mockito.any(User.class));
         assertEquals("rocket", response.getNickname());
     }
+
+    @Test
+    public void createUserInvalidNickname() {
+        CreateUserRequest request = new CreateUserRequest();
+        request.setNickname("rocket rocket");
```
zpChris, line 48: nit: Add comment on why this nickname is invalid (I can see us changing those details in the future, and if so, we may be confused by this method without an explanatory comment IMO).

zpChris (changes requested):
Requesting changes because I just have a few nits, but otherwise looks great! I'll try to review quickly after you address these, since it shouldn't take long 👍

src/main/java/com/rocketden/main/controller/v1/UserController.java
helenli8181: Yes to the first part. Sure I can switch it to an OK status; I mainly used ACCEPTED because I saw others use it with @deletemapping.

src/test/java/com/rocketden/main/UserTests.java
helenli8181: Whoops I forgot about this! Fixed now.

zpChris (approved):
LGTM! Thanks for making those changes 👍

zpChris:
[Delete users from room and database](https://trello.com/c/861Jcxvj/113-delete-users-from-room-and-database)

helenli8181:
[Delete users from room and database](https://trello.com/c/861Jcxvj/113-delete-users-from-room-and-database)

PR #43, opened 2020-09-24
Connect join room to REST endpoints

alankbi:
### List of Changes:

* Add verify and join room REST endpoints
* Check valid roomId before moving to nickname page
* Join room through PUT request after entering in a nickname
* Miscellaneous error and loading checks

### Notes:

* The waiting page will not work until #41 is also merged (after both go in though then it should hopefully all finally connect together)

### Screencast and Images:

[Screencast of the new join page behavior](https://www.loom.com/share/4fceb7f109cd4b0b94013fa12949cdf1)


frontend/src/views/Join.tsx
```diff
@@ -67,23 +103,25 @@ function JoinGamePage() {
...
           { focusInput && !validRoomId ? (
             <Text>
               The room ID must have exactly six digits (numbers 0 through 9).
             </Text>
           ) : null}
+          { loading ? <Loading /> : null }
+          { error ? <ErrorMessage message={error} /> : null }
         </div>
       );
       break;
```
zpChris, line 127: Cannot comment on an unchanged line, so this must've been on a separate PR (maybe mine). However, I just noticed that the comment on line 134 is no longer applicable, and should be removed.

frontend/src/views/Join.tsx
```diff
@@ -67,23 +103,25 @@ function JoinGamePage() {
...
           <LargeInputButton
             onClick={() => {
-              setPageState(2);
+              checkRoom();
             }}
             value="Enter"
-            // Input is disabled if no nickname exists, has a space, or is too long.
-            disabled={!validRoomId}
+            // Input is disabled if loading or if no nickname exists, has a space, or is too long.
+            disabled={!validRoomId || loading}
```
zpChris, line 116: Nice change! Good thinking, preventing spamming of that button 👍

frontend/src/views/Join.tsx
```diff
@@ -67,23 +103,25 @@ function JoinGamePage() {
                 event.preventDefault();
               }
               if (event.key === 'Enter' && validRoomId) {
```
zpChris, line 105: Add the `loading` parameter here as well, to prevent spamming of pressing that key.

frontend/src/components/core/EnterNickname.tsx
```diff
@@ -44,6 +44,9 @@ export default function EnterNicknamePage(props: EnterNicknameProps) {
    * with loading or error messages as necessary.
    */
   const enterNicknameActionUpdatePage = (nicknameParam: string): void => {
+    // Don't perform action if previous action is still running
+    if (loading) return;
+
```
zpChris, line 49: Similar to my other comment, can you change this to only run the function on `!loading`, and otherwise just not doing anything? This likely means you don't need a return type (even `void`?) for the function.

frontend/src/views/Create.tsx
```diff
@@ -23,9 +22,7 @@ function CreateGamePage() {
       .then((res) => {
         redirectToLobby(res);
         resolve();
-      }).catch((err) => {
-        reject(errorHandler(err.message));
-      });
+      }).catch((err) => reject(err));
```
zpChris, line 25: Is there no need for the `errorHandler(...)` method? If so, should we remove its other instances? (namely in `Socket.ts`, but elsewhere as well)

zpChris (changes requested):
Requesting changes, but they're essentially all nits -- no major changes. This looks good!

I also will want to re-review this once #41 is merged, to confirm that you're correct in thinking that the waiting page will work at that time. In the future, I think I'd branch this PR off of #41, so that we can see whether it is working or not while initially reviewing -- but that's open for discussion.

frontend/src/views/Create.tsx
alankbi: I think the way it's done in `Socket.ts` is correct, but for here I noticed that the REST calls already return `axiosErrorHandler(err)` so I think there's no need to make another call to `errorHandler`.

frontend/src/views/Join.tsx
alankbi: Sounds good - I also added a check in the if statement above to prevent default if the user has already entered 6 digits, lmk if you want me to undo that but I thought that it could be useful.

frontend/src/views/Join.tsx
```diff
@@ -62,28 +96,31 @@ function JoinGamePage() {
               setFocusInput(false);
             }}
             onKeyPress={(event) => {
-              // If the key pressed is not a number, do not add it to the input.
-              if (event.key < '0' || event.key > '9') {
+              // If the key pressed is not a number or the roomId is
+              // already at length 6, do not add it to the input.
```
zpChris, line 100: nit: Use `/** */` type comments.

zpChris (approved):
Just left one nit, but otherwise, looks good! I also created a new issue to remove the `errorHandler` method from the frontend #44.

frontend/src/views/Join.tsx
```diff
@@ -19,9 +23,15 @@ function JoinGamePage() {
...
   const [roomId, setRoomId] = useState('');
 
+  // Hold loading boolean, triggered upon entering nickname.
+  const [loading, setLoading] = useState(false);
+
+  // Set error if one exists
+  const [error, setError] = useState('');
+
   /**
    * The nickname is valid if it is non-empty and has exactly
```
helenli8181, line 33: Extreme nit: room id rather than nickname?

helenli8181 (approved):
Looks good to me, sorry for the late review!

alankbi:
Build failed; I'll push a fix after #41 is merged that way I can handle this + merge conflicts at the same time. Since I also pushed a new feature, I can also re-request you two after I'm finished with all that.

helenli8181:
[Delete users from room and database](https://trello.com/c/861Jcxvj/113-delete-users-from-room-and-database)

alankbi:
Pushed some significant changes to fix bugs and hook up all the behavior (it should be working properly), most notably the following:

* Deconstruct room into several state variables, since React apparently only does a shallow object check when determining whether to re-render (meaning it wouldn't detect if a new object was added to `room.users[]`)
* If unauthorized user navigates to `/game/lobby`, they'll be redirected to `/game/join` with an error message (which disappears after refresh)
* Identify host on the lobby page

There's still a bug where after page refresh on the lobby page, the room info reverts to whatever was stored in `location.state` on first arrival to the page (#45). I tried fixing it by updating location state every time a user joins, but that didn't work (plus I don't think that's how location is supposed to be used), so not sure what the best solution for that would be. A possible fix could also just be to warn before refreshing and if they do decide to refresh, remove them from the room.

frontend/src/views/Lobby.tsx
```diff
@@ -69,16 +80,16 @@ function LobbyPage() {
     // Grab the user and room information; otherwise, redirect to the join page
     if (location && location.state && location.state.user && location.state.room) {
       setCurrentUser(location.state.user);
-      setRoom(location.state.room);
+      setStateFromRoom(location.state.room);
```
zpChris, line 83: Maybe instead of grabbing all of the current room information from `location`, you could just grab the `roomId` and `user` then do a GET request for the rest of the information?

I tried a variant of updating location every time the `room` variable was updated, but wasn't able to get it to work. :/ I don't know if there's another way to update the location, and a GET request seems to make more sense in terms of flow, so I think that may be a good option.

zpChris (changes requested):
Mostly looks good to me! I thought maybe we could switch `location` with the room information with simply the `roomId` and `user` (or at least only those are used), and then do a GET request once on the lobby page to get the current room information. Thoughts?

frontend/src/views/Lobby.tsx
alankbi: After talking about the new design we're planning on going with, this sounds great - implemented in the latest commits.

alankbi:
New updates as per suggestions: 

* Convert GET endpoint to return RoomDto instead of just the roomId
* Instead of passing room info through location to the lobby page, just pass the roomId
* On the lobby page, call GET endpoint to get room info before connecting

These changes should fix #45.

zpChris (approved):
I think this looks great! I left one nit, and also wanted to confirm that the `GET` request now returns a `Room` instead of a `RoomDto` object -- what was your thinking on that?

Either way, it looks great -- approving! 👍

frontend/src/util/Utility.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (!(location && location.state)) {
+    return false;
+  }
+
+  let valid = true;
+  params.forEach((param) => {
+    valid = valid && (param in location.state);
+  });
+
+  return valid;
```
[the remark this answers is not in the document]
alankbi: Just tried that, I get the following warning:
Alternative checkLocationState implementation

I'm also not sure whether returning inside the callback does so for the callback itself or the actual checkLocationState object, so for now, I might just leave as is (but I'll be totally cool if someone wants to reimplement that logic at a later point)

alankbi:
About the GET return object, I think it should still be a RoomDto object, it's just that on the frontend the type is simply called Room

helenli8181 (approved):
New changes look great! The host state in the lobby will be super useful for the start game functionality too

frontend/src/util/Utility.ts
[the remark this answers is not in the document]
zpChris: No worries, that sounds good 👍 Thanks for the screenshot!

PR #48, opened 2020-10-02
Start Game functionality redirects users to game

helenli8181:
### List of Changes:

- Added a `Start Game` button in the Lobby page that only the host can see. Other users will see a message "Waiting for host to start game...". 
- When the host clicks the button, a POST request is sent which leads to the websocket redirecting all users to the game page. 

### Notes:

- I'm undecided on whether it's better to have separate Game files for all of these pieces or to put them together with the Room files. The first option probably makes more sense in the long run. For now, I made changes to `RoomService` instead of making a new `GameService` class. LMK if you think I should move it out instead. 

### Screencast and Images:

[Screencast of start game redirect](https://www.loom.com/share/9aea93b36fa943f1acac30829db9d8f9)

Screen Shot 2020-10-02 at 12 05 20 PM


zpChris:
> ### List of Changes:
> * Added a `Start Game` button in the Lobby page that only the host can see. Other users will see a message "Waiting for host to start game...".
> * When the host clicks the button, a POST request is sent which leads to the websocket redirecting all users to the game page.
> 
> ### Notes:
> * I'm undecided on whether it's better to have separate Game files for all of these pieces or to put them together with the Room files. The first option probably makes more sense in the long run. For now, I made changes to `RoomService` instead of making a new `GameService` class. LMK if you think I should move it out instead.
> * I think I want to add a five-second countdown on the Lobby after the host clicks the start game button.
> 
> ### Screencast and Images:
> Screen Shot 2020-10-02 at 12 05 20 PM

Two notes:
1. The image you uploaded has an error with socket connection. Can you add an image that accurately represents what the PR should look like?
2. This screencast seems particularly well-fitting for a screencast -- can you make one?

frontend/src/views/Lobby.tsx
```diff
@@ -127,6 +157,9 @@ function LobbyPage() {
           ))
         }
       </div>
+      {currentUser?.nickname === host?.nickname
+        ? <PrimaryButton onClick={handleClick}> Start Game </PrimaryButton>
+        : 'Waiting for the host to start the game...'}
```
zpChris, line 162: nit: Put `'Waiting for the host to start the game...'` in some sort of React element (`<Text>` maybe, or something slightly larger?). Also, if that change doesn't add a bit more spacing between the names and this text in the "waiting" version, then I'd suggest adding a tad more spacing as well (see image to know what version I'm talking about).
Current Lobby Page Non-Host Perspective

src/main/java/com/rocketden/main/controller/v1/GameController.java
```diff
@@ -0,0 +1,40 @@
...
+        RoomDto roomDto = service.getRoom(getRoomRequest);
+
+        if (!roomDto.getHost().getNickname().equals(request.getUser().getNickname())) {
+            return new ResponseEntity<>(HttpStatus.FORBIDDEN);
+        }
+
+        service.startGame(request);
+        return new ResponseEntity<>(HttpStatus.OK);
+    }
+}
```
zpChris, line 40: nit: Add ending newline.

src/main/java/com/rocketden/main/controller/v1/GameController.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @PostMapping("/start")
+    public ResponseEntity<Void> startGame(@RequestBody StartGameRequest request) {
+        // Check if current user ID is the host
+        GetRoomRequest getRoomRequest = new GetRoomRequest();
+        getRoomRequest.setRoomId(request.getRoomId());
+        RoomDto roomDto = service.getRoom(getRoomRequest);
+
+        if (!roomDto.getHost().getNickname().equals(request.getUser().getNickname())) {
```
zpChris, line 33: This is throwing an error for me, same as it did for you yesterday. (See images below.)

Can you upload a screencast of it working? That way I can see what it's supposed to look like, and also troubleshoot why it may be doing this (since I remember this happened for you occasionally as well).

Backend 405 Error

Frontend Error

src/main/java/com/rocketden/main/dto/game/StartGameRequest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class StartGameRequest {
+    private String roomId;
+    private UserDto user;
+}
```
zpChris, line 13: nit: Add newline.

frontend/src/views/Lobby.tsx
```diff
@@ -48,6 +53,15 @@ function LobbyPage() {
     console.log(user);
   };
 
+  const handleClick = (): Promise<void> => {
+    const request = { roomId: currentRoomId, User: currentUser };
+    return axios.post<void>('/api/v1/start', request)
+      .then(() => { console.log('redirect to game'); })
```
zpChris, line 59: nit: Remove `console.log(...)` and replace it with a `setLoading(...)` function that works as it does on other pages (you will need to add a React element to the actual function return as well for this).

frontend/src/views/Lobby.tsx
```diff
@@ -48,6 +53,15 @@ function LobbyPage() {
...
+      .catch((err) => {
+        throw axiosErrorHandler(err);
```
zpChris, line 61: Use the `setError` function (or add it if it does not exist), so the user can see that / when an error occurred. (See my image below for what happens otherwise -- it just prints to the console, but the user does not get any notice that it failed.)

Error on Starting Game

zpChris (changes requested):
Currently the PR isn't working for me, I'm getting the same `405` error that you had received earlier. I also left other comments on things to change, but most pressing I think is to upload a screencast of it working so I can see what the expected behavior is like (from the **host and non-host** points of view).

helenli8181:
> Currently the PR isn't working for me, I'm getting the same `405` error that you had received earlier. I also left other comments on things to change, but most pressing I think is to upload a screencast of it working so I can see what the expected behavior is like (from the **host and non-host** points of view).

That's strange, I just pulled this branch locally and it worked fine. Sure I'll upload a screen recording!

src/main/java/com/rocketden/main/controller/v1/GameController.java
helenli8181: Screencast uploaded! It's in the description of the PR.

helenli8181: 405 error fixed -- turns out the problem was typo in the handleClick function :)

src/main/java/com/rocketden/main/controller/v1/GameController.java
```diff
@@ -0,0 +1,40 @@
...
+    
+    private final RoomService service;
+
+    @Autowired
+    public GameController(RoomService service) {
+        this.service = service;
+    }
+
+    @PostMapping("/start")
+    public ResponseEntity<Void> startGame(@RequestBody StartGameRequest request) {
```
zpChris, line 27: This should return the `RoomDto` I believe; it's not used now, but it will for sure need to be used in the future as the frontend will need to know what users are in the room and such.

If you didn't add this because you didn't want to add something in that wouldn't be used, then I'm open for discussion (also thoughts @alankbi on this?), but otherwise, I think this should return the appropriate `RoomDto`, as well as updating that on the frontend (it wouldn't have to use it at all, perhaps just pass the details into the `/game` url.

The more I type this, the more unsure I get as you already have a lot added here, so just let me know if you think it should be added or not.

frontend/src/views/Lobby.tsx
```diff
@@ -108,6 +135,14 @@ function LobbyPage() {
           }}
         />
       ) : null}
+      { shouldRedirectToGame ? (
+        <Redirect
```
zpChris, line 139: I tested refreshing right as I clicked start game, and it redirected all the other users but kept the host in the lobby page.

Obviously we talked about this earlier as a total edge case, so no action needed. I wanted to note though, that in the future, I think we may be able to fix this by setting a variable in the `RoomDto` as `active` or something, which indicates that the game is ongoing and any user still in the lobby somehow should be redirected to the game page.

Food for thought. 🥘

zpChris (changes requested):
Left a few more comments, but essentially all nits -- this is looking great! Nice job 😃 

My last comment is do we need any tests on this PR? I'm curious on @alankbi's thoughts, since I think he tested similar socket-type code before.

frontend/src/views/Lobby.tsx
```diff
@@ -108,6 +135,14 @@ function LobbyPage() {
...
+          to={{
+            pathname: '/game',
+          }}
+        />
```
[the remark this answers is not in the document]
alankbi: I could be wrong but I thought you already fixed that in your PR #50?

src/main/java/com/rocketden/main/controller/v1/GameController.java
alankbi: I'm also not sure/don't have a preference; for consistency, RoomDto makes sense, but for minimizing data transfer, Void is also reasonable, so either way seems reasonable to me

frontend/src/views/Lobby.tsx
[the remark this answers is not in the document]
helenli8181: I'm pretty sure the other `<Redirect />` right above this one has already been changed to history in another PR that hasn't merged yet? Or maybe it's merged and I need to pull the changes in.

frontend/src/views/Lobby.tsx
alankbi: I also agree on this point of adding an `active` or other variable to the Room/RoomDto objects, as I think it has several design advantages:

1. Fixes the refreshing bug as @zpChris mentioned
2. A user tries to join a room that's already started - having an `active` parameter will allow us to prevent them from joining
3. Users can have all the info they need just from subscribing to a single socket endpoint
4. It allows us to keep track of valuable room/game info (e.g. which rooms have started) in the backend

frontend/src/views/Lobby.tsx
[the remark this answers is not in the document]
zpChris: Yeah that's right @alankbi and @helenli8181, I fixed that in PR #50 which I just merged -- not great that I forgot my own code haha, but yeah that should be fine.

zpChris:
Note: You're going to want to resolve the conflicts in `Lobby.tsx` that have arisen since I merged PR #50.

src/main/java/com/rocketden/main/controller/v1/GameController.java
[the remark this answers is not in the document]
helenli8181: Yep apparently `getNickname()` isn't necessary, so I'll remove it.

src/main/java/com/rocketden/main/controller/v1/GameController.java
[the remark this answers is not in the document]
helenli8181: I'm not really sure it's necessary, since in both cases someone who is not the host is attempting to use host rights. A normal non-host user wouldn't "accidentally" stumble upon the start game button. Do you think it makes a difference? If so I'm open to adding it.

helenli8181: Sure, I can have it return a RoomDto and the front end doesn't do anything with it until we need that information.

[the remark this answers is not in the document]
zpChris: That's fair reasoning, ok no worries, you don't have to add this (we can reopen the discussion if @alankbi thinks different, but resolving now).

frontend/src/views/Lobby.tsx
zpChris: @helenli8181 Thoughts?

zpChris (approved):
LGTM! I'll let @alankbi make suggestions on whether or not adding tests is important.

frontend/src/views/Lobby.tsx
```diff
@@ -46,6 +52,17 @@ function LobbyPage() {
...
+  const handleClick = (): Promise<void> => {
+    const request = { roomId: currentRoomId, user: currentUser };
+    return axios.post<void>('/api/v1/start', request)
+      .then(() => {
+        setLoading(true);
+      })
+      .catch((err) => {
+        setError(err.message);
+      });
+  };
```
alankbi, line 64: nit: Move this post request to a new file (maybe `Game.ts`) in the `api` folder using a similar format as the requests in `Room.ts` (from the looks of it the new post request also returns a RoomDto)

Also, I think `handleClick` can be void if you decide to do this, but I may be wrong so lmk

frontend/src/views/Lobby.tsx
```diff
@@ -63,16 +80,25 @@ function LobbyPage() {
       setStateFromRoom(JSON.parse(result.body));
     };
 
+    const startGameCallback = () => {
+      history.replace('/game');
```
alankbi, line 84: I'm actually not sure whether replace or push would be better here, open to hearing thoughts on this. On the one hand, using replace, clicking back will take you directly back to the join/create page which seems like good behavior. On the other hand, I feel like push makes more sense cause you're adding to the history, whereas we've mainly used replace when a user is on a page they're unauthorized for.

src/main/java/com/rocketden/main/controller/v1/GameController.java
```diff
@@ -0,0 +1,38 @@
...
+public class GameController extends BaseRestController {
+    
+    private final RoomService service;
+
+    @Autowired
+    public GameController(RoomService service) {
+        this.service = service;
+    }
+
+    @PostMapping("/start")
```
alankbi, line 25: What are your thoughts on making this something like `/rooms/{roomId}/start`? Just to match with the new (?) convention we started (idk if we actually all agreed on it so we can also talk about it during standup)

src/main/java/com/rocketden/main/controller/v1/GameController.java
```diff
@@ -0,0 +1,38 @@
...
+    public ResponseEntity<RoomDto> startGame(@RequestBody StartGameRequest request) {
+        // Check if current user ID is the host
+        GetRoomRequest getRoomRequest = new GetRoomRequest();
+        getRoomRequest.setRoomId(request.getRoomId());
+        RoomDto roomDto = service.getRoom(getRoomRequest);
+
+        if (!roomDto.getHost().equals(request.getUser())) {
+            return new ResponseEntity<>(HttpStatus.FORBIDDEN);
+        }
```
alankbi, line 34: nit: Move this logic to the `service.startGame` method (making use of `repository.findRoomByRoomId` and `RoomError.INVALID_PERMISSIONS`)

src/main/java/com/rocketden/main/dto/game/StartGameRequest.java
```diff
@@ -0,0 +1,13 @@
...
+import com.rocketden.main.dto.user.UserDto;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class StartGameRequest {
+    private String roomId;
+    private UserDto user;
```
alankbi, line 12: Small nit but could you call this initiator, or if not then maybe leave a comment mentioning that it's the person sending the POST request? Mainly to help others remember what it's used for down the line

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -133,10 +134,19 @@ public RoomDto updateRoomHost(String roomId, UpdateHostRequest request) {
 
     // Send updates about new users to the client through sockets
     public void sendSocketUpdate(RoomDto roomDto) {
-        String socketPath = String.format(SOCKET_PATH, roomDto.getRoomId());
+        String socketPath = String.format(SUBSCRIBE_USER_SOCKET_PATH, roomDto.getRoomId());
         template.convertAndSend(socketPath, roomDto);
     }
 
+    // Send request to redirect users to game when host clicks start
+    public RoomDto startGame(StartGameRequest request) {
```
alankbi, line 142: Also a suggestion/thought but if we have this method in the RoomService, could we move the endpoint from the GameController file to the RoomController file as well for consistency? Or maybe vice versa (create a new GameService class), whatever you think makes the more sense

Branching off of this, I'd also suggest adding some tests that cover this method/REST endpoint (either in RoomTests and RoomServiceTests or creating new Game test files), as well as maybe one test in RoomSocketTests to make sure the socket message goes through)

alankbi (changes requested):
Looks good! My biggest suggestion is adding some tests, but other than that I mainly just have smaller style preferences

frontend/src/views/Lobby.tsx
helenli8181: The Loading and Error states can't be accessed from outside this function I think. Is making a new file important? If so I can try to find a way around it.

frontend/src/views/Lobby.tsx
helenli8181: In that case I think push makes a little more sense if it can be recorded in the history. Do you mean the back button doesn't work with push?

src/main/java/com/rocketden/main/controller/v1/GameController.java
helenli8181: Sure, I think putting the room ID in it would be good.

frontend/src/views/Lobby.tsx
alankbi: You should be able to do something like `callPostRequestImportedFromAnotherFile().then(...).catch(...)` as long as you make the request return a promise (if you look at `Room.ts` you can see what each function should look like as well as how those functions are used in other React files

frontend/src/views/Lobby.tsx
alankbi: I think that was only the case when users are unauthorized to access the page and are immediately redirected elsewhere, cause then if you click back, then it would just automatically redirect you again. In this scenario I don't think that would be a problem (it's also like an edge case so we can always figure it out at a later time).

frontend/src/api/Game.ts
```diff
@@ -0,0 +1,21 @@
+import axios from 'axios';
+import { axiosErrorHandler } from './Error';
+import { User } from './User';
+
+export type Room = {
+  roomId: string;
+  host: User;
+  users: [User],
+};
```
alankbi, line 9: nit: I don't think this is used so I think it can be removed

frontend/src/api/Game.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type StartGameParams = {
+  roomId: string,
+  initiator: User,
+};
+
+export const startGame = (params: StartGameParams):
+  Promise<void> => axios.post<void>('/api/v1/rooms/{currentRoomId}/start', params)
```
alankbi, line 17: This POST request fails - I think you need to use a template string (which means you'll also need take in another parameter currentRoomId as I don't think that exists here yet)

Screenshot

frontend/src/api/Game.ts
```diff
@@ -0,0 +1,21 @@
...
+import { User } from './User';
+
+export type Room = {
+  roomId: string;
+  host: User;
+  users: [User],
+};
+
+export type StartGameParams = {
+  roomId: string,
```
alankbi, line 12: nit: I think roomId can be removed if it's being passed through the URL instead

frontend/src/views/Lobby.tsx
```diff
@@ -46,6 +52,17 @@ function LobbyPage() {
     console.log(user);
   };
 
+  const handleClick = () => {
+    const request = { roomId: currentRoomId, initiator: currentUser as User };
+    startGame(request)
+      .then(() => {
```
alankbi, line 58: I think request can be passed as a parameter to startGame instead of in the request body now that you changed it to use a path variable

src/main/java/com/rocketden/main/dto/game/StartGameRequest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.rocketden.main.dto.user.UserDto;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class StartGameRequest {
+    private String roomId;
```
alankbi, line 11: nit: Remove now that this is no longer being used

src/main/java/com/rocketden/main/controller/v1/GameController.java
```diff
@@ -0,0 +1,29 @@
...
+    private final GameService service;
+
+    @Autowired
+    public GameController(GameService service) {
+        this.service = service;
+    }
+
+    @PostMapping("/rooms/{roomId}/start")
+    public ResponseEntity<RoomDto> startGame(@RequestBody StartGameRequest request) {
+        return new ResponseEntity<>(service.startGame(request), HttpStatus.OK);
```
alankbi, line 27: I think you can add another parameter `@PathVariable String roomId` and pass that to `service.startGame` now that you added a path variable to the URL

src/main/java/com/rocketden/main/exception/UserError.java
```diff
@@ -9,7 +9,8 @@
 public enum UserError implements ApiError {
 
     INVALID_USER(HttpStatus.BAD_REQUEST, "Please provide a user with a valid nickname (1-16 characters without spaces)."),
-    NOT_FOUND(HttpStatus.NOT_FOUND, "A user with given nickname could not be found in database.");
+    NOT_FOUND(HttpStatus.NOT_FOUND, "A user with given nickname could not be found in database."),
+    ACCESS_DENIED(HttpStatus.FORBIDDEN, "Access Denied.");
```
alankbi, line 13: This is just a nit but I'd suggest reusing `RoomError.INVALID_PERMISSIONS` instead

src/main/java/com/rocketden/main/service/GameService.java
```diff
@@ -0,0 +1,49 @@
...
+
+		Room room = repository.findRoomByRoomId(request.getRoomId());
+
+		// If requested room does not exist in database, throw an exception.
+		if (room == null) {
+			throw new ApiException(RoomError.NOT_FOUND);
+		}
+
+		// if user making request is not the host, throw an exception.
+		if (request.getInitiator().getNickname() != room.getHost().getNickname()) {
```
alankbi, line 43: nit: Change to `!request.getInitiator().equals(room.getHost())`

src/main/java/com/rocketden/main/service/GameService.java
```diff
@@ -0,0 +1,49 @@
...
+	@Autowired
+	public GameService(RoomRepository repository, SimpMessagingTemplate template) {
+		this.repository = repository;
+		this.template = template;
+	}
+
+	// Send request to redirect users to game when host clicks start
+	public RoomDto startGame(StartGameRequest request) {
+		String socketPath = String.format(START_GAME_SOCKET_PATH, request.getRoomId());
+		template.convertAndSend(socketPath, request.getRoomId());
```
alankbi, line 33: I think this should be placed at the end after all the permission/room verification (otherwise this will send a socket message even if the POST request fails later down the line)

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -0,0 +1,69 @@
...
+		UserDto user = new UserDto();
+		user.setNickname("rocketrocket");
+		StartGameRequest request = new StartGameRequest();
+		request.setRoomId(roomDto.getRoomId());
+		request.setInitiator(user);
+
+		this.mockMvc.perform(post(String.format(START_GAME, roomDto.getRoomId()))
+				.contentType(MediaType.APPLICATION_JSON_VALUE)
+				.content(Utility.convertObjectToJsonString(request)))
+				.andExpect(status().isForbidden());
```
alankbi, line 67: nit: I would suggest sticking to the same format as in `RoomTests.java` (where you do the whole `ERROR = RoomErrors.SOMETHING` and then `assertEquals(error.getResponse, actual)`)

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
```
alankbi, line 69: nit: Add testing for room not found (I think it's ok to just add that at the end of the current test), and then another test for a successful response

src/test/java/com/rocketden/main/service/GameServiceTests.java
```diff
@@ -0,0 +1,69 @@
...
+		room.setHost(host);
+
+		StartGameRequest request = new StartGameRequest();
+		request.setRoomId(roomId);
+		request.setInitiator(UserMapper.toDto(host));
+
+		Mockito.doReturn(room).when(repository).findRoomByRoomId(request.getRoomId());
+		RoomDto response = gameService.startGame(request);
+
+		assertEquals(roomId, response.getRoomId());
```
alankbi, line 54: nit: Add a verify statement to make sure the socket send goes through (there should be a few examples in RoomServiceTests)

src/test/java/com/rocketden/main/service/GameServiceTests.java
```diff
@@ -0,0 +1,69 @@
...
+		UserDto user = new UserDto();
+		user.setNickname("rocket");
+		StartGameRequest request = new StartGameRequest();
+		request.setRoomId("123456");
+		request.setInitiator(user);
+
+		Mockito.doReturn(null).when(repository).findRoomByRoomId(request.getRoomId());
+		ApiException exception = assertThrows(ApiException.class, () -> gameService.startGame(request));
+		assertEquals(RoomError.NOT_FOUND, exception.getError());
+	}
```
alankbi, line 68: nit: Add another test for permission denied

alankbi (changes requested):
Thanks for adding the tests - just left a _couple_ more comments (13 actually haha sorry, X of them are nits tho) and then I think it should be good after that

frontend/src/api/Game.ts
helenli8181: As it turns out I forgot I was programming in Java and used `!=` to test string equality :)

frontend/src/api/Game.ts
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios';
+import { axiosErrorHandler } from './Error';
+import { User } from './User';
+
+export type StartGameParams = {
+  initiator: User,
+};
+
+export const startGame = (params: StartGameParams):
+  Promise<void> => axios.post<void>('/api/v1/rooms/{currentRoomId}/start', params)
```
alankbi, line 10: Change to the following (+ add import for Room): 

```suggestion
export const startGame = (roomId: string, params: StartGameParams):
  Promise<Room> => axios.post<Room>(`/api/v1/rooms/${roomId}/start`, params)
```

frontend/src/views/Lobby.tsx
```diff
@@ -46,6 +52,17 @@ function LobbyPage() {
     console.log(user);
   };
 
+  const handleClick = () => {
+    const request = { initiator: currentUser as User };
+    startGame(request)
```
alankbi, line 57:
```suggestion
    startGame(currentRoomId, request)
```

src/main/java/com/rocketden/main/service/GameService.java
```diff
@@ -0,0 +1,51 @@
...
+	public RoomDto startGame(String roomId, StartGameRequest request) {
+		Room room = repository.findRoomByRoomId(roomId);
+
+		// If requested room does not exist in database, throw an exception.
+		if (room == null) {
+			throw new ApiException(RoomError.NOT_FOUND);
+		}
+
+		// if user making request is not the host, throw an exception.
+		if (!request.getInitiator().getNickname().equals(room.getHost().getNickname())) {
```
alankbi, line 40:
```suggestion
		if (!request.getInitiator().equals(room.getHost())) {
```

alankbi (approved):
Approved! I left 3 suggested changes however; lmk what you think about those. 

Also, since there have been a lot of changes since Chris reviewed, maybe @zpChris can take a quick look again/do another run-through before merging to make sure there aren't any new bugs

src/main/java/com/rocketden/main/service/GameService.java
helenli8181: This doesn't work, maybe because one is a UserDto and the other is a User?

alankbi: Oh my bad, in that case maybe add a line above like this: 

```java
User initiator = UserMapper.toEntity(request.getInitiator());
```

And then you can compare them using `!room.getHost().equals(initiator)`

helenli8181:
[Implement start game functionality (all users rerouted to game page)](https://trello.com/c/1BGJZOnf/68-implement-start-game-functionality-all-users-rerouted-to-game-page)

PR #49, opened 2020-10-03
Add connected variable to User class.

zpChris:
### List of Changes:

- Add a `connected` variable to the `User` class to prevent the following exception:

```
java.sql.SQLException: Field 'connected' doesn't have a default value
```

### Notes:

- This is only necessary because the changes in the branch `DetectUserExit` added a `connected` boolean to the `User` object in the database, and since the boolean doesn't have a default value, you need to add this field to `User.java`.

### Screencast and Images:

No screencasts or images are needed as there are no changes to the UI or database -- just fixes.

### DIFF
--- a/src/main/java/com/rocketden/main/model/User.java
+++ b/src/main/java/com/rocketden/main/model/User.java
@@ -25,6 +25,9 @@ public class User {
     @EqualsAndHashCode.Include
     private String nickname;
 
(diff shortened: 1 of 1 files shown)


zpChris:
This is actually unnecessary as I don't need the `connected` variable; I can just use `sessionId == null` to determine whether the user is connected or not.

PR #52, opened 2020-10-04
Add PUT endpoint to change room hosts

alankbi:
### List of Changes:

* Add PUT endpoint for changing room host
* Create new error type for invalid permissions
* Change room's set of users to list of users
* Add tests for new endpoint

### Notes:

* The set of users was changed to a list because sets don't have a get method, and we need that to ensure our objects are equal on both an application level and database level
* A new PUT request in Postman has been added for changing room hosts
* This PR has a lot of lines changed, but probably more than 75% of that is tests

### DIFF
--- a/src/main/java/com/rocketden/main/controller/v1/RoomController.java
+++ b/src/main/java/com/rocketden/main/controller/v1/RoomController.java
@@ -11,6 +11,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
@@ -41,8 +42,9 @@ public class RoomController extends BaseRestController {
         return new ResponseEntity<>(service.createRoom(request), HttpStatus.CREATED);
     }
 
-    @PutMapping("/rooms/host")
-    public ResponseEntity<RoomDto> updateRoomHost(@RequestBody UpdateHostRequest request) {
-        return new ResponseEntity<>(service.updateRoomHost(request), HttpStatus.OK);
+    @PutMapping("/rooms/{roomId}/host")
+    public ResponseEntity<RoomDto> updateRoomHost(@PathVariable String roomId,
+                                                  @RequestBody UpdateHostRequest request) {
+        return new ResponseEntity<>(service.updateRoomHost(roomId, request), HttpStatus.OK);
     }
 }
--- a/src/main/java/com/rocketden/main/controller/v1/RoomController.java
+++ b/src/main/java/com/rocketden/main/controller/v1/RoomController.java
@@ -4,6 +4,7 @@ import com.rocketden.main.dto.room.CreateRoomRequest;
 import com.rocketden.main.dto.room.GetRoomRequest;
 import com.rocketden.main.dto.room.JoinRoomRequest;
 import com.rocketden.main.dto.room.RoomDto;
+import com.rocketden.main.dto.room.UpdateHostRequest;
 import com.rocketden.main.service.RoomService;
 
 import org.springframework.beans.factory.annotation.Autowired;
@@ -38,5 +39,10 @@ public class RoomController extends BaseRestController {
     @PostMapping("/rooms")
     public ResponseEntity<RoomDto> createRoom(@RequestBody CreateRoomRequest request) {
         return new ResponseEntity<>(service.createRoom(request), HttpStatus.CREATED);
+    }
+
+    @PutMapping("/rooms/host")
+    public ResponseEntity<RoomDto> updateRoomHost(@RequestBody UpdateHostRequest request) {
+        return new ResponseEntity<>(service.updateRoomHost(request), HttpStatus.OK);
     }
 }
--- a/src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
+++ b/src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
@@ -1,0 +1,4 @@
+package com.rocketden.main.dto.room;
+
+public class UpdateHostRequest {
+}
--- a/src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
+++ b/src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
@@ -1,4 +1,13 @@
 package com.rocketden.main.dto.room;
 
+import com.rocketden.main.dto.user.UserDto;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
 public class UpdateHostRequest {
+    private String roomId;
+    private UserDto initiator;
+    private UserDto newHost;
 }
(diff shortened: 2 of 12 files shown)


alankbi:
[Create PUT endpoint for changing the room host](https://trello.com/c/5ePjlDbw/137-create-put-endpoint-for-changing-the-room-host)

helenli8181 (approved):
This looks awesome! I just had a small comment but other than that it's good by me.

src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.rocketden.main.dto.user.UserDto;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class UpdateHostRequest {
+    private UserDto initiator;
+    private UserDto newHost;
```
zpChris, line 11: Can the host set themselves as the new host? I would assume yes, but want to make sure that if someone does that there aren't any unexpected consequences (bugs in the backend or whatnot for whatever reason).

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
```diff
@@ -98,24 +99,64 @@ public void handleFrame(StompHeaders headers, Object payload) {
                 blockingQueue.add((RoomDto) payload);
             }
         });
+    }
+
+    @Test
+    public void socketReceivesMessageOnJoin() throws Exception {
```
zpChris, line 105: Thanks for adding this! This is unrelated to your current PR, right? This is a great test to add either way.

src/test/java/com/rocketden/main/api/RoomTests.java
```diff
@@ -315,4 +313,163 @@ public void createAndJoinRoomNoUser() throws Exception {
...
+
+        jsonResponse = result.getResponse().getContentAsString();
+        actual = Utility.toObject(jsonResponse, ApiErrorResponse.class);
+
+        assertEquals(ERROR.getResponse(), actual);
+
+        // New host does not exist in the room
+        updateHostRequest = new UpdateHostRequest();
+        updateHostRequest.setInitiator(host);
+        updateHostRequest.setNewHost(new UserDto());
```
zpChris, line 460: If I understand correctly, you are testing multiple different errors in the same method. Thoughts on taking these out and separating them into separate methods?

If you are concerned about the large amount of duplicate code beforehand, maybe place that in a setup helper function, and have a separate test (you may already) to ensure that that code is indeed working. Let me know your thoughts.

zpChris (changes requested):
I left a few nits, requesting changes primarily because there is a test that I think could be separate into multiple different tests for readability and debugging purposes.

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
```diff
@@ -42,16 +42,41 @@ public void removeUserFromRoomSet() {
 
         User user = new User();
         user.setNickname("test");
+        user.setId(1);
         room.addUser(user);
         
         User userToRemove = new User();
-
         userToRemove.setNickname("nonexistent");
+        userToRemove.setId(2);
```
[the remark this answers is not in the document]
alankbi: We talked about this in a call and I think reached a consensus - feel free to re-request if there's anything else that comes up tho

src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
alankbi: Just tested and yeah they can technically set themselves as the host (no errors occur, although I'm thinking in the frontend we can probably just prevent that from being an option)

src/test/java/com/rocketden/main/api/RoomTests.java
alankbi: Yeah that's a good idea, done!

src/main/java/com/rocketden/main/controller/v1/RoomController.java
```diff
@@ -39,4 +41,10 @@ public RoomController(RoomService service) {
     public ResponseEntity<RoomDto> createRoom(@RequestBody CreateRoomRequest request) {
         return new ResponseEntity<>(service.createRoom(request), HttpStatus.CREATED);
     }
+
+    @PutMapping("/rooms/{roomId}/host")
+    public ResponseEntity<RoomDto> updateRoomHost(@PathVariable String roomId,
+                                                  @RequestBody UpdateHostRequest request) {
```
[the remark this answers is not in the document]
alankbi: Yeah based on what we talked about in standup (it might have been a while back) I think maybe we should change the join room and maybe get room endpoints to also use the path variables. I added a card for that in the backlog, but if we end up deciding to stick with room id as part of the request I can also come back and fix this if that sounds good

src/main/java/com/rocketden/main/dto/room/UpdateHostRequest.java
zpChris: Ok awesome, thanks for verifying that, and yeah the frontend design sounds good.

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -62,7 +62,11 @@ public boolean removeUser(User user) {
      * "equal" a user in the room at the application level, they are not the same at
      * the database level (different primary keys). This method therefore returns the
      * correct user object that's equal at both the application and database levels.
+     *
+     * @param userToFind the desired user to find
+     * @return the equivalent user in the database
      */
+
```
zpChris, line 69: nit: Remove line? Clearly you added that in, if there was a reason please lmk, but as I understood convent was to have no lines between the method comment and the method itself.

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -62,7 +62,11 @@ public boolean removeUser(User user) {
      * "equal" a user in the room at the application level, they are not the same at
      * the database level (different primary keys). This method therefore returns the
      * correct user object that's equal at both the application and database levels.
+     *
+     * @param userToFind the desired user to find
+     * @return the equivalent user in the database
```
zpChris, line 67: Include the fact that you return `null` if that user is not in the database.

src/test/java/com/rocketden/main/api/RoomTests.java
```diff
@@ -425,51 +372,116 @@ public void changeRoomHostFailure() throws Exception {
...
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(joinRequest)))
+                .andDo(print()).andExpect(status().isOk())
+                .andReturn();
+
+        jsonResponse = result.getResponse().getContentAsString();
+        room = Utility.toObject(jsonResponse, RoomDto.class);
+
+        assertEquals(host, room.getHost());
+        assertEquals(2, room.getUsers().size());
```
zpChris, line 483: nit, personal preference: Verify that the non-host user is indeed the one that you added.

zpChris (approved):
LGTM! I left a few nits for quick fixes, but after that, the PR looks great!

src/main/java/com/rocketden/main/model/Room.java
alankbi: Haha yep you're right, accidentally left a new line there

src/test/java/com/rocketden/main/api/RoomTests.java
alankbi: Added a contains statement to check if the list contains the UserDto that we specified

PR #53, opened 2020-10-05
Manage active and inactive users through connection and disconnection updates.

zpChris:
### List of Changes:

- Add `onSocketConnected` and `onSocketDisconnected` events to the backend that are triggered as described. The flow is as follows:
  - When the user initially joins or creates the join, they are added to the database. At this point, they are given a unique `userId`, which is then returned back to the frontend.
  - The user then connects to the socket, and passes in their `userId` through the socket headers. Then, the `onSocketConnected` in the backend is triggered and receives the `userId`, as well as a unique `sessionId` related to its connection. This method finds the user by its `userId`, and then adds the `sessionId` to that user. A socket update is then sent to all users in that particular room with this updated information.
    - Note: The user is connected to a socket whenever the `sessionId` is present on the backend.
    - Note Two: The socket update with the room details includes the list of `activeUsers` and list of `inactiveUsers`. This allows minimal logic to occur on the frontend. However, I am open to a restructuring here.
  - If any user disconnects, this triggers the `onSocketDisconnected` event on the backend, and provides just the `sessionId`. The user is then found by this `sessionId`, the `sessionId` is then removed to indicate that the user is no longer connected, and a new socket update is sent to the room.
- Update the frontend to indicate which users are **active** (connected to a socket) and which are **inactive** (not connected to a socket).
  - I am very open to design changes on this, or even removing the **inactive** users completely.
- Update the `RoomDto` to include `activeUsers` and `inactiveUsers`, which supports the formatting of the above change for the frontend. I also updated the Room `toDto(...)` method accordingly.
- Create a `Utility` class, separate from the new `TestUtility` class for tests, that generates IDs for now both the `userId` and the `roomId`.
  - There is not a test for this `Utility` class / specific method, but I am happy to add that to this PR, or I am also happy for leaving it for a future PR given our previous conversations.
- Update tests and methods across the codebase to use the `userId` to identify users, rather than the `nickname`.

### Notes:

- **Most important note:** I am expecting this to take a long time to get merged in. This PR is not finalized, has code across a very large number of files, and likely has many inconsistencies in tests and possibly in design. Please don't hold back on this PR. 👍 
- A broad overview of how the PR works is on [this Trello card](https://trello.com/c/rvDXYhPM/132-detect-when-a-user-leaves-the-room-on-backend-and-send-socket-update), both in the description and comments.
- I did not add relevant tests for this PR -- I posted a long message in the chat for why, but the short version is this PR has taken a long time and has a lot of changes, and I want to get something in.
- If a user closes a tab, and then returns to it through `cmd + shift + t`, the tab is not fully refreshed, and they have to refresh it for the socket information to show correctly on their page and the other users as well. This should be addressed / fixed in the future.
- Currently, the `sessionId` is not included as a decider to determine whether two User objects are equal; this is because not all `user` variables seemed to update on the frontend with this information, and I wanted to get in this PR rather than spend more time troubleshooting. This design decision could be good or bad, as it is technically relevant but the `userId` should really be the only necessary field when determining equality, so let me know.
- Currently, a host can be inactive as well. This could present issues with the host trying to change hosts or update users while "inactive."
- I think I poorly updated the current tests. There should be lots to improve there; namely, we can create file-wide variables from `roomId` and `userId` variables. There should also be some uniformity on where these variables are added; they are needed in many tests, as users are now identified by their `userId` instead of their `nickname`.
- There are a few other extraneous changes that may not be included in the list above. Any code that you come across that you don't know what it does, please let me know and I'll explain it the best I can (or change it if it's bad code).

### Screencast and Images:

_Note: The below screencast and images do not have the updated formatting with hovering over users to delete / change hosts, or with the users starting the game. However, these features should still work as expected. I am happy to update these, but please also test them locally to make sure they work for you as well._

[Screencast of Inactive and Active Room Management](https://www.loom.com/share/06561b554df1480ba2f86e59687cb921)

Image of the Room with One Inactive User
Image of the Room with One Inactive User

Image of the Room with No Inactive Users
Image of the Room with No Inactive Users


zpChris:
[Detect when a user leaves the room on backend, and send socket update](https://trello.com/c/rvDXYhPM/132-detect-when-a-user-leaves-the-room-on-backend-and-send-socket-update)

helenli8181:
[Implement start game functionality (all users rerouted to game page)](https://trello.com/c/1BGJZOnf/68-implement-start-game-functionality-all-users-rerouted-to-game-page)

src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -0,0 +1,24 @@
+package com.rocketden.main.util;
+
+import java.util.Random;
+
+import com.rocketden.main.controller.v1.BaseRestController;
+
+public class Utility {
+
```
alankbi, line 8: Thoughts on converting this into a Component that way you can autowire it? Similarly, (if autowiring) thoughts on moving/renaming it to something like GeneratorService?

src/main/java/com/rocketden/main/model/User.java
```diff
@@ -18,13 +18,21 @@
...
 public class User {
 
+    // ID used for the database.
     @Id
     @GeneratedValue(strategy = GenerationType.AUTO)
     private Integer id;
 
+    // User ID for user identification and accessible from frontend.
+    @EqualsAndHashCode.Include
+    private String userId;
```
alankbi, line 28: Thoughts on changing this to `private String userId = UUID.randomUUID().toString()`? For the following reasons:

1. Guaranteed uniqueness (at least for all intents and purposes)
2. Automatically initialized for every user, so we don't need a manual check for null-ness in our service classes
3. User ID is never shown on the UI so it shouldn't be too big a deal if it's long and ugly

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -50,36 +50,44 @@
     @InjectMocks
     private RoomService service;
 
+    @Mock
+    private static Utility utility;
```
alankbi, line 54: If you do decide to go the autowired route, I could be wrong but I think this would need to be above `service` and be non-static

alankbi: Also just a realization, if you decide to go with the UUID route, then there may not actually be a need to extract the generateRoomId method into Utility at all (so that might eliminate some of the difficulties with mocking)

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -347,17 +357,4 @@ public void sendSocketUpdate() {
...
-        // Verify room ids are generated correctly
-        String roomId = service.generateRoomId();
-
-        assertEquals(RoomService.ROOM_ID_LENGTH, roomId.length());
-
-        for (char c : roomId.toCharArray()) {
-            assertTrue(c >= '0');
-            assertTrue(c <= '9');
-        }
-    }
```
alankbi, line 362: Is this test added elsewhere?

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
```diff
@@ -144,6 +149,9 @@ public void socketReceivesMessageOnHostChange() throws Exception {
         assertEquals(expected.getHost(), actual.getHost());
         assertEquals(expected.getUsers(), actual.getUsers());
 
+        // Update newUser with the created userId and updated information.
+        newUser = expected.getUsers().get(1);
```
alankbi, line 153: I've noticed that this has come up a lot where you need to iterate through the users after joining to get the updated `userId` (both in the tests and on the frontend) - do you think it'd be worth creating a new response object for these cases? For example either adding like a `currentUser` field to `RoomDto` or creating like a new `JoinRoomResponse` class that has the two fields `currentUser` and `room`

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
```diff
@@ -161,7 +169,6 @@ public void socketReceivesMessageOnHostChange() throws Exception {
         assertEquals(expected.getUsers(), actual.getUsers());
     }
 
-    @Test
     public void socketReceivesMessageOnSettingsChange() throws Exception {
```
alankbi, line 172: nit: Re-add as test

alankbi: Or if it's still broken for some reason we can pair program to try to figure it out, or leave a TODO comment to make sure to address it in the next PR

src/test/java/com/rocketden/main/testutil/TestUtility.java
```diff
@@ -1,8 +1,8 @@
-package com.rocketden.main.util;
+package com.rocketden.main.testutil;
 
 import com.google.gson.Gson;
 
-public class Utility {
+public class TestUtility {
```
alankbi, line 5: Is there any reason this needs to be in a `testutil` package rather than `util`? Since you renamed it I thought that would fix the conflict, but I may be wrong on this so lmk

alankbi (changes requested):
Finally got through it all 🙂 Left some comments as well as the following general suggestions/questions:

1. That all sounds good with regards to a) using `userId` by itself for equality and b) leaving the CTRL+SHIFT+T thing for later (you should create an issue for that)
2. I'm still not sure what the best way is to go forward with the active/inactive users, which I'll talk about below:

* Right now, when users first join the room, they're briefly shown as inactive. I feel like it might be better for this to be shown as like grayed out or loading or something rather than red/inactive. However, the obvious problem with this is that there's no easy way to distinguish between just joining and people that actually have disconnected. 
* I also don't know whether we should have a separate section for users that are inactive. Instead maybe we can just have them all in the same group but just at the bottom of the list grayed out 
  * I think this would look better if we eventually convert the users into a list/column view of sorts rather than the current way of having cards lined up in a row
  * Having them grayed out at the bottom in a column view might also solve the issue of inactive vs. just joining, as I can imagine that'd look fine if someone appears briefly at the bottom then moves up once they're connected
  * This also brings up _another_ thing which is whether we want any type of sorting for the players or whether it's just order in which they enter haha
* Another option is we could even just not show inactive users at all; I'm pretty sure this is what Colonist does. When users refresh they completely disappear from the room before rejoining. If the room has only one person and they refresh, then the room is straight up deleted
  * If we do this, then would we even need the join endpoint? Since it'd seem like users are only truly "joined" if they have a session id assigned to them, so when they close the tab or refresh then they'd be actually removed from the room and then re-added
  * Then again if we just wanted to not show the inactive users we could always just keep the existing logic while just ignoring/deleting the `inactiveUsers` list and that'd be all we have to do

We should all definitely talk about the design for this in a call or something

src/main/java/com/rocketden/main/dao/UserRepository.java
```diff
@@ -7,5 +7,6 @@
 public interface UserRepository extends CrudRepository<User, Integer> {
 
     // Auto generated by CrudRepository
-    User findUserByNickname(String nickname);
+    User findUserByUserId(String userId);
+	User findUserBySessionId(String sessionId);
```
helenli8181, line 11: :)

helenli8181 (changes requested):
Great job managing such a large PR :) I have a couple of suggestions including nits for now.

frontend/src/components/card/PlayerCard.tsx
```diff
@@ -19,29 +19,50 @@ const Content = styled.div`
   border-radius: 20px;
 `;
 
+const ContentInactive = styled(Content)`
+  background-color: ${({ theme }) => theme.colors.lightRed};
+`;
+
```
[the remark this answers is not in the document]
zpChris: Ah, I tried to do this but didn't realize you had to put `Props` on that line -- thanks!

frontend/src/views/Lobby.tsx
```diff
@@ -122,11 +125,17 @@ function LobbyPage() {
...
       getRoom(location.state.roomId)
-        .then((res) => setStateFromRoom(res))
+        .then((res) => {
+          setStateFromRoom(res);
+          // Reset the user to hold the ID.
+          res.inactiveUsers.forEach((user: User) => {
+            if (user.nickname === location.state.user.nickname) {
+              setCurrentUser(user);
+            }
+          });
```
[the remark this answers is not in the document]
zpChris: Yeah I think since we have to make this REST call anyway to get all the room details anyway, I'm fine having this here; however, let me know if you think I should still move this.

src/main/java/com/rocketden/main/config/WebSocketConnectionEvents.java
```diff
@@ -0,0 +1,95 @@
...
+        Room room = user.getRoom();
+        RoomDto roomDto = RoomMapper.toDto(room);
+        sendSocketUpdate(roomDto);
+    }
+
+    // Send updates about new users to the client through sockets
+    public void sendSocketUpdate(RoomDto roomDto) {
+        String socketPath = String.format(Utility.SOCKET_PATH, roomDto.getRoomId());
+        template.convertAndSend(socketPath, roomDto);
+    }
```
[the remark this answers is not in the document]
zpChris: This is a good idea, thanks -- done!

src/main/java/com/rocketden/main/controller/v1/RoomController.java
```diff
@@ -35,12 +38,12 @@ public RoomController(RoomService service) {
...
     @PutMapping("/rooms")
     public ResponseEntity<RoomDto> joinRoom(@RequestBody JoinRoomRequest request) {
-        return new ResponseEntity<>(service.joinRoom(request), HttpStatus.OK);
+        return new ResponseEntity<>(service.joinRoom(request, utility), HttpStatus.OK);
     }
 
     @PostMapping("/rooms")
     public ResponseEntity<RoomDto> createRoom(@RequestBody CreateRoomRequest request) {
-        return new ResponseEntity<>(service.createRoom(request), HttpStatus.CREATED);
+        return new ResponseEntity<>(service.createRoom(request, utility), HttpStatus.CREATED);
```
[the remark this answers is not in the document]
zpChris: I'm not sure how this works? -- I'll ask clarifying questions during the call at 10am today.

src/main/java/com/rocketden/main/dto/user/UserDto.java
```diff
@@ -8,5 +8,9 @@
 @Setter
 @EqualsAndHashCode
 public class UserDto {
+    private String userId;
     private String nickname;
+
+    // The session ID of the user connection, auto-generated by sockets.
+    private String sessionId;
```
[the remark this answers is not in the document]
zpChris: I think it should be sent to the frontend -- it provides an easy way to determine whether or not any particular user is currently connected to the socket.

src/main/java/com/rocketden/main/model/User.java
```diff
@@ -18,13 +18,21 @@
...
+    
     @EqualsAndHashCode.Include
```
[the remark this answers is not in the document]
zpChris: I see this as a potential second layer of verification; for example, if two users are created but not added to the database, they may not have the `userId` field, so that value could be `null` despite them both having different `nickname`s. Furthermore, if they have the same `userId`, then the should presumably in all cases have the same `nickname`, so it shouldn't change any behavior under that circumstance. (The reason I didn't include it with `sessionId` is because not all places on the frontend are updated with this value I believe, so this _could_ actually return an incorrect value.)

Thoughts? I'm still very open to changing this.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -54,9 +53,17 @@ public RoomDto joinRoom(JoinRoomRequest request) {
...
-        // Return error if user is already in the room
-        if (room.getUsers().contains(user)) {
-            throw new ApiException(RoomError.USER_ALREADY_PRESENT);
+        // Return error if a user with the same nickname is in the room.
+        List<User> users = room.getUsers();
+        for (User roomUser : users) {
+            if (roomUser.getNickname().equals(user.getNickname())) {
+                throw new ApiException(RoomError.USER_WITH_NICKNAME_ALREADY_PRESENT);
+            };
+        }
```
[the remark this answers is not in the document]
zpChris: Sounds good -- done!

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -54,9 +53,17 @@ public RoomDto joinRoom(JoinRoomRequest request) {
...
+
+        // Add userId if not already present.
+        if (user.getUserId() == null) {
+            user.setUserId(utility.generateId(UserService.USER_ID_LENGTH));
         }
```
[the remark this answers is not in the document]
zpChris: What do you mean -- we don't check if the `user` already has a `userId`? If so, where should it be added?

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -79,8 +86,17 @@ public RoomDto createRoom(CreateRoomRequest request) {
...
 
+        // Create user ID for the host if not already present.
+        if (host.getUserId() == null) {
+            host.setUserId(utility.generateId(UserService.USER_ID_LENGTH));
+        }
+
+        // Add the host to a new user list.
+        List<User> users = new ArrayList<>();
+        users.add(host);
```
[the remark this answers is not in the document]
zpChris: Good catch -- removed!

src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -0,0 +1,24 @@
...
+    private static final Random random = new Random();
+    public static final String SOCKET_PATH = BaseRestController.BASE_SOCKET_URL + "/%s/subscribe-user";
```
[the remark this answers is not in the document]
zpChris: That's a good idea, I'll move this, thanks! 👍

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
zpChris: Shoot sorry, I didn't realize I removed this test -- re-added and it works perfect! 🙂

src/test/java/com/rocketden/main/testutil/TestUtility.java
zpChris: Good point -- fixed 👍

src/main/java/com/rocketden/main/util/Utility.java
zpChris: I like the idea of renaming it to `GeneratorService`, that could allow me to keep the name `Utility.java` for the test file. I'll ask the question of how the auto-wiring is done in this context on the call today.

src/main/java/com/rocketden/main/model/User.java
zpChris: This is a good idea! Would I want to ensure that this doesn't override any existing ID, so I could place this in a constructor, and if `userId` is `null`, then use that line above? I'm not sure if this is needed, but wanted to check.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Yeah I realized this too, I could potentially just put that back in the `Room` class -- we can talk about this today during our standup.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: It isn't, and it should be -- depending on how we implement the `generateId` method, this could be re-added as it is above, or added in a separate test class.

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
zpChris: Yeah, I think this could be a good idea! We'd want to differentiate this response object from what is sent out through the socket message though I think, because that is sent out to all users.

We can talk more about the potential design of this on the call today.

src/main/java/com/rocketden/main/dao/UserRepository.java
zpChris: Haha thank you, fixed :)

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -144,10 +154,10 @@ public void joinRoomDuplicateUserFailure() {
 
         // Mock repository to return room when called
         Mockito.doReturn(room).when(repository).findRoomByRoomId(eq(roomId));
-        ApiException exception = assertThrows(ApiException.class, () -> service.joinRoom(request));
+        ApiException exception = assertThrows(ApiException.class, () -> service.joinRoom(request, utility));
 
         verify(repository).findRoomByRoomId(roomId);
-        assertEquals(RoomError.USER_ALREADY_PRESENT, exception.getError());
+        assertEquals(RoomError.USER_WITH_NICKNAME_ALREADY_PRESENT, exception.getError());
```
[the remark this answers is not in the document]
zpChris: Hahaha that's so much better 🤣 not sure how I came up with the first constant name. Thanks!

frontend/src/components/card/PlayerCard.tsx
```diff
@@ -19,29 +19,50 @@ const Content = styled.div`
...
+        {showActionCard ? actionCard : null}
+      </Content>
+    ) : (
+      <ContentInactive
+        onMouseEnter={() => setShowActionCard(true)}
+        onMouseLeave={() => setShowActionCard(false)}
+      >
+        <UserNicknameText>
+          {user.nickname}
+          {isHost ? ' (host)' : ''}
```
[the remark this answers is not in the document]
zpChris: This is a good question -- we'll discuss this during the call today!

src/main/java/com/rocketden/main/model/User.java
zpChris: Based on the conversation in today's meeting, we'll leave this as a task for another PR, especially because testing this static method could potentially be very difficult. See the [Trello card](https://trello.com/c/EeQrpcem/159-uuid).

src/main/java/com/rocketden/main/service/RoomService.java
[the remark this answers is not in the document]
zpChris: Based on our conversation earlier today, we can look into addressing this (and `roomId` generation) in another PR.

src/main/java/com/rocketden/main/util/Utility.java
zpChris: Ok, I autowired the `Utility` class -- I didn't change the name to `GeneratorService` because I thought that we could keep adding utility-type methods.

I also made `Utility.java` a `@Component` class so it could be autowired -- is this the proper way of doing this @alankbi?

src/main/java/com/rocketden/main/controller/v1/RoomController.java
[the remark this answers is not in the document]
zpChris: Done!

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Done! I don't think the `static` component mattered, but I removed it anyway.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Ok, I implemented this in the `UtilityTests.java` class. I renamed the other file that has the utility methods used for tests to `UtilityTestMethods.java` to better indicate its use.

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
zpChris: As talked about on the call, we will look into this on a separate PR; see the [Trello card](https://trello.com/c/CjRqLj7o/161-return-current-user-info-after-a-join-room-call).

frontend/src/components/card/PlayerCard.tsx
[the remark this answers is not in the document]
zpChris: We talked about this during the call, and we'll be answering these questions in future PRs! For reference, one of my next PRs is to reassign the host when the current one disconnects.

frontend/src/views/Lobby.tsx
```diff
@@ -162,13 +173,40 @@ function LobbyPage() {
...
+              ) : null}
+            </PlayerCard>
+          ))
+        }
+      </div>
+      <div>
+        {
+          (inactiveUsers !== null && inactiveUsers.length > 0) ? (
+            <MediumText>Below is a list of inactive users.</MediumText>
+          ) : <MediumText>There are no inactive users.</MediumText>
```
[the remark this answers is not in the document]
zpChris: I've updated this to just show the inactive / loading users as greyed out at the end of the list.

src/main/java/com/rocketden/main/util/Utility.java
alankbi: I think that's good - either Component or Service I think are appropriate

alankbi (approved):
Approved! Just left one minor optional suggestion

src/main/java/com/rocketden/main/util/Utility.java
zpChris: Ok perfect! -- I'll keep `@Component` then.

frontend/src/views/Lobby.tsx
```diff
@@ -194,14 +206,33 @@ function LobbyPage() {
...
+              {currentUser?.nickname === host?.nickname
+              && (user.nickname !== currentUser?.nickname) ? (
+                // If currentUser is host, pass in an on-click action card for all other users
+                <HostActionCard
+                  user={user}
+                  onMakeHost={changeHosts}
+                  onDeleteUser={deleteUser}
+                />
+              ) : null}
+            </PlayerCard>
```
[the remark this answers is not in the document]
zpChris: I did the same, but extracted the whole list into a separate block; let me know if that looks good!

helenli8181 (approved):
The new changes look great! Approved :)

PR #54, opened 2020-10-05
Allow hosts to transfer ownership

alankbi:
### List of Changes:

* Add ability for hosts to transfer ownership to other users
* Refactor UserNicknameText to a Card component
* Create action popover component for when hosts click on a name

### Notes:

* Relied heavily on [this](https://stackoverflow.com/a/43863886/7386515) answer to create the CSS for the popover
* Open to changing the design and structure for the PlayerCard/HostActionCard as nothing really resonated with me there

### Screencast and Images:

[Short and informal screencast](https://www.loom.com/share/4e61d0d2f76b4711a3ce606c0398f649)

### DIFF
--- a/frontend/src/components/card/HostActionCard.tsx
+++ b/frontend/src/components/card/HostActionCard.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import styled from 'styled-components';
+import { SmallActionText } from '../core/Text';
+import { User } from '../../api/User';
+
+const Content = styled.div`
+  padding: 10px;
+  background-color: ${({ theme }) => theme.colors.white};
+  border-radius: 5px;
+  border: 1px solid ${({ theme }) => theme.colors.gray};
+`;
+
+type HostActionCardProps = {
+  user: User,
+  onMakeHost: (newHost: User) => void,
+  onDeleteUser: (userToDelete: User) => void,
+};
+
+function HostActionCard(props: HostActionCardProps) {
+  const { user, onMakeHost, onDeleteUser } = props;
+
+  return (
+    <Content>
+      <div>
+        <SmallActionText onClick={() => onMakeHost(user)}>Make Host</SmallActionText>
+        <SmallActionText onClick={() => onDeleteUser(user)}>Kick</SmallActionText>
+      </div>
+    </Content>
+  );
+}
+
+export default HostActionCard;
--- a/frontend/src/components/card/HostActionCard.tsx
+++ b/frontend/src/components/card/HostActionCard.tsx
@@ -12,38 +12,14 @@ const Content = styled.div`
   
   width: 150px;
   background-color: ${({ theme }) => theme.colors.white};
-  border-radius: 5px;
-  box-shadow: 0 2px 2px rgba(0, 0, 0, 0.24);
+  border-radius: 5px 5px 0 0;
+  box-shadow: 0 -1px 4px rgba(0, 0, 0, 0.12);
   
   height: 20px;
-  padding: 10px;
+  padding: 8px;
   
-  // -(height + 2 * padding + :before-border-width / sqrt(2))
-  margin-top: -55px;
-  
-  // Create the arrow at the bottom of the popup
-  &:before {
-    content: "";
-    position: absolute;
-    width: 0;
-    height: 0;
-    left: 50%;
-    box-sizing: border-box;
-    z-index: 1;
-    box-shadow: -2px 2px 2px rgba(0, 0, 0, 0.24);
-    
-    border-style: solid;
-    border-color: transparent transparent ${({ theme }) => theme.colors.white} ${({ theme }) => theme.colors.white};
-    transform-origin: 0 0;
-    transform: rotate(-45deg);
-    
-    border-width: 8px;
-    
-    // -(sqrt(2) * border-width)
-    margin-left: -12px;
-    // -(2 * border-width)
-    bottom: -16px;
-  }
+  // -(height + 2 * padding - 3px)
+  margin-top: -36px;
 `;
 
 type HostActionCardProps = {
--- a/frontend/src/components/card/HostActionCard.tsx
+++ b/frontend/src/components/card/HostActionCard.tsx
@@ -4,16 +4,46 @@ import { SmallActionText } from '../core/Text';
 import { User } from '../../api/User';
 
 const Content = styled.div`
+  // Center div above parent
   position: absolute;
-  left: 0;
-  right: 0;
-  margin-left: auto;
-  margin-right: auto;
+  top: 0;
+  left: 50%;
+  transform: translate(-50%, 0);
   
-  padding: 10px;
+  width: 150px;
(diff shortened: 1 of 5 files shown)


alankbi:
[Implement host change functionality on frontend](https://trello.com/c/qFm2kDfT/150-implement-host-change-functionality-on-frontend)

frontend/src/components/card/HostActionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+    height: 0;
+    left: 50%;
+    box-sizing: border-box;
+    z-index: 1;
+    box-shadow: -2px 2px 2px rgba(0, 0, 0, 0.24);
+    
+    border-style: solid;
+    border-color: transparent transparent ${({ theme }) => theme.colors.white} ${({ theme }) => theme.colors.white};
+    transform-origin: 0 0;
+    transform: rotate(-45deg);
```
zpChris, line 38: When is this line used?

frontend/src/components/card/HostActionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+    
+    border-width: 8px;
+    
+    // -(sqrt(2) * border-width)
+    margin-left: -12px;
+    // -(2 * border-width)
+    bottom: -16px;
```
zpChris, line 45: Are these comments still true after taking and updating the room details?

frontend/src/components/card/HostActionCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onDeleteUser: (userToDelete: User) => void,
+};
+
+function HostActionCard(props: HostActionCardProps) {
+  const { user, onMakeHost, onDeleteUser } = props;
+
+  return (
+    <Content>
+      <SmallActionText onClick={() => onMakeHost(user)}>Make Host</SmallActionText>
+      <SmallActionText onClick={() => onDeleteUser(user)}>Kick</SmallActionText>
```
zpChris, line 61: Thoughts on bolding the text or something on hover, so the host knows which action they're doing?

frontend/src/components/card/PlayerCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const toggleActionCard = () => {
+    // Only toggle showActionCard if this user is the host
+    if (actionCard) {
+      setShowActionCard(!showActionCard);
+    }
+  };
+
+  return (
+    <Content onClick={toggleActionCard}>
```
zpChris, line 38: Could we do this onhover of the card, instead of onclick?

frontend/src/views/Lobby.tsx
```diff
@@ -114,13 +136,25 @@ function LobbyPage() {
...
       <div>
         {
           users?.map((user) => (
-            <UserNicknameText onClick={() => deleteUser(user)}>
-              {user.nickname}
-              {user.nickname === host?.nickname ? ' (host)' : ''}
-            </UserNicknameText>
+            <PlayerCard
+              user={user}
+              isHost={user.nickname === host?.nickname}
```
zpChris, line 146: This works for now, and honestly should work in the future as well, but I just wanted to note that potentially in the future this could be replaced by an equals-type function.

zpChris (approved):
Overall, LGTM! I left comments and a few nits, the main one in my mind being that I think it's more natural for these cards to appear and disappear on hover. If you end up doing this, please re-request a review from me; however, I can see hover causing problems with not being able to reach the card above, so if you don't implement this no problem.

frontend/src/components/card/HostActionCard.tsx
alankbi: It's used to create the arrow look for the popup (idk exactly how it works but my guess is it rotates a normal square div which causes that downwards pointy look)

frontend/src/components/card/HostActionCard.tsx
alankbi: Not sure what taking and updating the room details means, but those comments are what I experimented/guessed to be correct (just in case in the future if we wanted to change the size of everything we'd know how to make it look correct)

frontend/src/components/card/HostActionCard.tsx
alankbi: Sounds good, added an underline effect for good measure (I can remove it if it doesn't look good tho)

frontend/src/components/card/PlayerCard.tsx
alankbi: Nice suggestion! I implemented this, although it required a slight CSS gimmick involving an invisible border in order to get it to work (there's a gap between the player name and action card, so without some invisible content, moving the cursor up towards the action card would cause it to leave the component hide the action card before you can use it). If that looks bad, lmk and I can undo it.

frontend/src/components/card/HostActionCard.tsx
zpChris: Ah, that makes sense -- thanks!

frontend/src/components/card/HostActionCard.tsx
zpChris: I think it looks great, thanks for making that change!

frontend/src/components/card/PlayerCard.tsx
zpChris: Yeah I figured that's why this would be difficult, and possible why you didn't implement it at first -- good solution though, I like it! I think it looks great 👍

zpChris:
Just one more nit I noticed -- the "hand" cursor is still present on the text of all people, whereas it's not really a button anymore. I'm not sure what the best practice is here (maybe only keep the hand from the host's view on all non-host participants?), but I wanted to point it out either way.

helenli8181 (commented):
Looks great to me! I have a suggestion though -- for the action card, I think it would look better smaller in proportion to the size of the text. The pointy corner under it also doesn't seem necessary; if we get rid of that one idea is to make the card look like a Google Chrome tab where it's attached to the player's name box on one side. This would also make it possibly small enough to avoid blocking other boxes above it. Not a necessary change, but what do you think of this?

alankbi:
Sounds interesting, tho I'm not too sure if I understand the proposed design. If you want to sketch something up I can take a look and try to implement it if that sounds good; either way works for me

helenli8181:
@alankbi Here's a rough sketch of what I'm imagining! 

IMG_0110

alankbi:
How does this look: 

Screenshot

alankbi:
I haven't been able to figure out how to get it to look good for small usernames tho: 

Screenshot

helenli8181:
I think that looks awesome! Yes haha I guess it's more awkward for short nicknames; I'm not really sure what to do about that either. Someone mentioned a design where you have a little drop down arrow on each user's block that you have to click to see the menu. That would eliminate this issue, and the problem of having the menu cover other blocks annoyingly when you move the mouse around. But if that's a big change we could leave it for later when aesthetics are more important.

helenli8181 (approved):
Looks great!

PR #55, opened 2020-10-07
Create endpoint for updating room difficulty setting

alankbi:
### List of Changes:

* Add PUT endpoint for updating room difficulty
* Create enum class for accepted difficulty values (easy, medium, and hard)
* Include appropriate tests and error handling

### Notes:

* How to handle enums in entity classes [[1]](https://stackoverflow.com/questions/14437384/enum-field-in-jpa-entity)
* How to convert request body strings into enum values [[2]](https://stackoverflow.com/questions/55399593/how-to-custom-convert-string-to-enum-in-requestbody)

### Screencast and Images:

Difficulty column in database:
Difficulty column in database

### DIFF
--- a/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
+++ b/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
@@ -1,7 +1,7 @@
 package com.rocketden.main.dto.room;
 
 import com.rocketden.main.dto.user.UserDto;
-import com.rocketden.main.model.ProbemDifficulty;
+import com.rocketden.main.model.ProblemDifficulty;
 import lombok.Getter;
 import lombok.Setter;
 
@@ -9,5 +9,5 @@ import lombok.Setter;
 @Setter
 public class UpdateSettingsRequest {
     private UserDto initiator;
-    private ProbemDifficulty difficulty;
+    private ProblemDifficulty difficulty;
 }
--- a/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
+++ b/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
@@ -1,7 +1,7 @@
 package com.rocketden.main.dto.room;
 
 import com.rocketden.main.dto.user.UserDto;
-import com.rocketden.main.model.Difficulty;
+import com.rocketden.main.model.ProbemDifficulty;
 import lombok.Getter;
 import lombok.Setter;
 
@@ -9,5 +9,5 @@ import lombok.Setter;
 @Setter
 public class UpdateSettingsRequest {
     private UserDto initiator;
-    private Difficulty difficulty;
+    private ProbemDifficulty difficulty;
 }
--- a/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
+++ b/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
@@ -1,6 +1,7 @@
 package com.rocketden.main.dto.room;
 
 import com.rocketden.main.dto.user.UserDto;
+import com.rocketden.main.model.Difficulty;
 import lombok.Getter;
 import lombok.Setter;
 
@@ -8,5 +9,5 @@ import lombok.Setter;
 @Setter
 public class UpdateSettingsRequest {
     private UserDto initiator;
-    private String difficulty;
+    private Difficulty difficulty;
 }
--- a/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
+++ b/src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
@@ -1,0 +1,12 @@
+package com.rocketden.main.dto.room;
+
+import com.rocketden.main.dto.user.UserDto;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class UpdateSettingsRequest {
+    private UserDto initiator;
+    private String difficulty;
+}
--- a/src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
+++ b/src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
@@ -1,8 +1,13 @@
 package com.rocketden.main.exception.api;
 
+import com.rocketden.main.exception.RoomError;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.HttpMessageNotReadableException;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
 import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 
 @ControllerAdvice
@@ -12,6 +17,15 @@ public class ApiExceptionHandler extends ResponseEntityExceptionHandler {
     @ExceptionHandler(ApiException.class)
     protected ResponseEntity<ApiErrorResponse> handleApiException(ApiException e) {
         ApiError apiError = e.getError();
+        return new ResponseEntity<>(apiError.getResponse(), apiError.getStatus());
+    }
+
+    // Handle failure to convert from string to enum (e.g. "easy" -> Difficulty.EASY)
+    @Override
+    protected ResponseEntity<Object> handleHttpMessageNotReadable(HttpMessageNotReadableException ex,
+                                                                  HttpHeaders headers, HttpStatus status,
+                                                                  WebRequest request) {
+        ApiError apiError = RoomError.BAD_SETTING;
         return new ResponseEntity<>(apiError.getResponse(), apiError.getStatus());
     }
 }
--- a/src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
+++ b/src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
@@ -25,7 +25,13 @@ public class ApiExceptionHandler extends ResponseEntityExceptionHandler {
     protected ResponseEntity<Object> handleHttpMessageNotReadable(HttpMessageNotReadableException ex,
                                                                   HttpHeaders headers, HttpStatus status,
                                                                   WebRequest request) {
-        ApiError apiError = RoomError.BAD_SETTING;
-        return new ResponseEntity<>(apiError.getResponse(), apiError.getStatus());
+
+        // Check for a RoomError.BAD_SETTING triggered in the Difficulty.fromString method
+        if (ex.getCause() != null && ex.getCause().getCause() instanceof ApiException) {
+            ApiError apiError = RoomError.BAD_SETTING;
+            return new ResponseEntity<>(apiError.getResponse(), apiError.getStatus());
+        }
+
+        return super.handleHttpMessageNotReadable(ex, headers, status, request);
     }
 }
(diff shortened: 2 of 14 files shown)


alankbi:
[Enable game settings (e.g.; Problem Difficulty) on rooms](https://trello.com/c/qZm8qLOa/138-enable-game-settings-eg-problem-difficulty-on-rooms)

src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
```diff
@@ -0,0 +1,13 @@
...
+import com.rocketden.main.dto.user.UserDto;
+import com.rocketden.main.model.Difficulty;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class UpdateSettingsRequest {
+    private UserDto initiator;
+    private Difficulty difficulty;
```
helenli8181, line 12: In the future when we have more settings for the host to toggle, would there only be one request that contains values for every type of setting, or a different request for each one?

helenli8181 (approved):
Looks good to me! I just had a couple of small clarifying questions.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -128,7 +129,30 @@ public RoomDto updateRoomHost(String roomId, UpdateHostRequest request) {
...
+        if (!room.getHost().equals(initiator)) {
+            throw new ApiException(RoomError.INVALID_PERMISSIONS);
+        }
+
+        // Set new difficulty value
+        room.setDifficulty(request.getDifficulty());
+        repository.save(room);
+
+        RoomDto roomDto = RoomMapper.toDto(room);
+        sendSocketUpdate(roomDto);
```
zpChris, line 154: As it currently stands, the problem difficulty (if I understand correctly) is not reflected on the frontend, so no update will be triggered from this socket message.

That being said, I'm assuming that will be the next PR step, so I don't really have a problem with this -- just wanted to point it out.

src/test/java/com/rocketden/main/api/RoomTests.java
```diff
@@ -445,6 +450,164 @@ public void changeRoomHostNewHostNotFound() throws Exception {
...
+        MvcResult result = this.mockMvc.perform(put(String.format(PUT_ROOM_SETTINGS, room.getRoomId()))
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(Utility.convertObjectToJsonString(updateRequest)))
+                .andDo(print()).andExpect(status().isOk())
+                .andReturn();
+
+        String jsonResponse = result.getResponse().getContentAsString();
+        room = Utility.toObject(jsonResponse, RoomDto.class);
+
+        assertNull(room.getDifficulty());
```
zpChris, line 510: Should the `room` be assigned a default difficulty? I'm open to whatever on this -- it might help avoid having a bunch of Null Pointer Exceptions in our code, but null might be useful if we want to show an "undecided" state on the frontend.

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
```diff
@@ -159,4 +161,22 @@ public void socketReceivesMessageOnHostChange() throws Exception {
         assertEquals(expected.getUsers(), actual.getUsers());
     }
 
+    @Test
+    public void socketReceivesMessageOnSettingsChange() throws Exception {
```
zpChris, line 165: If the difficulty is just set to its current setting, should we still send an update?

I think it'd be fine to, but I just wanted to bring up the question.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -253,6 +263,69 @@ public void changeRoomHostFailure() {
...
+
+        assertEquals(request.getDifficulty(), response.getDifficulty());
+
+        verify(template).convertAndSend(
+                eq(String.format(BaseRestController.BASE_SOCKET_URL + "/%s/subscribe-user", response.getRoomId())),
+                eq(response));
+    }
+
+    @Test
+    public void updateRoomSettingsFailure() {
```
zpChris, line 294: Can we separate this test into separate ones, as done before?

src/test/java/com/rocketden/main/api/RoomTests.java
```diff
@@ -445,6 +450,164 @@ public void changeRoomHostNewHostNotFound() throws Exception {
         assertEquals(ERROR.getResponse(), actual);
     }
 
+    @Test
+    public void updateRoomSettingsSuccess() throws Exception {
+        UserDto host = new UserDto();
+        host.setNickname("host");
+        UserDto user = new UserDto();
+        user.setNickname("test");
```
zpChris, line 458: The user is not used here, right? Could you remove these lines?

zpChris (changes requested):
Requesting changes, but really it looks good -- I just had a few clarifying nits and potential changes on the tests. 👍

src/main/java/com/rocketden/main/model/Difficulty.java
```diff
@@ -0,0 +1,19 @@
...
+import com.rocketden.main.exception.api.ApiException;
+
+public enum Difficulty {
+    EASY, MEDIUM, HARD;
+
+    // Convert a matching string (ignoring case) to enum object
+    @JsonCreator
+    public static Difficulty fromString(String value) {
+        try {
+            return Difficulty.valueOf(value.toUpperCase());
```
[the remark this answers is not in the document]
alankbi: Mainly a precaution to support any mixture of say EASY, easy, Easy, EaSY, etc. If we want to lock this only to exact matches I'm okay with that too, just lmk

src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
alankbi: That's a good question - my thought would be that it'd all be in one request that way we wouldn't have a bunch of very similar endpoints and associated tests, but I'm okay either way

src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
```diff
@@ -14,4 +19,19 @@
...
+                                                                  HttpHeaders headers, HttpStatus status,
+                                                                  WebRequest request) {
+
+        // Check for a RoomError.BAD_SETTING triggered in the Difficulty.fromString method
+        if (ex.getCause() != null && ex.getCause().getCause() instanceof ApiException) {
+            ApiError apiError = RoomError.BAD_SETTING;
+            return new ResponseEntity<>(apiError.getResponse(), apiError.getStatus());
+        }
+
+        return super.handleHttpMessageNotReadable(ex, headers, status, request);
```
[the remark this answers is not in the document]
alankbi: I could be wrong but I think it also results anytime when your request body content is empty, as in the following:

Screenshot

There might be other cases as well, but I'm not too sure. If you think it's good I can also make a custom error for this situation too like EMPTY_CONTENT or something (although idk if that might mess up some other edge cases not yet discovered)

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -128,7 +129,30 @@ public RoomDto updateRoomHost(String roomId, UpdateHostRequest request) {
...
+    public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request) {
+        Room room = repository.findRoomByRoomId(roomId);
+
+        // Return error if room could not be found
+        if (room == null) {
+            throw new ApiException(RoomError.NOT_FOUND);
+        }
+
+        // Return error if the initiator is not the host
+        User initiator = UserMapper.toEntity(request.getInitiator());
```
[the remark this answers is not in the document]
alankbi: Not sure about other invalid scenarios, but for non-ascii characters it should trigger the invalid permissions exception on the next line (based on personal reasoning and one very non-exhaustive test on Postman, so I might be wrong on that). If there are other cases you think I should test for lmk and I can add some stuff

src/test/java/com/rocketden/main/api/RoomTests.java
alankbi: Yeah I was thinking that a null value represents no difficulty selected, as I think that might be a good default esp. if people don't care as much about advanced settings. Lmk your thoughts on this tho

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
alankbi: Personally I'm okay with either, if no settings change then it'd just send a message and the front-end would ignore it (I think changing room hosts to the same person causes the same thing). I'd be open to implementing a check tho, just lmk

src/test/java/com/rocketden/main/api/RoomTests.java
alankbi: It's only used in the call to `setUpRoomWithTwoUsers`. The second user isn't actually necessary for this test case tho, so lmk if you want and I can create another setup method that just creates a room with a single host, as that might make more sense

src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
zpChris: Yeah I think that's a good idea, preferably it'd be just one endpoint, I think that'd be a simpler setup.

src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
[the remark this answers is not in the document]
zpChris: I'm just thinking of how the JavaScript will handle this situation. If a bug occurs where the body content is empty due to some error of not being sent properly, we'll receive nothing. I can easily see this causing an error on the frontend, as I believe this would go into the `then` response, which assumes that the result is of a certain format.

I think `EMPTY_CONTENT` might be necessary because of this -- but let me know what you think. Should we handle the case above, and if so, how should we do that?

src/test/java/com/rocketden/main/api/RoomTests.java
zpChris: Yeah I actually think that's good, you should keep this 👍

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
zpChris: No, I don't think there's any need for that, just wanted to bring it up in case. Thanks!

src/test/java/com/rocketden/main/api/RoomTests.java
zpChris: Yeah I think that'd be helpful, if possible -- thanks!

src/main/java/com/rocketden/main/exception/api/ApiExceptionHandler.java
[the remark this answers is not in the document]
zpChris: Actually, scratch that -- I realized that anytime we have just `{}` at the very least, it will still return `INVALID_PERMISSIONS`. I'm free to have a conversation on this, but I don't think this is an edge case we have to account for.

src/main/java/com/rocketden/main/dto/room/UpdateSettingsRequest.java
```diff
@@ -0,0 +1,13 @@
...
+import com.rocketden.main.dto.user.UserDto;
+import com.rocketden.main.model.ProbemDifficulty;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class UpdateSettingsRequest {
+    private UserDto initiator;
+    private ProbemDifficulty difficulty;
```
zpChris, line 12: Spelling? @alankbi 😁 

At least you get a lot of points for consistency 💯

zpChris (changes requested):
I'll approve after you add the new "set up room with one user" test, and fix the spelling issue with `ProbemDifficulty`. 😀 

Just requesting changes mainly to verify that it will work for me too.

alankbi:
Lmao my bad haha 

One of the downsides of using the IDE's one-click refactor button ¯\\\_(ツ)\_/¯

zpChris (approved):
LGTM! Thanks for making these changes 👍

PR #58, opened 2020-10-13
Update the REST endpoints to include roomId as a path variable.

zpChris:
### List of Changes:

- Update the `getRoom`, `joinRoom`, and `createRoom` endpoints to include the `roomId` as a path variable.
- Update the corresponding tests to use path variables for the applicable methods.

### Notes:

- This PR is used just to clean up the codebase -- there are no changes to the actual code behavior.

### Screencast and Images:

There are no images or screencasts attached to this PR, as there are no changes to the UI.


frontend/src/api/Room.ts
```diff
@@ -13,7 +13,6 @@ export type CreateRoomParams = {
 };
 
 export type JoinRoomParams = {
-  roomId: string,
   user: User,
```
helenli8181, line 16: This isn't much of an issue but is it weird that there is only one field in the params? This also happens in several other places since we took out roomId as a path variable.

helenli8181 (approved):
Looks good to me! Having the roomId in all the paths looks a lot more consistent :)

frontend/src/api/Room.ts
zpChris: Yeah I think this is fine, I don't think the number of fields in the params really matters too much, the object does what its supposed to.

alankbi (commented):
Overall looks good! Before approving I just wanted to poll your thoughts on the naming conventions for the join and create endpoints:

1. For the create endpoint, I'd still prefer if we keep it as `/rooms` to keep with [convention](https://restfulapi.net/resource-naming/) if that seems reasonable
2. Similarly, for join, if not keeping it as `/rooms/{roomId}`, then maybe another alternative would be `/rooms/{roomId}/users`?

My main reasoning for this (from the article) is to keep the URL as the noun/object you're acting on, and using the method to specify what action you want to take on that object (GET/POST/PUT). If you want tho we can talk more about it during standup 🙂

alankbi (approved):

zpChris:
> Overall looks good! Before approving I just wanted to poll your thoughts on the naming conventions for the join and create endpoints:
> 
> 1. For the create endpoint, I'd still prefer if we keep it as `/rooms` to keep with [convention](https://restfulapi.net/resource-naming/) if that seems reasonable
> 2. Similarly, for join, if not keeping it as `/rooms/{roomId}`, then maybe another alternative would be `/rooms/{roomId}/users`?
> 
> My main reasoning for this (from the article) is to keep the URL as the noun/object you're acting on, and using the method to specify what action you want to take on that object (GET/POST/PUT). If you want tho we can talk more about it during standup 🙂

These changes make sense -- all fixed! Thanks for attaching that link 👍

PR #61, opened 2020-10-15
Added Active variable to room when game starts

helenli8181:
### List of Changes:

- Added `isActive` variable to roomDto and room model.
- Users who somehow end up in the lobby after the game is already active are redirected to game.

### Notes:

- Since we haven't banned users from joining mid-game it's easy to test by adding new users after the host starts game. 
- Should I add tests for this?

### Screencast and Images:

No UI changes


frontend/src/api/Room.ts
```diff
@@ -9,6 +9,7 @@ export type Room = {
   users: [User],
   activeUsers: [User],
   inactiveUsers: [User],
+  isActive: boolean;
```
alankbi, line 12: nit: I think this should be a comma

frontend/src/views/Lobby.tsx
```diff
@@ -215,7 +217,10 @@ function LobbyPage() {
     if (!socketConnected && currentRoomId && currentUser && currentUser.userId) {
       connectUserToRoom(currentRoomId, currentUser.userId);
     }
-  }, [socketConnected, connectUserToRoom, currentRoomId, currentUser]);
+    if (isActive) {
+      history.push('/game');
+    }
```
alankbi, line 222: nit: I would move this into its own useEffect block that depends only on `isActive`

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -47,6 +47,8 @@
     @Setter(AccessLevel.PRIVATE)
     private List<User> users = new ArrayList<>();
 
+    private Boolean isActive;
```
alankbi, line 50: Maybe set a default value of false? That way we don't have to deal with null values

src/main/java/com/rocketden/main/service/GameService.java
```diff
@@ -40,6 +40,9 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
 			throw new ApiException(RoomError.INVALID_PERMISSIONS);
 		}
 
+		room.setIsActive(true);
+		repository.save(room);
+
 		RoomDto roomDto = RoomMapper.toDto(room);
 
 		String socketPath = String.format(START_GAME_SOCKET_PATH, roomId);
```
alankbi, line 48: Suggestion: change this to send the socket update on the usual endpoint (`subscribe-user`), that way we don't need a separate endpoint just for starting a game

alankbi (changes requested):
Looks good! Before approving, I had two suggestions:

1. Now that we have an `isActive` parameter, can we delete the `start-game` socket endpoint (and any associated code) and just use the `subscribe-user` one as usual? Since you can just check for when `isActive` switches to true and then redirect everyone once that happens
2. To respond to your question, yeah I think some tests would be good - if you decide to go through with suggestion 1, I think you can pretty much just modify the existing tests in `GameTests` and `GameServiceTests`, and then add one test in `RoomSocketTests`

src/main/java/com/rocketden/main/model/Room.java
zpChris, line 50: Is there a reason this is `Boolean` with a capital-b instead of `boolean`?

zpChris (changes requested):
Thanks for making this update! I don't have any code changes beyond what @alankbi noted, but I have a few notes:
1. The expected behavior of `isActive` in my mind is that a new user **cannot** join the room after the room becomes active. You could modify the join request to handle this if you wanted.
2. If the host or user already in the game goes back to the lobby page, I think the right behavior is that they should be redirected back to the game page -- right now, they simply stay in the lobby page in that case.
3. If the host starts the game and then clicks the back button to go to the lobby page, it doesn't refresh properly -- I believe this is the same issue I had on my socket PR with `command + shift + t`. I'd recommend creating a new issue for this.
Let me know which of these changes you're open to making!

src/main/java/com/rocketden/main/model/Room.java
helenli8181: Originally it was lower case, but for some reason it couldn't compile and did not recognize `isActive` as a field of room that way. I'm not entirely sure why that is.

alankbi: Yeah I guess they need to be objects for some reason; I think the id is the same thing (Integer instead of int)

src/main/java/com/rocketden/main/service/GameService.java
```diff
@@ -15,7 +15,7 @@
 @Service
 public class GameService {
 
-	private static final String START_GAME_SOCKET_PATH = BaseRestController.BASE_SOCKET_URL + "/%s/start-game";
+	private static final String START_GAME_SOCKET_PATH = BaseRestController.BASE_SOCKET_URL + "/%s/subscribe-user";
```
zpChris, line 18: I don't think you need this constant; you can just get the one in `WebSocketConfig.java` as `WebSocketConfig.SOCKET_PATH`.

zpChris (changes requested):
Right now, the update doesn't seem to work for me locally? I attached a screencast of how it currently works.

[Screencast of the Active Parameter Not Causing Redirect](https://www.loom.com/share/41f7dde2b173452385c87f05b5ee2a07)

helenli8181:
> You might have already seen/come across this @helenli8181, but when using booleans the lombok getters sometime have weird names (getActive vs. isActive vs. getIsActive): [post](https://stackoverflow.com/questions/42619986/lombok-annotation-getter-for-boolean-field)
> 
> Just thought to bring it up in case that ever becomes an issue

Yeah the issue I was trying to debug today had to do with the naming! I think there's a naming convention when using `template.convertandsend()` that automatically removes the verb so that `isActive` becomes `active` which doesn't match up in the front end, and the redirect doesn't happen. There's probably some way to suppress that but for now I've just renamed it to `active`.

zpChris (changes requested):
This looks great! I left two small nits, but I'm requesting changes because I think you could update the `Join.tsx` file so that people do not get past the first stage of the active room (entering the `roomId`). I left a screencast to show what I'm talking about: **[HERE](https://www.loom.com/share/5dd4db347b424a10897eada0a1336056)**.

The relevant change would likely be on the `checkRoom` function in `Join.tsx` by the way; let me know what you think, or if I can clarify any better.

frontend/src/views/Lobby.tsx
```diff
@@ -216,7 +209,14 @@ function LobbyPage() {
...
   }, [socketConnected, connectUserToRoom, currentRoomId, currentUser]);
-
+  
+  // Redirect user to game page if room is active.
+  useEffect(() => {
+    if (active) {
+      history.push('/game');
+    }
+  }, [history, active]);
+
```
zpChris, line 219: Are you seeing the same errors that are in this attached image? It's just deleting spaces, but this should show up as ESLint warnings on your editor (I think we have the same setup). Either way, small nit just to delete those spaces.

ESLint Space Warnings

src/test/java/com/rocketden/main/service/GameServiceTests.java
```diff
@@ -1,5 +1,6 @@
 package com.rocketden.main.service;
 
+import com.rocketden.main.config.WebSocketConfig;
 import com.rocketden.main.controller.v1.BaseRestController;
```
zpChris, line 4: You don't use `WebSocketConfig` or `BaseRestController`, so remove these imports.

Similar to before, do you get these warnings?

Unused Import Warnings

src/test/java/com/rocketden/main/socket/RoomSocketTests.java
```diff
@@ -17,6 +18,8 @@
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.messaging.converter.MappingJackson2MessageConverter;
```
zpChris, line 22: You don't use either of these, same as the other nits.

zpChris (changes requested):
The changes look great! I left _tiny_ nits because I'm not sure if you're getting the same ESLint or Java Styling warnings I do (screenshots attached on below comments). Otherwise, this looks great!

zpChris (approved):
LGTM!

frontend/src/views/Lobby.tsx
```diff
@@ -255,7 +255,7 @@ function LobbyPage() {
       <br />
 
       {currentUser?.nickname === host?.nickname
-        ? <PrimaryButton onClick={handleStartGame} disabled={loading}>Start Game</PrimaryButton>
+        ? <PrimaryButton onClick={handleStartGame}>Start Game</PrimaryButton>
```
alankbi, line 258: nit: Re-add in this line that disables the start button when something is loading

alankbi (approved):
Looks great! Just left one nit but otherwise approved

PR #57, opened 2020-10-13
Allow users to choose coding language

alankbi:
### List of Changes:

* Allow users to toggle coding language (Java, Python, JavaScript, C#)
* Enable syntax highlighting 
* Add default code for each language

### Notes:

* [Source](https://github.com/react-monaco-editor/react-monaco-editor#usage-with-create-react-app) behind config file and `react-app-rewired`
* Syntax validation works but IntelliSense and error validation don't seem to be well-documented with the Monaco editor we're using

### Screencast and Images:

Game page with code editor

PR #62, opened 2020-10-16
Allow users to choose coding language

alankbi:
### List of Changes:

* Allow users to toggle coding language (Java, Python, JavaScript, C#)
* Enable syntax highlighting 
* Add default code for each language

### Notes:

* [Source](https://github.com/react-monaco-editor/react-monaco-editor#usage-with-create-react-app) behind config file and `react-app-rewired`
* Syntax validation works but IntelliSense and error validation don't seem to be well-documented with the Monaco editor we're using

### Screencast and Images:

Game page with code editor

_Duplicated from #57, which fails to build on GitHub Actions for an unknown reason._

### DIFF
--- a/frontend/src/components/core/Editor.tsx
+++ b/frontend/src/components/core/Editor.tsx
@@ -1,11 +1,19 @@
-import React from 'react';
-import MonacoEditor, { MonacoEditorProps } from 'react-monaco-editor';
+import React, { useState } from 'react';
+import MonacoEditor from 'react-monaco-editor';
+import styled from 'styled-components';
 
-export const languages = {
+type LanguageType = {
+  [key: string]: {
+    name: string,
+    defaultCode: string,
+  }
+};
+
+export const languages: LanguageType = {
   java: {
     name: 'Java',
     defaultCode:
-      'public class Solution {\n\n'
+      'public class Solution {\n'
       + '    public static void main(String[] args) {\n'
       + '        \n'
       + '    }\n'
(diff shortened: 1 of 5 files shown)


alankbi:
[Allow users to change code editor language](https://trello.com/c/7acGLrch/65-allow-users-to-change-code-editor-language)

alankbi:
[Allow users to change code editor language](https://trello.com/c/7acGLrch/65-allow-users-to-change-code-editor-language)

frontend/src/components/core/Editor.tsx
```diff
@@ -1,9 +1,61 @@
...
+import styled from 'styled-components';
+
+type LanguageType = {
+  [key: string]: {
+    name: string,
+    defaultCode: string,
+  }
+};
+
+export const languages: LanguageType = {
```
zpChris, line 12: Could the `defaultCode` be updated whenever the user types new things?

This could just be for later, but I'm thinking that if the user types some things in Java then briefly switches to another language, their old code shouldn't be deleted.

This could also be implemented when other things like saving code even after refresh are built.

zpChris (approved):
This code looks great; I only left one optional comment for saving the code -- if you do this, I'll review again, but it's just optional so otherwise this LGTM!

frontend/src/components/core/Editor.tsx
alankbi: That's a great idea - just made the change :)

helenli8181 (approved):
This looks great to me!

zpChris (approved):
Thanks for making that change, LGTM! I tested that the code editor could save characters like `Â`, and it did perfectly, so yay 💯 

In the future, I think we can add a warning or something before the user refreshes; or, alternatively, save the latest version of the code in localStorage or `location`. For now however, I think this looks great!

PR #63, opened 2020-10-17
Test socket connection and disconnection, and add user / room constants to testing files.

zpChris:
### List of Changes:
- Add constants to represent user and room attributes for the testing files.
- Fix the test in `SocketService.java` to run (the `@Test` annotation had the wrong import previously).
- Remove warnings on `assertThrows` tests by using `try / catch` blocks.
- @alankbi's input: Add tests to `RoomSocketTests.java` to ensure that a socket message is sent when a user connects and disconnects.

### Notes:
- Later, I think we may be able to add comments to these tests and order them or structure them into groups for better readability.
- Some of the `assertThrows` declarations caused warnings in the `RoomServiceTests.java`, while others didn't; I'm not completely certain of why this is, but I fixed the warnings that were present by using `try / catch`.
- In the `WebSocketConnectionEvents.java` class, I added a few if statements that do the following:
  - If the `userId` is not added in headers, no `sessionId` is added. In reality, I don't know what the expected behavior should be, since if the `userId` isn't provided then we can't tell which user connected. I don't know when this issue would happen, but I just wanted to note it in case.
  - If the `sessionId` passed in on disconnect is not associated with any user, then I do nothing. This is a similar issue, we don't know who just disconnected. I'm again not sure when this may arise, but I wanted to note it down here.

### Screencast and Images:
There are no updates to the UI, only tests, so no screencasts or images are needed.


src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -213,36 +215,37 @@ public void changeRoomHostSuccess() {
...
         invalidPermRequest.setNewHost(UserMapper.toDto(host));
 
-        ApiException exception = assertThrows(ApiException.class, () ->
-                roomService.updateRoomHost(room.getRoomId(), invalidPermRequest));
-        assertEquals(RoomError.INVALID_PERMISSIONS, exception.getError());
+        try {
+            roomService.updateRoomHost(room.getRoomId(), invalidPermRequest);
+        } catch (ApiException e) {
+            assertEquals(RoomError.INVALID_PERMISSIONS, e.getError());
+        }
```
alankbi, line 241: What happens if the `updateRoomHost` call succeeds without throwing an error? 

Also out of curiosity, what was the warning you got with the `assertThrows` here (I remember I also had some warning here that I fixed when writing these, so maybe this one is similar)? Main reason I ask is cause if there's an easy fix then I might prefer keeping the `assertThrows` over using try/catch blocks for cleanliness and that (potential) bug mentioned above

alankbi (commented):
Looks good, just a few nits and questions. 

Also, before I approve, I think it'd probably be good to get some feedback on `RoomSocketTests` and `SocketTestMethods` as well

src/test/java/com/rocketden/main/mapper/RoomMapperTests.java
```diff
@@ -21,16 +21,26 @@
...
     public void entityToDto() {
         User host = new User();
-        host.setNickname("rocket");
+        host.setNickname(NICKNAME);
+        host.setUserId(USER_ID);
         User user = new User();
-        user.setNickname("test");
-        user.setSessionId("678910");
+        user.setNickname(NICKNAME_2);
+        user.setNickname(USER_ID_2);
```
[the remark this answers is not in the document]
zpChris: Haha imagine if I said no, that's right as-is 👍 

Good catch 🙂

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
```diff
@@ -14,6 +14,12 @@
 @SpringBootTest
```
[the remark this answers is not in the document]
zpChris: Ok, that works -- will do 👍

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Attached is an image of the `assertThrows(...)` error; I think I actually misinterpreted it, so thanks for bringing this up. The error is that there were multiple method calls, and it wasn't clear which one threw the error -- I've updated these lines to avoid the errors while not doing `try / catch`.

AssertThrows Error

zpChris: Fixed! Just had to replace `room.getRoomId()` with `ROOM_ID` in all instances.

zpChris:
@alankbi -- addressed your comments, also your changes look great 👍

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
```diff
@@ -62,21 +72,21 @@ public void getEquivalentUserSucceeds() {
...
 
         assertNull(room.getEquivalentUser(userToGet));
 
-        userToGet.setNickname("test");
-
-        User actual = room.getEquivalentUser(userToGet);
+        User actual = room.getEquivalentUser(user);
 
         assertEquals(user.getNickname(), actual.getNickname());
-        assertEquals(user.getId(), actual.getId());
```
alankbi, line 80: nit: re-add this line in line with the above comment

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
```diff
@@ -62,21 +72,21 @@ public void getEquivalentUserSucceeds() {
...
+        userToGet.setNickname(NICKNAME_2);
+        userToGet.setId(ID_2);
+        userToGet.setUserId(USER_ID_2);
 
         assertNull(room.getEquivalentUser(userToGet));
 
-        userToGet.setNickname("test");
-
-        User actual = room.getEquivalentUser(userToGet);
+        User actual = room.getEquivalentUser(user);
```
alankbi, line 87: nit: Keep the original code where it modifies `userToGet` to have the same `userId`/`nickname` but different id, so that this test can verify that the equivalent user has the same `id` values

So like if `room.users` contains `user{id=1, uid='test'}`, then calling `getEquivalenteUsers` with `otherUser{id=null, uid='test'}` will return the correct object reference of `user{id=1, uid='test'}`

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
```diff
@@ -41,19 +50,20 @@ public void removeUserFromRoomSet() {
...
+        userToRemove.setNickname(NICKNAME_2);
+        userToRemove.setId(ID_2);
+        userToRemove.setUserId(USER_ID_2);
 
         assertFalse(room.removeUser(userToRemove));
         assertTrue(room.getUsers().contains(user));
 
-        userToRemove.setNickname("test");
-        assertTrue(room.removeUser(userToRemove));
+        assertTrue(room.removeUser(user));
```
alankbi, line 66: Suggestion: instead of calling `removeUser` with `user`, modify `userToRemove` to have matching `userId` and `nickname` as `user` (but still with `ID_2`), and verify that calling `removeUser` with `userToRemove` succeeds despite the different `id` values

alankbi (approved):
Approved! Before merging though I left some nits/clarifications on the entity tests if you could take a look - lmk if any of those comments don't make sense

helenli8181 (approved):
This was quite the refactoring process haha. I think it looks pretty good!

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
zpChris: Ok, done! I can't see a case where this is actually applicable but it is good to check that our code works as expected, so thanks for this change.

src/test/java/com/rocketden/main/entity/RoomEntityTests.java
zpChris: Done!

zpChris:
Note: Initially, the Maven build had some timeout exceptions, but the tests pass now and work correctly locally so I'm gonna merge this.

PR #64, opened 2020-10-17
Transfer room ownership after host leaves, update through socket message, and add relevant tests.

zpChris:
### List of Changes:
- When a user disconnects from the socket, do the following:
  - Determine whether the user is the room host. If not, continue as normal; if they are, then do the following:
  - Get the first active non-host user in the room.
  - If an active non-host user is found in the room, then make this user the host through the `roomService.updateRoomHost(...)` method.
- Add test to ensure that in a room with two active users, if the host disconnects, then the other user becomes the host.

### Notes:
**Important Note:** This PR merges into `TestSocketConnection`, so we should merge that before merging this branch.

- No actual changes were made to frontend files, but the changes are reflected on the frontend through the socket updates with the host change.

### Screencast and Images:
[Screencast of the Transfer Room Host](https://www.loom.com/share/94dfb55f818b473f9ffade50e6c5d454)


helenli8181 (approved):
All this looks great to me!

alankbi (approved):
Looks great! Approved

zpChris:
This PR had failing tests a couple times with issues from `blockingQueue.poll(...)`, but it is working now; if we continue to encounter this error in the future, we could extend the time of poll or re-examine this problem.

PR #67, opened 2020-10-20
Add new utility generate unique id method.

zpChris:
### List of Changes:
- Update the `generateId` method to `generateUniqueId` that takes in the `idType` (`roomId` or `userId`, for now), and continues generating IDs until it finds one that does not exist.
- Update the corresponding tests and method calls throughout the codebase.

### Notes:
- I left a note on the modified method `generateUniqueId` that there is a possibility that this method runs forever, or at least for a very long time. There are `9^6`, or `531,441`, possible combinations, so right now this should not be a problem, but if we reach more users than we will want to change the `userId` generation method (the `roomId` generation method could potentially still stay the same depending on how we handle finished games).
- I am happy to update the `ROOM_ID_KEY` and `USER_ID_KEY` constants to use an enum, I was just more uncertain of how that works / when it's applicable so wanted to implement this simpler version first.
- The `joinRoomSuccess` and the `createRoomSuccess` tests interchanged the `userId` for a `roomId` instead; I fixed that mistake in this PR.
- The first time the test ran, it failed due to the same socket issues; it worked the second time.

### Screencast and Images:
There are no updates to the UI so no screencasts or images are provided.

### DIFF
--- a/src/main/java/com/rocketden/main/util/Utility.java
+++ b/src/main/java/com/rocketden/main/util/Utility.java
@@ -51,7 +51,7 @@ public class Utility {
     }
 
     // Determine if the id of idType already exists in the database.
-    public boolean idExists(String id, String idType) {
+    private boolean idExists(String id, String idType) {
         switch (idType) {
             case (ROOM_ID_KEY):
                 return roomRepository.findRoomByRoomId(id) != null;
--- a/src/main/java/com/rocketden/main/util/Utility.java
+++ b/src/main/java/com/rocketden/main/util/Utility.java
@@ -35,25 +35,29 @@ public class Utility {
          * Note: This loop could technically run forever, but there are 9^6
          * (531,441) possible combinations, so this works for now (10/20/2020).
          */
-        while (id == null || idExists(id, idType)) {
-            for (int i = 0; i < values.length; i++) {
-                int index = random.nextInt(numbers.length());
-                values[i] = numbers.charAt(index);
+        try {
+            while (id == null || idExists(id, idType)) {
+                for (int i = 0; i < values.length; i++) {
+                    int index = random.nextInt(numbers.length());
+                    values[i] = numbers.charAt(index);
+                }
+                id = new String(values);
             }
-            id = new String(values);
+            return id;
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(e);
         }
-        return id;
     }
 
     // Determine if the id of idType already exists in the database.
     private boolean idExists(String id, String idType) {
-        if (idType.equals(ROOM_ID_KEY)) {
-            return roomRepository.findRoomByRoomId(id) != null;
-        } else if (idType.equals(USER_ID_KEY)) {
-            return userRepository.findUserByUserId(id) != null; 
+        switch (idType) {
+            case (ROOM_ID_KEY):
+                return roomRepository.findRoomByRoomId(id) != null;
+            case (USER_ID_KEY):
+                return userRepository.findUserByUserId(id) != null; 
+            default:
+                throw new IllegalArgumentException(String.format("The provided id type of %s is invalid.", idType));
         }
-        
-        // Return false for case where invalid or not present idType is given.
-        return true;
     }
 }
--- a/src/main/java/com/rocketden/main/util/Utility.java
+++ b/src/main/java/com/rocketden/main/util/Utility.java
@@ -45,12 +45,13 @@ public class Utility {
             }
             return id;
         } catch (IllegalArgumentException e) {
+            // Throw IllegalArgumentException for invalid id type.
             throw new IllegalArgumentException(e);
         }
     }
 
     // Determine if the id of idType already exists in the database.
-    private boolean idExists(String id, String idType) {
+    public boolean idExists(String id, String idType) {
         switch (idType) {
             case (ROOM_ID_KEY):
                 return roomRepository.findRoomByRoomId(id) != null;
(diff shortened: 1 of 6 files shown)


src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -2,23 +2,58 @@
...
-            int index = random.nextInt(numbers.length());
-            values[i] = numbers.charAt(index);
+        String id = null;
+        
+        /**
+         * Continue generating new id values until a valid one is provided.
+         * Note: This loop could technically run forever, but there are 9^6
+         * (531,441) possible combinations, so this works for now (10/20/2020).
+         */
+        while (id == null || idExists(id, idType)) {
```
helenli8181, line 38: Nit: I think checking if an ID exists could be simplified to a ternary statement inside the loop that sets `id` to `values` or `null`. The code still works the same way so it's not required but you save a line of code I guess?

src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -2,23 +2,58 @@
...
-        return new String(values);
+    // Determine if the id of idType already exists in the database.
+    private boolean idExists(String id, String idType) {
+        if (idType.equals(ROOM_ID_KEY)) {
+            return roomRepository.findRoomByRoomId(id) != null;
+        } else if (idType.equals(USER_ID_KEY)) {
+            return userRepository.findUserByUserId(id) != null; 
+        }
+        
+        // Return false for case where invalid or not present idType is given.
```
helenli8181, line 56: If you plan to have a case that returns false doesn't that mess with the while loop condition above?

helenli8181 (approved):
I have a couple of comments, but the code works great so I'll approve!

src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -2,23 +2,58 @@
...
+    }
 
-        return new String(values);
+    // Determine if the id of idType already exists in the database.
+    private boolean idExists(String id, String idType) {
+        if (idType.equals(ROOM_ID_KEY)) {
+            return roomRepository.findRoomByRoomId(id) != null;
+        } else if (idType.equals(USER_ID_KEY)) {
+            return userRepository.findUserByUserId(id) != null; 
+        }
```
alankbi, line 54: Suggestion: change this to a switch statement to help with readability

src/main/java/com/rocketden/main/util/Utility.java
alankbi: Maybe like throw an illegal argument exception if an invalid idType is given? Technically this statement should never be reached if we use it correctly, so maybe just a comment explaining that might be good as well, idk

src/test/java/com/rocketden/main/util/UtilityTests.java
```diff
@@ -30,7 +36,7 @@ public void generateValidRoomId() {
     @Test
     public void generateValidUserId() {
         // Verify user ids are generated correctly
-        String userId = utility.generateId(UserService.USER_ID_LENGTH);
+        String userId = utility.generateUniqueId(UserService.USER_ID_LENGTH, Utility.USER_ID_KEY);
 
         assertEquals(UserService.USER_ID_LENGTH, userId.length());
```
alankbi, line 42: Suggestion: Add a test that mocks the repository to return true when findRoomById is called the first time, and check that it generates a second id

alankbi (approved):
Same as Helen, I left a few comments + a suggested test, but overall I think it's good to be approved!

src/main/java/com/rocketden/main/util/Utility.java
zpChris: Hahaha I actually did use a switch statement first because I agreed with you, look at the warning I got:

Switch vs If Statement Readability

Turns out if I add three or more components to the switch statement it thinks that is good readability, so I added the default case for the `IllegalArgumentException`.

src/main/java/com/rocketden/main/util/Utility.java
zpChris: This is a good idea -- included!

src/main/java/com/rocketden/main/util/Utility.java
zpChris: If I understand correctly, this would just move the ID validation to within the loop (and have it continually `null` until it reaches a valid value) instead of outside of the loop; since I'd still need the check `id == null` in the while loop, I don't think this would save any lines? Since I think it's more intuitive to include all of the checking within the while condition, I'll leave it as-is for now; however, please let me know if you still think I should change this.

src/test/java/com/rocketden/main/util/UtilityTests.java
zpChris: I added a test for this, but that required that the `utility` I used is a `@Spy` with `@InjectMocks` so I could inject the `RoomRepository` and `UserRepository` mocks.

On several methods I did not include output for those mocks, and I believe that they default to `null`, which is the desired behavior on these cases. Please let me know if this is correct, and I can also add a comment explaining this (if it is true).

alankbi (approved):
Approved! Thanks for adding those tests

src/test/java/com/rocketden/main/util/UtilityTests.java
alankbi: Yep I think you're right about the return null so it should be good :)

PR #68, opened 2020-10-21
General saving and UI updates to the game page.

zpChris:
### List of Changes:

- Provide a warning to the user that their code or data may not be saved before exiting the game page (this occurs on refresh, close tab, close window, and close browser).
- Save the current room information in `location`.
- Prevent users from visiting the game page without being connected to a room; this is checked by determining whether a `room` was saved in `location`, so this is always passed through the `Lobby.tsx` page on redirect.
  - Add a `rebuildRoom(...)` function in `Lobby.tsx` that recreates the current room from all of the variables that are in `Lobby.tsx`. Note: Some of these variables are unused, such as `users`, but are only included for this purpose. We could also just pass in the `roomId` (and possibly the current user) and then perform a GET request to ensure all the latest values are retrieved -- I will let both of you decide whether that should be changed.

### Notes:

- Eventually, we can save the current Game information in `location`, but I don't yet know what this entails.
- The "loading" page should look a lot better, and I'm happy to make improvements now or leave that until later.

### Screencast and Images:

[Screencast of the General Game Page Saving and UI Updates](https://www.loom.com/share/3164433036464efb8ad5679685673891)

Image of User Not Able to Access the Game Page
Image of User Not Able to Access the Game Page

Image of Loading Game Page (split-second screen)
Image of Loading Game Page (split-second screen)

Image of Popup Preventing Accidentally-Closed Tab
Image of Popup Preventing Accidentally-Closed Tab

Image of Normal Game Page with Room ID Shown
Image of Normal Game Page with Room ID Shown


frontend/src/views/Game.tsx
```diff
@@ -1,45 +1,65 @@
...
   room: Room,
 }
 
 function GamePage() {
+  // Get history object to be able to move between different pages
+  const history = useHistory();
   const location = useLocation<LocationState>();
   const [room, setRoom] = useState<Room | null>(null);
   const [problems, setProblems] = useState<Problem[]>([]);
+  const [fullPageLoad, setFullPageLoad] = useState<boolean>(true);
```
helenli8181, line 28: Nit: naming this `fullPageLoading` may be more consistent since we have a `loading` state in Lobby.tsx

helenli8181 (approved):
This looks awesome! Thanks for the video explanation. I just have a nit but otherwise approved!

frontend/src/views/Game.tsx
zpChris: Done!

frontend/src/views/Game.tsx
```diff
@@ -1,45 +1,65 @@
...
+  const [fullPageLoading, setFullPageLoading] = useState<boolean>(true);
   const [error, setError] = useState<string>('');
 
+  /**
+   * Display beforeUnload message to inform the user that they may lose
+   * their code / data if they leave the page.
+   * Some browsers will display this message, others will display a fixed
+   * message; see https://github.com/jacobbuck/react-beforeunload.
+   */
+  useBeforeunload(() => 'Leaving this page may cause you to lose your current code and data.');
```
alankbi, line 37: The fact that unload isn't capitalized is bothering me way more than it should haha

frontend/src/views/Game.tsx
```diff
@@ -1,45 +1,65 @@
...
+          setError('Problem cannot be found.');
+        } else {
+          setProblems(res as Problem[]);
+          setError('');
+        }
+      }).catch((err) => {
+        setError((err as ErrorResponse).message);
+        setProblems([]);
+      });
+      setFullPageLoading(false);
```
alankbi, line 56: nit: I would move this to just under `setRoom(...)` so it's more clear it's part of the if statement and not the then/catch blocks

alankbi: or maybe it actually is supposed to be in the getProblems callback?

frontend/src/views/Lobby.tsx
```diff
@@ -51,13 +52,31 @@ function LobbyPage() {
...
+   * the useEffect function.
+   */
+  const rebuildRoom = useCallback((): Room => ({
+    roomId: currentRoomId,
+    host: host!,
+    users: users!,
+    activeUsers: activeUsers!,
+    inactiveUsers: inactiveUsers!,
+    active,
+    difficulty: difficulty!,
```
alankbi, line 76: Thoughts on just sending the `roomId` and `currentUser`? I'm thinking if we did something similar to the lobby page where we just pass those two and have a GET request or something, we might be able to simplify the stuff held in location state, but I'm not sure - we can talk about it during standup

alankbi: On a related note, if we decide on reducing this to just 1-2 variables, what are your thoughts on removing this callback and just adding the logic to the useEffect block itself? Since that's the only place it's used and it would end up just being something like `const params = { roomId: currentRoomId, user: currentUser }`

alankbi (commented):
Looks great! Before approving I just had some nits and one question about the location stuff, which I think maybe I can ask about during standup tomorrow

frontend/src/views/Game.tsx
zpChris: Here's what we do -- make a new npm package that just uses this existing package but then renames it, publish the package, then use _that_ in our codebase. Or we can make a Utility function that just invokes this method but is called `useBeforeUnload`, then we use _that_ function.

For now though, I'll leave it as-is. 🙂

frontend/src/views/Lobby.tsx
zpChris: I replaced the code to just have a `roomId` and `currentUser`, and I removed this callback as well as you suggested -- these are both good ideas, thanks! I haven't added a `GET` request to get the whole room, since I think this could be changed if we get a Game object, for example, on that page, and I don't think it's necessary to add that to this PR.

I also reflected the `roomId` and `currentUser` on the frontend on the Game page.

frontend/src/views/Game.tsx
zpChris: I was hoping all of the page information could be populated before the actual page is shown; eventually, the loading symbol could be a prettier transition, but for now, I thought text would be a fine substitute.

I don't think it should be in the getProblems callback, since if the problems are not fetched properly, that should be displayed to the user rather than a static loading page.

alankbi: Sounds good - FYI I think right now the problem actually does load in a little bit later than the rest of the page (cause of async stuff), but personally I actually kind of like that so no changes here on my end unless that's something you care for 👍

frontend/src/views/Game.tsx
```diff
@@ -1,45 +1,68 @@
...
+    } else {
+      history.push('/game/join', {
```
alankbi, line 61: nit: Change to replace

alankbi (approved):
Approved!

frontend/src/views/Game.tsx
zpChris: Done!

PR #69, opened 2020-10-22
REST endpoints for creating problem test cases and getting a single problem

alankbi:
### List of Changes:

* Add GET endpoint for a single problem
* Add POST endpoint to create a test case for a problem
* Set up DTOs, mapper classes, entity methods, bidirectional mapping, etc. 
* Create unit and integration tests as usual
* Minor refactorings and updates to old code

### Notes:

* Lots of lines changed sorry - as usual it's mainly the tests, but feel free to leave comments if you think any of the code or proposed design needs significant changes
* Right now the database of the problem id is used for identification and exposed to the frontend - thoughts on this? It makes the logic much easier but exposes the id to the client, so a question of how important that is for us. In either case, I added a card for it in the backlog (this PR is big enough already so it'd def come at a later time)
* Hidden options - there seem to be multiple options for test case visibility: fully visible, only visible if submitted code fails on that test case (either both input and output or just input), etc. Added a card for this to look into for the future
  * There's also a need to somehow filter test cases to only send non-hidden test cases to the game page, BUT when an admin is creating a problem, they should be able to fetch everything
* Explanation field - thoughts? Stored in the test case or in the problem description?
* Test case ordering - we might want to allow admins to change the order of test cases (such as which test case is the first example), so I added a card for this as well

### Screencast and Images:

Get single problem:
Get problem

Create test case:
Create test case

Test case database table:
Test case database table

Problem database table: 
Problem database table


src/main/java/com/rocketden/main/model/problem/ProblemTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+
+@Entity
+@Getter
+@Setter
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+public class ProblemTestCase {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Integer id;
```
zpChris, line 23: Is there a reason you didn't add the `@EqualsAndHashCode.Include` annotation here?

src/test/java/com/rocketden/main/api/ProblemTests.java
```diff
@@ -39,16 +42,37 @@
...
     private static final String POST_PROBLEM_CREATE = "/api/v1/problems";
+    private static final String POST_TEST_CASE_CREATE = "/api/v1/problems/%s/test-case";
 
     private static final String NAME = "Sort an Array";
     private static final String DESCRIPTION = "Sort an array from lowest to highest value.";
     private static final String NAME_2 = "Find Maximum";
     private static final String DESCRIPTION_2 = "Find the maximum value in an array.";
 
+    private static final String INPUT = "[1, 2, 8]";
+    private static final String OUTPUT = "8";
```
zpChris, line 56: Were you intending these `INPUT` / `OUTPUT` values to match the **first** `NAME` / `DESCRIPTION`?  In that case, you'd change the `OUTPUT` to `[1, 2, 8]`, otherwise no problem.

src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
```diff
@@ -1,4 +1,12 @@
...
 
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
 public class CreateTestCaseRequest {
+    private String input;
+    private String output;
+    private boolean hidden = false;
```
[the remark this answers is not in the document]
zpChris: Also, thoughts on allowing the user to set the value of `hidden`?

src/test/java/com/rocketden/main/api/ProblemTests.java
```diff
@@ -39,16 +42,37 @@
...
+                .andReturn();
+
+        String jsonResponse = result.getResponse().getContentAsString();
+        ApiErrorResponse actual = UtilityTestMethods.toObject(jsonResponse, ApiErrorResponse.class);
+
+        assertEquals(ERROR.getResponse(), actual);
+    }
+
+    @Test
+    public void createAndGetProblemSuccess() throws Exception {
```
zpChris, line 75: If you allow the user to create a test case with a provided `hidden` value, could you include a test for that as well?

src/test/java/com/rocketden/main/api/ProblemTests.java
```diff
@@ -135,4 +172,145 @@ public void getProblemsEmptyList() throws Exception {
...
+        assertEquals(OUTPUT_2, case2.getOutput());
+        assertFalse(case2.isHidden());
+    }
+
+    /**
+     * Helper method that sends a POST request to create a new problem
+     * @return the created problem
+     * @throws Exception if anything wrong occurs
+     */
+    private ProblemDto createSingleProblem() throws Exception {
```
zpChris, line 297: nit, personal preference: Add this method to the top of the file; since you call it above, I feel that makes more sense, so the user can see this helper method more quickly.

src/main/java/com/rocketden/main/model/problem/Problem.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Entity // This tells Hibernate to make a table out of this class
+@Getter
+@Setter
+public class Problem {
```
zpChris, line 20: We should add a difficulty level, and problem type, to these problems. This could be done now or in the future, but it will be necessary if people want to populate the game with particular problem sets.

src/test/java/com/rocketden/main/mapper/ProblemMapperTests.java
```diff
@@ -0,0 +1,68 @@
...
+
+@SpringBootTest
+public class ProblemMapperTests {
+
+    private static final int ID = 10;
+    private static final String NAME = "Sort a List";
+    private static final String DESCRIPTION = "Sort the given list in O(n log n) time.";
+
+    private static final String INPUT = "[1, 2, 8]";
+    private static final String OUTPUT = "8";
```
zpChris, line 24: nit: Change `OUTPUT` to match `DESCRIPTION`, same as before?

src/test/java/com/rocketden/main/mapper/ProblemMapperTests.java
```diff
@@ -0,0 +1,68 @@
...
+        testCase.setOutput(OUTPUT);
+        testCase.setHidden(true);
+        expected.addTestCase(testCase);
+
+        ProblemDto actual = ProblemMapper.toDto(expected);
+
+        assertEquals(expected.getId(), actual.getId());
+        assertEquals(expected.getName(), actual.getName());
+        assertEquals(expected.getDescription(), actual.getDescription());
+
```
zpChris, line 44: nit: Remove extra line break.

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -32,9 +37,45 @@
...
+        expected.setId(ID);
+        expected.setName(NAME);
+        expected.setDescription(DESCRIPTION);
+
+        ProblemTestCase testCase = new ProblemTestCase();
+        testCase.setInput(INPUT);
+        testCase.setOutput(OUTPUT);
+        expected.addTestCase(testCase);
+
+        Mockito.doReturn(Optional.of(expected)).when(repository).findById(ID);
```
zpChris, line 59: What is the use of `Optional` here? It might be useful to add a comment for this?

zpChris (changes requested):
Requesting changes just because I left several nits, but it looks great overall! This will be approved after this review IMO.

src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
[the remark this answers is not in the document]
alankbi: Yeah sounds good, I think I'll leave this for a later time since making enums can be a hassle and might cause the PR to be more bloated (there should be a card in the backlog for this though). 

Second point: yep I agree (it actually should already be a feature, the false is just the default value :))

src/main/java/com/rocketden/main/model/problem/ProblemTestCase.java
alankbi: My main thought process being that it's mainly used for the database equality stuff whereas we might want to use a business ID of some sort in our application (similar to in Room how we use userId and nickname)

alankbi: lmk what you think tho, I can add it in for now if you think it's preferable

src/test/java/com/rocketden/main/api/ProblemTests.java
alankbi: Haha I wasn't but that's a good catch - updated in all the test files

src/test/java/com/rocketden/main/api/ProblemTests.java
alankbi: Sounds good - there should be a test below and in the service tests that cover that case

src/test/java/com/rocketden/main/api/ProblemTests.java
alankbi: Good point - I also updated the RoomTests file to move those helper methods to the top as well - lmk if you think I should move it back tho

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
alankbi: `findById` returns an Optional as a wrapper around the actual object for whatever reason rather than just returning null when not found. Kinda weird I agree but yeah I added a comment to clarify this

src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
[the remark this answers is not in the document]
zpChris: Ah gotcha, that makes sense -- thanks!

src/main/java/com/rocketden/main/model/problem/ProblemTestCase.java
zpChris: Well I think I agree if we add a problemId, but since we're using the ID for comparisons at least right now, I think we might want to compare by that?

This is just a bit though, and if you think we'll change it in the future I don't have too much of a problem with this.

src/test/java/com/rocketden/main/api/ProblemTests.java
zpChris: Ah great, thanks!

src/test/java/com/rocketden/main/api/ProblemTests.java
zpChris: That's a good change, thanks for doing that in both files 👍

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
zpChris: Huh interesting -- any idea why this is different than the other "findByX" methods?

src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
```diff
@@ -1,4 +1,12 @@
 package com.rocketden.main.dto.problem;
 
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
 public class CreateTestCaseRequest {
+    private String input;
+    private String output;
```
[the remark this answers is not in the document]
alankbi: Not sure - the difficulty with custom objects like trees, maps, arrays, etc. is that those things don't really convert into database tables/rows very well, and also idk how to allow users to create Java objects on the frontend if they want to write custom test cases (like LeetCode does). However I'm not too familiar so I def think it's possible depending on what we go with for the compiling process. 

As for images, I think it would be easiest if those just go in the problem description section (which maybe could be formatted as markdown?) since the majority of test cases will likely be hidden or auto generated or something, but again this totally depends on how we go about it I think

Also, remember [this](https://github.com/DrDab/LaunchTestRun) haha? Might be a useful reference (although this uses scanners instead of method args)

src/main/java/com/rocketden/main/model/problem/Problem.java
alankbi: Yep, adding the difficulty in #71 - it doesn't have the problem type yet (I think that could be post-MVP since we prob won't have enough problems to support that yet), but I'm open to adding that too if we think it'll be useful now

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
alankbi: Not sure tbh haha - [this](https://stackoverflow.com/questions/60638660/why-findbyid-returns-optional) goes into some of the use cases tho

src/main/java/com/rocketden/main/model/problem/ProblemTestCase.java
alankbi: Sounds good - added a commit for this

src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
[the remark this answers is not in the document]
zpChris: That first point is fair -- we can keep strings for input and output, at least in this PR.

Yeah I had meant to bring up LaunchTestRun -- could we reach out to him and ask how that works and how to set ours up potentially?

src/main/java/com/rocketden/main/model/problem/Problem.java
zpChris: That makes sense, I'm open to adding that during the MVP but no need to add it in this PR, so all good either way.

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
zpChris: Oh interesting, there's a similar concept in OCaml. Ok yeah no problem, this looks good.

src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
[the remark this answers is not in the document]
alankbi: I think that could be worth looking into

zpChris (approved):
LGTM! Thanks for making those changes 👍

alankbi:
[REST endpoints for problem test cases](https://trello.com/c/wefQ9gbs/167-rest-endpoints-for-problem-test-cases)

helenli8181 (approved):
OML sorry I keep forgetting to review this one! This looks great to me :)

PR #73, opened 2020-10-27
Add console panel to game page

alankbi:
### List of Changes:

* Add a console panel to the game page
* Display the input of the first test case (if one exists)
* Create unimplemented code structure for running code against custom input
  

### Notes:

* The current output field might eventually be converted into either a diff-like field or a separate field for each of expected output and actual output
* Everything is based on a hypothetical structure for how players will test and submit their code, so feel free to lmk if we want to change any of it

### Screencast and Images:

[Screencast of game page with new console panel](https://www.loom.com/share/fa4e9a6d2f2a42ee89be209a219407a5)


alankbi:
[Frontend display test cases for problems](https://trello.com/c/OZjY8lSV/176-frontend-display-test-cases-for-problems)

frontend/src/components/core/Text.tsx
```diff
@@ -4,6 +4,10 @@ export const Text = styled.p`
   font-size: ${({ theme }) => theme.fontSize.default};
 `;
 
+export const BoldText = styled(Text)`
+  font-weight: bold;
+`;
+
```
zpChris, line 10: Thoughts on replacing `BoldText` with `Text` that has a `bold` attribute parameter?

frontend/src/views/Game.tsx
```diff
@@ -20,57 +20,80 @@ type LocationState = {
...
-
   // Creates Event when splitter bar is dragged
   const onSecondaryPanelSizeChange = () => {
     const event = new Event('secondaryPanelSizeChange');
     window.dispatchEvent(event);
   };
 
+  // Callback when user runs code against custom test case
+  const runSolution = (input: string) => {
+    // Needs to be implemented
```
zpChris, line 51: nit: Maybe add "TODO" if you're going to keep this in? Or just remove the comment?

frontend/src/components/game/Console.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <SmallButton onClick={() => onRun(input)}>Run Code</SmallButton>
+      </FixedContent>
+      <div>
+        <BoldText>Input</BoldText>
+        <ConsoleTextArea value={input} onChange={(e) => setInput(e.target.value)} />
+      </div>
+      <div>
+        <BoldText>Output</BoldText>
+        <ConsoleTextArea value={output} readOnly />
+      </div>
```
zpChris, line 51: Is this UI issue fixable? [Screencast here](https://www.loom.com/share/c1de1a7b4fa849289b480dc8ba4d8867), image below.

UI Issue with Draggable Input

If not, maybe you could temporarily disable the draggability of the text boxes, or just leave it as-is -- not a huge issue.

zpChris (changes requested):
Requesting changes just because I have a few nits plus a potential UI issue you may want to fix, depending on its difficulty. Looks good! 👍

frontend/src/components/core/Text.tsx
alankbi: Good idea - added an optional bold property to the Text component to allow for this

frontend/src/components/game/Console.tsx
alankbi: Good catch - I made the console scrollable and added constraints to the text box sizes, so it should no longer allow for this situation

frontend/src/components/game/Console.tsx
```diff
@@ -1,12 +1,13 @@
...
 import styled from 'styled-components';
 import { SubmissionResult, TestCase } from '../../api/Problem';
-import { BoldText } from '../core/Text';
+import { Text } from '../core/Text';
 import { ConsoleTextArea } from '../core/Input';
 import { SmallButton } from '../core/Button';
 
 const Content = styled.div`
   height: 100%;
+  overflow-y:auto;
```
zpChris, line 10: nit: Add space after colon.

zpChris (approved):
LGTM -- only left one nit, thanks for addressing my previous ones!

helenli8181 (changes requested):
This looks awesome! I just have one suggestion -- it's possible to drag the horizontal dividing bar all the way down so that the test cases box disappears. Is that intentional, or would it be better to have a minimum height on that box? I think something similar is implemented for the main vertical dividing bar.

alankbi:
Yeah I meant for that to be intentional so that users can hide it when they're not using it (similar to the console panel on LeetCode) but I'm open to setting a larger minimum height if y'all think that would be better - just lmk

zpChris:
> Yeah I meant for that to be intentional so that users can hide it when they're not using it (similar to the console panel on LeetCode) but I'm open to setting a larger minimum height if y'all think that would be better - just lmk

I think this is good, I wouldn't change that personally 👍

alankbi:
After we talked about it in the chat, I think we might push the min console height feature till later (at which point we can also have a hide button which minimizes it completely)

helenli8181 (approved):
Approved!

PR #74, opened 2020-11-02
Create client-facing problemId field and random problem endpoint

alankbi:
### List of Changes:

* Add auto-generated `problemId` field for problems
* Replace any client-facing usages of `id` with `problemId`
* Create GET endpoint for random problem
* Update game page to get random problem by difficulty
* Add necessary tests

### Notes:

* On the game page, the random problem changes on every refresh, which I can see us wanting to prevent in the future - or in this PR as well, just lmk


alankbi:
[Create a Problem ID separate from ID](https://trello.com/c/rOOz0tDi/188-create-a-problem-id-separate-from-id)

src/test/java/com/rocketden/main/mapper/ProblemMapperTests.java
```diff
@@ -40,7 +38,7 @@ public void entityToDto() {
 
         ProblemDto actual = ProblemMapper.toDto(expected);
 
-        assertEquals(expected.getId(), actual.getId());
+        assertEquals(expected.getProblemId(), actual.getProblemId());
```
zpChris, line 41: I will note that since you're using UUID generation, you don't know what form this ID takes. You corrected me in this PR that you should be able to use UUID with testing by just fetching the problemId from the object, but there still is this small note -- is this an issue, or no?

frontend/src/api/Problem.ts
```diff
@@ -2,15 +2,17 @@ import axios from 'axios';
...
   name: string,
   description: string;
 }
 
 const basePath = '/api/v1/problems';
 const routes = {
   getProblems: `${basePath}/`,
-  postProblem: `${basePath}/`,
+  createProblem: `${basePath}/`,
+  getSingleProblem: (problemId: string) => `${basePath}/${problemId}`,
```
zpChris, line 14: When would this be used? Could you create a `getSingleRandomProblem` method that **doesn't** require a `problemId`, so that can be used instead of `getProblems`? How could you use the `getSingleRandomProblem` method to ensure that the same problem isn't returned twice in a row, if some users restart the game?

zpChris (commented):
Looks good! Left two small comments that may lead to changes, so I won't approve or request changes right now, but respond to those and ping me and I'll review again 👍

src/test/java/com/rocketden/main/mapper/ProblemMapperTests.java
alankbi: I think it should hopefully be a non-issue, but yeah definitely something to consider before doing that other Trello task I think

frontend/src/api/Problem.ts
alankbi: I'm not sure about the part with ensuring that a problem isn't received twice in a row. Is this like on each page refresh or do you mean like if people play multiple games one after the other? 

Both situations seem tricky - might involve some localStorage stuff

alankbi: As for `getSingleProblem`, right now it's not used anywhere but I'm imagining once we get to the point where we start creating problems the endpoint might be used somewhat often for like an admin panel of some sorts (or like if users can create/edit their own problems in a kahoot-style game, although that's way in the future)

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -61,9 +67,30 @@ public ProblemDto getProblem(int id) {
...
     }
 
-    public ProblemTestCaseDto createTestCase(int problemId, CreateTestCaseRequest request) {
-        Problem problem = repository.findById(problemId)
-                .orElseThrow(() -> new ApiException(ProblemError.NOT_FOUND));
+    public ProblemDto getRandomProblem(ProblemSettingsDto request) {
+        ProblemDifficulty difficulty = request.getDifficulty();
+        List<Problem> problems;
+
+        if (difficulty == null || difficulty == ProblemDifficulty.RANDOM) {
```
helenli8181, line 74: In what situations could a difficulty not be specified? If it's not supposed to be allowed, should it return an error instead of assuming a random difficulty?

helenli8181 (approved):
This looks great to me! I just had a question but otherwise approved.

frontend/src/api/Problem.ts
zpChris: I think storing in `localStorage` or perhaps in a group of user's accounts would be good, but that's definitely for the future, not necessary to change in this PR.

Thoughts on the `getSingleRandomProblem` method? I'll approve anyway since it's been too long since I've reviewed, but do you think you could implement that?

src/main/java/com/rocketden/main/dto/problem/ProblemSettingsDto.java
```diff
@@ -0,0 +1,12 @@
+package com.rocketden.main.dto.problem;
+
+import com.rocketden.main.model.problem.ProblemDifficulty;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class ProblemSettingsDto {
+
```
zpChris, line 10: nit: Remove this extra line.

src/test/java/com/rocketden/main/api/ProblemTests.java
```diff
@@ -338,4 +342,40 @@ public void createProblemWithTestCasesSuccess() throws Exception {
...
+
+        assertEquals(problem.getName(), actual.getName());
+        assertEquals(problem.getDescription(), actual.getDescription());
+        assertEquals(ProblemDifficulty.EASY, actual.getDifficulty());
+        assertEquals(problem.getProblemId(), actual.getProblemId());
+        assertEquals(problem.getTestCases(), actual.getTestCases());
+    }
+
+    @Test
+    public void getRandomProblemNotFound() throws Exception {
```
zpChris, line 366: Can you add a third test where you add no problems to the database, and then test that you still get `ProblemError.NOT_FOUND`?

zpChris (approved):
LGTM! Works great locally, left a few nits but otherwise this looks great.

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: My assumption behind it was that if no difficulty is specified, that means that the user doesn't care and is open to getting anything, hence treating it the same as random - lmk if you think I should change it tho and I can do that in a separate PR

frontend/src/views/Lobby.tsx
```diff
@@ -227,9 +227,10 @@ function LobbyPage() {
       history.push('/game', {
         roomId: currentRoomId,
         currentUser,
+        difficulty,
```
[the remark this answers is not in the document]
alankbi: Yeah that makes sense - I'm not sure if that's implemented right now but I can see in the future if on the backend the game object is created and automatically assigned a random problem from there, which is then sent to the frontend on the game page (that would also help to prevent problems from changing on refresh). But yeah lmk if you think that I should do that in this or the next PR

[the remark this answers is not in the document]
zpChris: You can do that in the next PR, I see no need now since we haven't implemented the game fetch logic yet -- just wanted to note this down. 👍

src/test/java/com/rocketden/main/api/ProblemTests.java
alankbi: This should be covered in the `getRandomProblemNotFound` test in `ProblemServiceTests` - lmk if you think I should add one here as well tho

src/main/java/com/rocketden/main/service/ProblemService.java
zpChris: Curious on @helenli8181's thoughts, but I think my preference would skew towards making the user specify this attribute? But like this is the smallest of the teeny-tiny nits, so either way I don't think you have to worry too much.

src/test/java/com/rocketden/main/api/ProblemTests.java
zpChris: Yeah it's testing the same thing, so there's no real need (which in my mind again raises the question of why we have these two separate sets of tests). I suppose just in the future, if we remove one of the sets of testing, to ensure that this behavior is still tested.

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: Yeah sure that works - just pushed a commit to update this

alankbi: I'll merge this for now (eager to move past it haha) but if @helenli8181 you feel differently lmk and I can update it in the next PR

PR #78, opened 2020-11-25
Change remove users to a DELETE endpoint

alankbi:
### List of Changes:

* Change remove users to a DELETE endpoint
* Update frontend and tests accordingly

### Notes:

* This is to stay consistent with REST guidelines, although to be fair this does introduce including a request body in a DELETE request, which is slightly unusual
* In order to get a DELETE request with a request body using axios, the syntax had to be slightly changed

### DIFF
--- a/frontend/src/api/Room.ts
+++ b/frontend/src/api/Room.ts
@@ -82,10 +82,11 @@ export const changeRoomHost = (roomId: string, roomParams: ChangeHostParams):
   });
 
 export const removeUser = (roomId: string, roomParams: RemoveUserParams):
-  Promise<Room> => axios.delete<Room>(routes.removeUser(roomId), {
-    headers: { 'Content-Type': 'application/json' },
-    data: roomParams,
-  })
+  Promise<Room> => axios({
+  url: routes.removeUser(roomId),
+  method: 'DELETE',
+  data: roomParams,
+})
   .then((res) => res.data)
   .catch((err) => {
     throw axiosErrorHandler(err);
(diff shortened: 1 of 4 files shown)


alankbi:
[Problem with frontend params when using axios DELETE](https://trello.com/c/PAW8NkcS/191-problem-with-frontend-params-when-using-axios-delete)

frontend/src/api/Room.ts
```diff
@@ -82,7 +82,11 @@ export const changeRoomHost = (roomId: string, roomParams: ChangeHostParams):
   });
 
 export const removeUser = (roomId: string, roomParams: RemoveUserParams):
-  Promise<Room> => axios.put<Room>(routes.removeUser(roomId), roomParams)
+  Promise<Room> => axios({
+  url: routes.removeUser(roomId),
+  method: 'DELETE',
+  data: roomParams,
+})
```
zpChris, line 89: Is there a reason you didn't use `axios.delete`. [[1]](https://stackoverflow.com/questions/51069552/axios-delete-request-with-body-and-headers) If we are switching it to this format, should we keep it consistent across the rest of the `axios` functions?

zpChris (approved):
LGTM! Left one nit on how `axios` is used, and possibly changing the formatting of this function or others as well.

frontend/src/api/Room.ts
helenli8181: This probably has something to do with the issue I couldn't figure out where you'll keep getting Bad Request when you try to send a body with the delete request without enclosing it in "data".

helenli8181 (approved):
This looks awesome! Thanks for figuring out the fix to this issue :)

frontend/src/api/Room.ts
alankbi: Yeah I think I ended up coming across the same error as Helen, even when using `config.data` with `axios.delete`for the request body as a ton of posts online suggested, so this ended up being the first solution for me that worked. Personally I think this is less clean so I'm ok with leaving the other ones as they are, unless you have any preference for consistency sake or like seeing if there are other ways to get `axios.delete` to work

alankbi: I'll merge for now but am free to update this anytime

PR #79, opened 2020-12-11
Fix lobby page bug with back button and reopening tab

### LINKED ISSUE #51: Socket already connected error shows on lobby page when using forward/back buttons
Once connected to a room, if you click the back button and then the forward button again, the "socket is already connected" shows on the screen. This should not show up as it conveys that something went wrong when I think our current code actually handles the case just fine. 

Socket already connected error

### LINKED ISSUE #60: Improper Refresh on Command + Shift + T
After the changes on PR #53, we noticed that opening up a closed tab through `command + shift + t` does not properly refresh that page, and makes that user seem like they are connected when in actuality they are not. This should be fixed.

alankbi:
### List of Changes:

* Fixes #51 where a "socket already connected" error is shown on navigation with back/forward buttons
* Fixes #60 where lobby page shows wrong info on page refresh or reopening (CTRL+SHIFT+T)

### Notes:

* In order to avoid introducing another bug (where users receive socket messages with the wrong host due to the host change not flushing to the DB in time), I added a 0.5 second delay to the socket connection event listener

### DIFF
--- a/frontend/src/api/Socket.ts
+++ b/frontend/src/api/Socket.ts
@@ -53,7 +53,7 @@ export const connect = (roomId: string, userId: string):
         reject(errorHandler('The socket failed to connect.'));
       });
     } else {
-      reject(errorHandler('The socket is already connected.'));
+      resolve();
     }
   });
 
--- a/frontend/src/views/Lobby.tsx
+++ b/frontend/src/views/Lobby.tsx
@@ -172,7 +172,6 @@ function LobbyPage() {
      */
     const subscribeCallback = (result: Message) => {
       setStateFromRoom(JSON.parse(result.body));
-      console.log(JSON.parse(result.body).host);
     };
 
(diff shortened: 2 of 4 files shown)


alankbi:
[Lack of refresh on "back" button or command + shift + t](https://trello.com/c/vAiob8I2/158-lack-of-refresh-on-back-button-or-command-shift-t)

zpChris (approved):
LGTM!

I wanted to note that there is still an issue when a user is kicked that they are not booted off the page; if a disconnected / inactive user is kicked then that user does **ctrl + shift + t**, it does not show them as a user on the list of users, but it does still show their nickname in "Your nickname is ___". No changes necessary here though -- I think this may be fixed once we finish kick functionality.

helenli8181 (approved):
This looks great, approved!

PR #80, opened 2020-12-16
Connect game page to backend Game objects

alankbi:
### List of Changes:

* Implement stub methods in GameManagementService
* Flesh out and refactor a few details with the existing Game code
* Add GET endpoint for game objects and basic GameMapper logic
* Implement axios call on frontend game page

### Notes:

* I left a `TODO` in the Game class cause I wasn't sure what the best design choice was with that moving forward
* A lot of fields are currently blank in the GameDto as idk if we've fully decided on how we're going to implement that yet

### Screencast and Images:

Console showing game object


alankbi:
[Connect to game object on game start](https://trello.com/c/njbvlf6F/193-connect-to-game-object-on-game-start)

src/main/java/com/rocketden/main/game_object/Game.java
```diff
@@ -13,9 +14,10 @@
 
     private Room room;
 
-    private Map<String, Player> players;
+    // TODO discussion: is the key userId or nickname or none?
+    private Map<String, Player> players = new HashMap<>();
```
zpChris, line 18: Generally, I think players are going to be identified by `userId`, both now and in the future. Although there is the concern of exposing these values to the users, which leads to security issues, I think that if we ever allow duplicate usernames in the same game (or just across the platform, may be a concern), or accounts with specific settings, `userId` will be the best.

Not sure whether this needs to be a Map or just a List -- though I think eventually this will be useful when attempting to find a user quickly, so I'd say a Map with `userId` key and `Player` value is the best design for now. I would also be fine with just a List though.

src/main/java/com/rocketden/main/game_object/Player.java
```diff
@@ -18,12 +19,12 @@
     private PlayerCode playerCode;
 
     // Submissions in order of first-to-last submitted.
-    private List<Submission> submissions;
+    private List<Submission> submissions = new ArrayList<>();
```
zpChris, line 22: This is just setting the base value, correct? (And the other same changes as well?)

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -38,21 +46,54 @@ protected GameManagementService(RoomRepository repository, SocketService socketS
...
+        }
+
+        return game;
     }
 
     protected void removeGame(String roomId) {
         currentGameMap.remove(roomId);
     }
 
+    public GameDto getGameDto(String roomId) {
```
zpChris, line 62: nit: Maybe change to `getGameDtoFromRoomId`, or change the above analogous method to just `getGame`, to keep some uniformity across method names?

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -38,21 +46,54 @@ protected GameManagementService(RoomRepository repository, SocketService socketS
...
+        // If user making request is not the host, throw an exception.
+        if (!request.getInitiator().getNickname().equals(room.getHost().getNickname())) {
+            throw new ApiException(RoomError.INVALID_PERMISSIONS);
+        }
+
+        room.setActive(true);
+        repository.save(room);
+
+        // Initialize game state
+        createAddGameFromRoom(room);
```
zpChris, line 84: Eventually, there will be other "conversion" steps like making the timer start and any necessary steps for taking care of room settings; however, no other changes now. This looks great!

src/main/java/com/rocketden/main/dto/game/GameMapper.java
```diff
@@ -0,0 +1,51 @@
...
+        if (room == null) {
+            return null;
+        }
+
+        Game game = new Game();
+        game.setRoom(room);
+
+        for (User user : room.getUsers()) {
+            Player player = playerFromUser(user);
+            game.getPlayers().put(user.getNickname(), player);
```
zpChris, line 35: nit: Save `game.getPlayers()` as a variable outside of the for loop, like in a value called `players` or `playerList`, then use `put` on this variable. Not sure if this is actually more efficient, but then `getPlayers()` is only called once.

src/test/java/com/rocketden/main/mapper/GameMapperTests.java
```diff
@@ -0,0 +1,67 @@
...
+        room.setRoomId(ROOM_ID);
+        User user = new User();
+        user.setNickname(NICKNAME);
+        room.addUser(user);
+        
+        Game game = GameMapper.fromRoom(room);
+
+        assertEquals(room, game.getRoom());
+        assertNotNull(game.getPlayers().get(NICKNAME));
+        assertEquals(user, game.getPlayers().get(NICKNAME).getUser());
```
zpChris, line 37: This works because there is only one user in the room, correct?

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
```diff
@@ -37,13 +45,40 @@
...
+    private GameManagementService gameService;
     
     // Predefine user and room attributes.
     private static final String NICKNAME = "rocket";
     private static final String NICKNAME_2 = "rocketrocket";
     private static final String ROOM_ID = "012345";
 
+    @Test
+	public void addGetAndRemoveGame() {
+    	// Initially, room doesn't exist
```
zpChris, line 57: nit: Move comments forward by tab.

zpChris (changes requested):
Only requesting changes so you can address the `TODO` you created; otherwise, I left a few nits, but this looks great!

frontend/src/views/Game.tsx
```diff
@@ -60,6 +62,14 @@ function GamePage() {
         setFullPageLoading(false);
         setError(err.message);
       });
+
+      // Get game object with room details
+      getGame(location.state.roomId)
+        .then((res) => {
+          setGame(res);
+          console.log(res);
```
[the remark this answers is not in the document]
alankbi: Yep :)

frontend/src/views/Game.tsx
```diff
@@ -60,6 +62,14 @@ function GamePage() {
         setFullPageLoading(false);
         setError(err.message);
       });
+
+      // Get game object with room details
+      getGame(location.state.roomId)
+        .then((res) => {
+          setGame(res);
```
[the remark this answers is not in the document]
alankbi: Could you clarify this? Although if I'm understanding correctly kind of like a `setStateFromGame` function like we have on the lobby page?

src/main/java/com/rocketden/main/game_object/Player.java
alankbi: Yeah I believe so, basically to make it default to an empty list rather than null

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
alankbi: Just reindented the file although idk if that actually fixed it or not - lmk if not

src/main/java/com/rocketden/main/game_object/Game.java
alankbi: Sounds good, I updated the code to use userId (previously was using nickname)

src/test/java/com/rocketden/main/mapper/GameMapperTests.java
alankbi: This was actually outdated on my part (using nickname instead of userId) - after the update it should work by indexing the map using the userId, so I think it should work even with multiple users

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
zpChris: I meant that right now, the game state is only updated when `location` or `history` changes; however, in the future, we'll need this to update when we take new socket messages or update the code or whatnot. Regardless, there are no changes now, I just wanted to note this for the future.

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
zpChris: Yes, this fixed it -- I'll resolve.

src/test/java/com/rocketden/main/mapper/GameMapperTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.rocketden.main.dto.game.GameDto;
+import com.rocketden.main.dto.game.GameMapper;
+import com.rocketden.main.dto.room.RoomMapper;
+import com.rocketden.main.game_object.Game;
+import com.rocketden.main.game_object.Player;
+import com.rocketden.main.model.Room;
+import com.rocketden.main.model.User;
+import org.junit.jupiter.api.Test;
+import org.modelmapper.ModelMapper;
```
zpChris, line 11: nit: Remove unused import.

zpChris (approved):
LGTM! Just left a nit to remove an unused import.

helenli8181 (approved):
Wow these are a lot of changes! Looks good to me!

PR #81, opened 2020-12-28
Add an explanation field to problem test cases.

zpChris:
### List of Changes:

- Add an `explanation` field to the `ProblemTestCase` object, which can be used to provide an explanation for more complicated test cases for players.
- Make relevant changes to add the `explanation` field to codebase tests.

### Notes:

- This PR does not have any changes to the current MVP, as the test case explanation is not shown in the UI next to the Input and Output; however, this can be used later to explain more complicated test cases to the players.
- I made a couple extraneous changes in the files containing problem test case components, including adding clarifying comments and removing unused imports.
- The `explanation` field is default set to `null` (which means no explanation is included). Eventually, this field (and the `description` field of the `Problem` object) could be a "markdown" object and support that style -- but that is far past the MVP.

### Screencast and Images:

There are no changes to the UI, so no screencasts or images are attached here.

### DIFF
--- a/src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
+++ b/src/main/java/com/rocketden/main/dto/problem/CreateTestCaseRequest.java
@@ -9,4 +9,5 @@ public class CreateTestCaseRequest {
     private String input;
     private String output;
     private boolean hidden = false;
+    private String explanation;
 }
--- a/src/main/java/com/rocketden/main/dto/problem/ProblemTestCaseDto.java
+++ b/src/main/java/com/rocketden/main/dto/problem/ProblemTestCaseDto.java
@@ -11,4 +11,5 @@ public class ProblemTestCaseDto {
     private String input;
     private String output;
     private boolean hidden;
+    private String explanation;
 }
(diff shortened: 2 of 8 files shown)


zpChris:
[Create field for problem test case explanation](https://trello.com/c/dT2A1KYT/172-create-field-for-problem-test-case-explanation)

alankbi (approved):
Approved!

I will leave a note though: I always imagined that the explanation would be like an example input/output as part of the description displayed on the left panel similar to LeetCode, and many times there isn't an explanation. Thoughts? It might end up that the explanation column is null for all or all but one of the test cases for each problem. This isn't really an issue though since it's easy to discuss or flesh out later when we get to that point.

helenli8181 (approved):
This looks great to me!

zpChris:
> Approved!
> 
> I will leave a note though: I always imagined that the explanation would be like an example input/output as part of the description displayed on the left panel similar to LeetCode, and many times there isn't an explanation. Thoughts? It might end up that the explanation column is null for all or all but one of the test cases for each problem. This isn't really an issue though since it's easy to discuss or flesh out later when we get to that point.

I thought of this too, which may indeed be the case -- but I also thought that this may also be used if we make any tests public **after** the game, if there is any test that a player fails that they may be able to see this. Either way, I still think this setup works (though I understand that it would make less sense in the case that you described), so I'll stick with this PR.

zpChris:
I accidentally merged this first into `QuickBuild` instead of merging `QuickBuild` into master first -- sorry about the mistake. I am working on fixing the issues with merging `QuickBuild` into master now.

PR #84, opened 2021-01-04
Add a Problem list to the Game class.

### LINKED ISSUE #83: Different Problems Given in Same Room
When multiple users are in the same Room, different Problems are given to the different users. This is a bug as the same Problem should be given to all users in the same Room.

See the attached screencast for the error: [Different Problems Given Bug](https://www.loom.com/share/2c135949022549f981890ea9c3d3d472)

zpChris:
### List of Changes:

- Fixes #83 where different problems are given to different players in the same game
- Add a `List<Problem>` field to the `Game` object so that all users get the same problem and multiple problems can be given in the future.
- Remove the frontend usage of the `getRandomProblem` API call, as the `Problem` is instead attached to the `Game` object (thus giving all players the same problem).
- Add `toObjectList` method in `UtilityTestMethods.java` in order to convert a JSON list into a Java object (the typical Gson method does not work for JSON lists).
- Return an empty `List<Problem>` from `getProblemsFromDifficulty` if an `ApiException` is reached.
- Remove unused imports across various test files.

### Notes:

- For the purposes of the MVP, there is only a need for a single `Problem` in the game. However, I anticipate that we may increase the number of problems that it would take to win a game in the future, so I decided to change this to a `List<Problem>`. My reasoning was that it would make decision-making easier in the future, and for now, we can simply ask for one `Problem` for every game; however, admittedly, there's a chance we don't use this, so I'm open to removing the List and replacing it with a single `Problem`.
- I changed the `roomDto` field in the `Game` object to just be `room` -- I'm open to changing this, it was just personal preference as I started coding. Sorry for not asking this earlier on PR #80.
- If an `ApiException` occurs in the `getRandomProblems ` method, an empty `List<Problem>` is attached to the `Game` object; I am open to changing this, but I'm not sure what the best design choice would be.

### Screencast and Images:

[Screencast of Problem Attached to Game Object](https://www.loom.com/share/c5a59b710324421cb44e8705015cdcc7)

Image of Problem Attached to Game Object
Image of Problem Attached to Game Object


zpChris:
[Different problems given on game start](https://trello.com/c/QIhsaTa5/194-different-problems-given-on-game-start)

src/main/java/com/rocketden/main/game_object/Game.java
```diff
@@ -4,16 +4,20 @@
...
 import com.rocketden.main.model.Room;
+import com.rocketden.main.model.problem.Problem;
 
 @Getter
 @Setter
 public class Game {
 
     private Room room;
 
+    private List<Problem> problems;
```
alankbi, line 19: nit: set to `= new ArrayList<>()` so it's defaulted to an empty list

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -91,9 +98,28 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
     // Initialize and add a game object from a room object
     public void createAddGameFromRoom(Room room) {
         Game game = GameMapper.fromRoom(room);
+        game.setProblems(getProblemsFromDifficulty(room.getDifficulty(), 1));
         currentGameMap.put(room.getRoomId(), game);
     }
 
+    // Choose the problem based on problem difficulty settings (empty list for invalid request)
+    private List<Problem> getProblemsFromDifficulty(ProblemDifficulty difficulty, int n) {
```
alankbi, line 106: Suggestion: could we move this method to the problemService class? Or also now that we have an `n` parameter maybe we could also just modify the original `getRandomProblems` method (since we don't have a need for it anymore)

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -91,9 +98,28 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
+        try {
+            List<Problem> problems = new ArrayList<>();
+            for (ProblemDto problemDto : problemService.getRandomProblems(request)) {
+                problems.add(ProblemMapper.toEntity(problemDto));
+            }
+
+            return problems;
+        } catch (ApiException e) {
+            return new ArrayList<>();
+        }
```
alankbi, line 120: Personal preference but if an ApiException occurs (like invalid game setting) I feel like we would want to display that to the user rather than returning just an empty list - I'm open to thoughts on this though

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -91,9 +98,28 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
+    // Choose the problem based on problem difficulty settings (empty list for invalid request)
+    private List<Problem> getProblemsFromDifficulty(ProblemDifficulty difficulty, int n) {
+        ProblemSettingsDto request = new ProblemSettingsDto();
+        request.setDifficulty(difficulty);
+        request.setN(n);
+
+        try {
+            List<Problem> problems = new ArrayList<>();
+            for (ProblemDto problemDto : problemService.getRandomProblems(request)) {
+                problems.add(ProblemMapper.toEntity(problemDto));
```
alankbi, line 114: Now that we no longer need to call `getRandomProblems` from a GET request, maybe we could change it to return `Problem` instead of `ProblemDto` that way we don't need this conversion?

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -67,10 +69,17 @@ public ProblemDto getProblem(String problemId) {
...
     }
 
-    public ProblemDto getRandomProblem(ProblemSettingsDto request) {
+    /**
+     * Get a list of random problems with the provided parameters.
+     * 
+     * @param difficulty The problem difficulty the problems must match.
+     * @param n The number of problems to fetch.
+     */
+    public List<ProblemDto> getRandomProblems(ProblemSettingsDto request) {
```
alankbi, line 78: Same as before (just copied over) but basically a suggestion was to just return `Problem` instead, since we might not need this method to be called directly through a GET request anymore, and then we could just directly call this method from GameManagementService instead of getProblemsFromDifficulty

alankbi (changes requested):
Left some suggestions but overall no major issues!

src/main/java/com/rocketden/main/dto/problem/ProblemMapper.java
```diff
@@ -18,6 +19,18 @@ public static ProblemDto toDto(Problem entity) {
         return mapper.map(entity, ProblemDto.class);
     }
 
+    public static Problem toEntity(ProblemDto dto) {
```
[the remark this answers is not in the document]
alankbi: Although depending on your thoughts on my other suggestion this might not even be necessary

src/main/java/com/rocketden/main/dto/game/GameDto.java
```diff
@@ -3,13 +3,16 @@
...
+import java.util.List;
 import java.util.Map;
 
 import com.rocketden.main.dto.room.RoomDto;
+import com.rocketden.main.model.problem.Problem;
 
 @Getter
 @Setter
 public class GameDto {
+    private List<Problem> problems;
```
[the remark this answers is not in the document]
zpChris: Turns out I already fixed this but forgot to push the latest update that modified tests.

src/main/java/com/rocketden/main/dto/problem/ProblemSettingsDto.java
```diff
@@ -1,11 +1,15 @@
...
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 
 @Getter
 @Setter
+@EqualsAndHashCode
 public class ProblemSettingsDto {
     private ProblemDifficulty difficulty;
+    private Integer n;
```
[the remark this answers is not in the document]
zpChris: Sounds good, done!

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: I was thinking that the empty `ArrayList` could be a signal of this exception, since I wasn't sure how I would pass along this `ApiException` to the frontend in this framework -- though I agree with you, and I'm open to ideas on how to implement this.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: I moved it to the `ProblemService` class 👍  I kept the `getRandomProblems` method because I figured we may want to keep the API call (I'm still open to removing it, but since we already have all the tests and everything configured, I wanted to be more cautious about removing it).

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: I still kept the GET request and that setup in `ProblemController.java` as well as the associated tests, since I could see us using that in the future, but I made the modification to return `Problem` instead of `ProblemDto`. This does require a conversion in `ProblemController.java`, which is the only one across the controllers, so let me know if anything should change there.

src/main/java/com/rocketden/main/service/ProblemService.java
zpChris: Done! 👍

src/main/java/com/rocketden/main/dto/problem/ProblemMapper.java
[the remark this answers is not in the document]
zpChris: Yeah, after modifications I don't need this method anymore, so I removed it.

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: I haven't tested this so I could be way off, but if you don't catch the ApiException, is it possible it'll just propagate that error to the frontend?

zpChris:
For some reason, I'm running into a test failure of `RoomSocketTests.socketReceivesMessageOnStartGame` that I cannot seem to fix. The line `repository.findAll()` in the method `getProblemsFromDifficulty` in `ProblemService.java` returns `null` or an empty `ArrayList`, even though I created a mock to return a proper list of problems for this purpose. I'll try to figure this out sometime after the meeting with pair programming.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Yeah I think that's right -- I'm running into some other issues after making these changes though, I'll try to work with you to solve them in the morning.

src/main/java/com/rocketden/main/exception/ProblemError.java
```diff
@@ -9,7 +9,8 @@
 public enum ProblemError implements ApiError {
 
     BAD_SETTING(HttpStatus.BAD_REQUEST, "Please choose either Easy, Medium, or Hard " +
-            "(or Random if choosing a room difficulty)."),
+            "(or Random if choosing a room difficulty), as well as a number of problems."),
```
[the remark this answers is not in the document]
zpChris: What are you looking to distinguish -- between the difficulty levels and the number of problems? Should I change this to `NULL_DIFFICULTY` and `NULL_NUM_PROBLEMS` errors?

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: The error seems to propagate to the frontend, but never reaches the game page, and seems to cause other errors -- we may need to restructure this. There are no "HARD" difficulty problems in the database, so you can see the error propagation and resulting effects in the [this screencast](https://www.loom.com/share/759c0d2b4b4a41afbf515b2e50d65b11). I can look into fixing this, I'd be interested in your feedback as well.

zpChris:
Re-requested review @alankbi -- I think there may still be issues with this PR, namely as it connects to propagating errors properly to the frontend, but I wanted to get your feedback on these changes so far.

src/main/java/com/rocketden/main/exception/ProblemError.java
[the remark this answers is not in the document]
alankbi: Yeah, in that case that might make more sense (although I would think something like "bad" instead of "null") - that way it can distinguish between bad difficulty setting vs. bad choice for num problems.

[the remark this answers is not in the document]
zpChris: Ok, I'll make this change.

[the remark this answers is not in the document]
zpChris: Done! 👍

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Nice catch - in that case I think we can leave that for a future change bug fix (or possibly just leave it as is). I added a card for it just in case though

src/main/java/com/rocketden/main/model/problem/Problem.java
```diff
@@ -36,7 +37,7 @@
      * is deleted or test cases are removed from this list, the test cases will also be deleted.
      * Setter is set to private to ensure proper use of addTestCase and removeTestCase methods.
      */
-    @OneToMany(mappedBy = "problem", cascade = CascadeType.ALL, orphanRemoval = true)
+    @OneToMany(mappedBy = "problem", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.EAGER)
```
alankbi, line 40: Just out of curiosity how come this requires the eager fetch?

alankbi (approved):
Looks good!

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Ok awesome thanks -- I added the Loom screencast to the card as well.

src/main/java/com/rocketden/main/model/problem/Problem.java
zpChris: To be honest, I did not research it past finding [this Stack Overflow post](https://stackoverflow.com/questions/22821695/how-to-fix-hibernate-lazyinitializationexception-failed-to-lazily-initialize-a); I was getting errors about LazyInitializationException as it related to the Problem class and its ProblemTestCase field, so this seemed to fix it.

helenli8181 (approved):
This looks great!

PR #86, opened 2021-01-08
Add Timer to frontend and backend, end game on timer end.

zpChris:
### List of Changes:

- Add a `GameTimer` object to the backend that starts on game start, and ends the game and redirects users on timer end through a socket message on the backend.
- Create a countdown timer on the frontend that refreshes every second, and recalculates time left every time from `Date.now()` and the `endTime` variable, in order to maintain synchronized regardless of refresh or other interruption.
- Unsubscribe from the specific room subscription channel and resubscribe to this same channel, replacing the subscription callback to expect and handle the `GameDto` object rather than the `RoomDto` object.
  - This is necessary because I alert the frontend to redirect the users on timer end through the `GameDto` object (has a boolean field on the timer labeled `timeUp`); I also set this up so future updates can be sent through this channel.
- Add `@EqualsAndHashCode` label to various DTO classes to enable correct testing.
- Create `TimerError` error that handles `null` errors.
  - Note: I am open to removing this since it is not a publicly-accessible API call through Postman, and is just used internally.
- Create `EndGameTimerTaskTests.java` class to test all edge cases and confirm that socket message is sent after the given amount of time.
- Create a `toObjectLocalDateTime` function to specifically handle [the `LocalDateTime` Gson "to object" issue](https://stackoverflow.com/questions/22310143/java-8-localdatetime-deserialized-using-gson).

### Notes:

- The `GameTimer` object uses the `Timer` object (`java.util.Timer`) and the `TimerTask` object (`java.util.TimerTask`), which runs the timer in a separate thread to not interrupt the rest of the code. [Read more on `TimerTask`, the overridden `run()` function, and the `schedule` function.](https://www.journaldev.com/1050/java-timer-timertask-example)
- The `GameClock` on the frontend is set up to handle the following edge cases:
  - Time left becomes negative (show `00:00`).
  - Time left on minutes or seconds is `0`, show `00` for that field.
  - Time left on minutes or seconds is a single digits, show `0` plus that digit.
  - Note: I made an assumption that the time remaining would never be more than an hour. I can modify this assumption in the future if we make longer games, but for now, it will show all of those hours in the minutes field (i.e., `2 hours = 120:00`).
- As described in the screencast attached in the below section, the timer end is triggered on the backend, so no code on the frontend can hack this process.
- [The `subscription?.unsubscribe()`](https://github.com/rocketden/main/compare/master...DisplayTimer#diff-0e48089598b398a1fce461bcccf5ff61693ec954880e97a02b2ae39c7cb8a31cR231) line causes a `eslint(no-unused-expressions)` that I cannot seem to fix -- if @alankbi or @helenli8181 knows a fix (or if I should just disable that line), please note in the review!
- I pass the SocketService into the `EndGameTimerTask` class as a parameter in the constructor (the only solution I've found, as autowiring through constructor and just autowiring the field causes other issues).

### Screencast and Images:

[Screencast of the Game Timer Redirecting Users to Results Page](https://www.loom.com/share/23f37232005949d8bc2226d10bfe086f)

**Image of the Game Timer with Console**

Image of the Game Timer with Console


zpChris:
[Display timer and end game once timer ends](https://trello.com/c/ZlgtdrYI/211-display-timer-and-end-game-once-timer-ends)

frontend/src/views/Game.tsx
```diff
@@ -47,6 +54,65 @@ function GamePage() {
    */
   useBeforeunload(() => 'Leaving this page may cause you to lose your current code and data.');
 
+  // Calculate and set the new clock on the frontend.
+  const calculateSetClock = useCallback((gameTimerParam: GameTimer) => {
```
alankbi, line 58: Is it possible to move this to a separate file in like a helper method? Or maybe create a new component like GameClock or something that takes in the gameTimerParam + an end game callback as props and then handles all the counting down stuff in that class

frontend/src/views/Game.tsx
```diff
@@ -47,6 +54,65 @@ function GamePage() {
...
+
+      const seconds: number = Math.floor(newCurrentClock % 60);
+      let secondsStr: string;
+      if (seconds >= 10) {
+        secondsStr = `${seconds}`;
+      } else if (seconds > 0 && seconds < 10) {
+        secondsStr = `0${seconds}`;
+      } else {
+        secondsStr = '00';
+      }
```
alankbi, line 80: Nice work on figuring out all this

src/main/java/com/rocketden/main/dto/game/GameTimerMapper.java
```diff
@@ -0,0 +1,23 @@
+package com.rocketden.main.dto.game;
+
+import com.rocketden.main.game_object.GameTimer;
+
+public class GameTimerMapper {
```
alankbi, line 5: Could you add a GameTimerMapperTests class?

src/main/java/com/rocketden/main/exception/TimerError.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.rocketden.main.exception.api.ApiError;
+import com.rocketden.main.exception.api.ApiErrorResponse;
+import lombok.Getter;
+import org.springframework.http.HttpStatus;
+
+@Getter
+public enum TimerError implements ApiError {
+
+    NULL_SETTING(HttpStatus.BAD_REQUEST, "The game, associated game timer, room, room ID, and socket service must not be null.");
```
alankbi, line 11: When does this error get triggered? Like would it be a user error or an "us" error (i.e. like a 500 internal server mistake perhaps)?

src/main/java/com/rocketden/main/game_object/GameTimer.java
```diff
@@ -0,0 +1,45 @@
...
+ * The Timer class handles the timing for the Game.
+ * This class is dynamic, so depending on player actions time could
+ * be paused, added, or subtracted.
+ */
+@Getter
+@Setter
+public class GameTimer {
+
+    // 15 minute duration for the GameTimer, in seconds.
+    public static final Long DURATION_15 = (long) 900;
```
alankbi, line 19: Is there any specific reason for choosing long here? Not that there's anything wrong with it of course but mainly curious

src/main/java/com/rocketden/main/game_object/GameTimer.java
```diff
@@ -0,0 +1,45 @@
...
+     * Instantiate the GameTimer class, and schedule the end game task after
+     * timer ends.
+     * 
+     * @param duration The duration, in seconds, until the timer ends.
+     */
+    public GameTimer(Long duration) {
+        this.duration = duration;
+        this.endTime = this.startTime.plusSeconds(duration);
+        this.timer = new Timer();
+    }
```
alankbi, line 31: nit: Move constructor below all the fields

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -81,17 +82,29 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
+    }
+
+    // Set and start the Game Timer.
+    public void setStartGameTimer(Game game, Long duration) {
+        GameTimer gameTimer = new GameTimer(duration);
+        game.setGameTimer(gameTimer);
+
+        // Schedule the game to end after <duration> seconds.
+        EndGameTimerTask endGameTimerTask = new EndGameTimerTask(socketService, game);
+        gameTimer.getTimer().schedule(endGameTimerTask, duration * 1000);
```
alankbi, line 107: Could this method be called from the `createAddGameFromRoom` method or absorbed into it or absorbed in the `GameMapper.fromRoom` method? I can imagine sometime in the future the duration might come from the room settings, though I guess that means we don't have to worry too much about this now

alankbi (changes requested):
Looks good! Only had a couple of suggestions/questions but otherwise very impressive

frontend/src/views/Game.tsx
zpChris: Yeah I intended to create a new component at some point -- I can do that now!

src/main/java/com/rocketden/main/dto/game/GameTimerMapper.java
zpChris: Done!

src/main/java/com/rocketden/main/exception/TimerError.java
zpChris: I think it would be an "us" error only -- I'm open to removing this if we want, what do you think?

src/main/java/com/rocketden/main/game_object/GameTimer.java
zpChris: I chose long because the inherent Java functions `plusSeconds` and `schedule` take long as the input, that's all.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: It could be absorbed / called from either of those methods -- do you have a preference? I'm open to changes on this.

frontend/src/views/Game.tsx
zpChris: Done! I will note that when I was doing this and testing the result, it seems like on the game page the socket is being subscribed to twice -- I'm not sure why this is, I don't think it's because of any change I made here, but there does seem to be an error somehow.

zpChris:
The game seems to be triggering two subscriptions: see [my comment](https://github.com/rocketden/main/pull/86#discussion_r554606720).

frontend/src/views/Game.tsx
zpChris: I also want to note that there is a brief moment while the timer shows "00:00" before the `GameTimer` is transmitted through. The timer then starts about 5 seconds after the actual timer amount; so if the timer is 15:00, the first time you usually see is 14:55. If this is an issue, let me know 🙂

zpChris: I fixed the issue with the double-subscribed socket (I added a dependency in a `useEffect` call that was not necessary). 👍

frontend/src/components/game/GameTimerContainer.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }, [updateClock, props]);
+
+  return (
+    <div>
+      Time:
+      {' '}
+      {(currentClock) ? currentClock.minutes : '00'}
+      :
+      {(currentClock) ? currentClock.seconds : '00'}
```
alankbi, line 69: Maybe to deal with the 00:00 showing at the beginning when it's null, is there a way to show something like "Loading..." or "Starting..." or something? Idk if that would be more intuitive or not tho so feel free to leave as is otherwise

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: No preference really - maybe just move the call to inside the `createAddGameFromRoom` method? Might be the easiest/quickest solution

alankbi (approved):
1-2 last comments but otherwise approved! 

I do have one last question though - I noticed in the screencast that at one point a "socket already connected" error showed up - do you know why this is the case? I thought #79 fixed that but maybe I didn't get all edge cases or something new came up

frontend/src/components/game/GameTimerContainer.tsx
zpChris: Yeah I think that would be more intuitive -- I'll try to make that change. 👍

zpChris:
> 1-2 last comments but otherwise approved!
> 
> I do have one last question though - I noticed in the screencast that at one point a "socket already connected" error showed up - do you know why this is the case? I thought #79 fixed that but maybe I didn't get all edge cases or something new came up

I don't know unfortunately :/ Let me know if I can help seek out edge cases though.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Sounds good -- fixed!

frontend/src/components/game/GameTimerContainer.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { useCallback, useEffect, useState } from 'react';
+import { GameClock, GameTimer } from '../../api/GameTimer';
+
+type GameTimerProps = {
+  gameTimer: GameTimer | null,
+};
+
+// This function refreshes the width of Monaco editor upon change in container size
```
helenli8181, line 8: nit: may want to update this comment :)

frontend/src/views/Game.tsx
```diff
@@ -82,6 +108,13 @@ function GamePage() {
     setSubmission(tempSubmission);
   };
 
+  // Redirect user to game page if room is active.
```
helenli8181, line 111: nit: and this comment

helenli8181 (approved):
Approved! I just left a couple nits where some comments could be updated.

frontend/src/components/game/GameTimerContainer.tsx
zpChris: Ah thanks for the heads up -- done 👍

PR #89, opened 2021-01-11
Add numProblems as an updatable room setting

alankbi:
### List of Changes:

* Add numProblems as a customizable field to room objects
* Use numProblems setting from rooms to initialize game objects
* Create potential fixes for the invalid numProblems errors 
  * Set a max of 10 numProblems
  * If numProblems is more than total problems than exist, return as many problems as exists
  * Prevent room from being active if game creation fails

### Notes:

* I didn't add the ability to customize numProblems on the frontend since for the MVP I'm guessing we'll only be able to support solving one problem per game (the first problem) - let me know what your thoughts are on that

### DIFF
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -92,8 +92,13 @@ public class ProblemService {
             throw new ApiException(ProblemError.NOT_FOUND);
         }
 
-        if (numProblems <= 0 || (numProblems > problems.size())) {
+        if (numProblems <= 0) {
             throw new ApiException(ProblemError.INVALID_NUMBER_REQUEST);
+        }
+
+        // If the user wants more problems than exists, just return all of them
+        if (numProblems > problems.size()) {
+            return problems;
         }
 
         // Get numProblem random integers used to map to problems.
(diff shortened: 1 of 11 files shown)


alankbi:
[Allow numProblems to be customized](https://trello.com/c/vs9UC6ta/216-allow-numproblems-to-be-customized)

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -92,10 +92,15 @@ public ProblemDto getProblem(String problemId) {
             throw new ApiException(ProblemError.NOT_FOUND);
```
alankbi, line 92: Not allowed to comment on it but thoughts on changing lines 78-82 to return a generic `ProblemError.EMPTY_FIELD` error if either setting is missing?

zpChris:
> * I didn't add the ability to customize numProblems on the frontend since for the MVP I'm guessing we'll only be able to support solving one problem per game (the first problem) - let me know what your thoughts are on that

Yeah I agree, I think for the MVP it'll just be one problem at a time 👍

src/main/java/com/rocketden/main/service/ProblemService.java
zpChris: Yeah I'm ok with that 👍

zpChris (approved):
LGTM! No nits or comments at all 🎉 😃

zpChris:
I merged #86 so I'd advocate pulling into this branch to ensure all the tests and whatnot still work 👍

PR #91, opened 2021-01-13
Add Color to Player class to create Player Icon.

zpChris:
### List of Changes:

- Add `Color` class on frontend and backend.
- Create `Player` class on frontend and provide the relevant `toDto` functions and tests.
- Clean up old tests by removing unused & unnecessary imports.

### Notes:

- This PR is essentially incomplete and meant to be merged into another, as there was similar work on creating the `Player` objects and whatnot in branch `ConnectSubmitEndpoint`.
- I think you could remove the `PlayerMapper.toDto` function if you wanted, as this is essentially the same as the `mapper` function we had before; however, we could keep it if more complex mapping may be required in the future, including potentially a ranking of the different players or submissions or whatnot.

### Screencast and Images:

There are no changes to the UI and thus no attached screencasts or images.


zpChris:
[Assign color, get players, and display player icons on frontend](https://trello.com/c/EOkhfmNp/218-assign-color-get-players-and-display-player-icons-on-frontend)

src/main/java/com/rocketden/main/dto/game/PlayerMapper.java
```diff
@@ -0,0 +1,48 @@
...
+import org.modelmapper.convention.MatchingStrategies;
+
+public class PlayerMapper {
+
+    protected PlayerMapper() {}
+
+    public static PlayerDto toDto(Player player) {
+        if (player == null) {
+            return null;
+        }
```
alankbi, line 20: I saw you addressed this in your notes comments; maybe we can keep this method but continue to use the loose mapper object? That way if more fields are added they should be taken care of automatically, and if we need to do any complex stuff (like sorting) we could do it after using the mapper

src/main/java/com/rocketden/main/util/Color.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Getter
+@Setter
+@EqualsAndHashCode
+public class Color {
+
+    private static final String INVALID_COLOR_NUM =
+        "The r, g, and b fields must all be between 0 and 255.";
+
+    public Color(int r, int g, int b) {
```
alankbi, line 15: This looks great; however what are your thoughts on simply using strings such as `#0099FF`? My only thinking behind this is that it might be easier to integrate those with styled-components (a component can accept a color string prop and we can assign it by simply using `background-color: ${({ color }) => color};`

However since we already have this Color class I can understand if we want to keep it this way too - I'm sure we'll be able to figure out a way to parse it on the frontend

src/main/java/com/rocketden/main/dto/game/GameMapper.java
```diff
@@ -51,26 +51,19 @@ public static Game fromRoom(Room room) {
...
 
+        // Create players and assign colors in order.
+        int index = 0;
         Map<String, Player> players = game.getPlayers();
         for (User user : room.getUsers()) {
-            Player player = playerFromUser(user);
+            Player player = PlayerMapper.playerFromUser(user);
+            player.setColor(Utility.COLOR_LIST.get(index));
             players.put(user.getUserId(), player);
+            index++;
```
alankbi, line 61: To prevent out of bounds errors (some users might have the same colors I guess)

Also, do we ever want to randomize the colors? Maybe like create a copy of `COLOR_LIST` and shuffle it would be a quick and easy way - this is beyond the MVP tho so doesn't matter either way

```suggestion
            index = (index + 1) % Utility.COLOR_LIST.size();
```

alankbi (commented):
Looks good - not sure whether to approve or request changes or basically what direction we're doing in terms of merging this - do we want to go to master or ConnectSubmitEndpoint?

src/main/java/com/rocketden/main/util/Color.java
zpChris: I think that'd make sense (I didn't do this initially because I thought it wouldn't work for some reason, but I was wrong) -- I'll go ahead and change this after reviewing all of these PRs.

src/main/java/com/rocketden/main/dto/game/GameMapper.java
zpChris: The idea was that since there is a max of ten players in a game (I forgot where I remember this from tbh), the `Utility.COLOR_LIST` field will always have enough colors for all the players. I can still add this though, in case that might change in the future -- though I think it'd be good to eventually have some secondary attribute that we can then pair to make like "100" different combinations, and thus have a unique symbol for each person.

I'm cool with randomizing the colors -- I initially thought of pairing more "opposite" colors across the different stages, like red and blue rather than red and orange and whatnot, but I'm open to whatever for this. I'll implement these after reviewing the remaining PRs, and I can commit this suggestion as well if you think that'd be a good idea.

zpChris:
> Looks good - not sure whether to approve or request changes or basically what direction we're doing in terms of merging this - do we want to go to master or ConnectSubmitEndpoint?

I think it'd be best to merge with **ConnectSubmitEndpoint**, because there may be some choices in this PR that conflict with **ConnectSubmitEndpoint**, and since that PR has the actual features in place its changes should take precedence (IMO).

src/main/java/com/rocketden/main/dto/game/PlayerMapper.java
zpChris: Yeah I think that makes sense -- I'll revert this back after reviewing the remaining PRs as well.

zpChris: Done! ✅

alankbi: Oh I mean I was totally fine with leaving the toDto method (I think it actually makes more sense too) but just keeping the mapper.map logic as the bulk of the implementation 

Either way though not too big of a deal so whatever you think makes sense

zpChris: Ah whoops lol -- maybe this is because I already switched it back, but since the code I provided doesn't do anything more that the `mapper.map` function does, I think it makes sense to keep just that until we actually need the `toDto` method.

src/main/java/com/rocketden/main/util/Color.java
zpChris: @alankbi If I switch the `Color` object to use a String, were you saying that I really don't need a `Color` class in the first place? Should I just replace that with a String?

alankbi: I'm not sure - maybe it could still be used to hold all the different colors? I don't have too much of a preference on that though

zpChris: Ok, I'll just keep the Color class because perhaps it could be expanded in the future to include other symbols or whatnot.

src/main/java/com/rocketden/main/dto/game/GameMapper.java
zpChris: Should I do randomization in this PR @alankbi?

alankbi: Nah I think that can probably be for a later PR

alankbi (approved):
Changes look good!

zpChris:
> No preference either - feel free to merge into master now or wait till ConnectSubmitEndpoint is merged, I think both are fine

I'll go ahead and merge now!

PR #92, opened 2021-01-14
Allow game duration to be set by host on lobby page

alankbi:
### List of Changes:

* Add duration field to room objects
* Allow host to set duration of game from frontend
* Add general tests and socket tests to ensure timer end sends socket message 
* Small code cleanups with error messages and host checks

### Screencast and Images:

[Setting timer on lobby page](https://www.loom.com/share/c45514457e6f4f21940648da282f389e)


zpChris:
[Allow duration to be customized](https://trello.com/c/qKPKzGjd/219-allow-duration-to-be-customized)

frontend/src/views/Join.tsx
```diff
@@ -129,6 +129,7 @@ function JoinGamePage() {
               setFocusInput(false);
             }}
             onKeyPress={(event) => {
+              setError('')
```
zpChris, line 132: What's the thinking here? -- when the user starts typing a username or `roomId` again, to remove the error if one is displayed? (Same question for the rest of these same changes.)

frontend/src/views/Lobby.tsx
```diff
@@ -265,13 +296,33 @@ function LobbyPage() {
...
+          : 'The game will last for the following minutes:'}
+      </Text>
+      <NumberInput
+        min={1}
+        max={60}
+        value={duration}
+        disabled={!isHost(currentUser)}
+        onChange={(e) => {
+          const newDuration = Number(e.target.value);
+          if (newDuration >= 0 && newDuration <= 60) {
```
zpChris, line 319: The instructions say between 1 and 60 minutes, but you allow any duration between 0 and 60 minutes here. Is there a reason for this discrepancy or was this purposeful? (I saw in the example screencast that you chose 0.1 minutes, for example.)

Also, are we allowing any unit of time, like "30.5910239" or something random as a valid time?

frontend/src/views/Lobby.tsx
```diff
@@ -265,13 +296,33 @@ function LobbyPage() {
...
       ))}
+
+      <MediumText>Duration</MediumText>
+      <Text>
+        {isHost(currentUser) ? 'Choose a game duration between 1-60 minutes:'
+          : 'The game will last for the following minutes:'}
+      </Text>
+      <NumberInput
+        min={1}
+        max={60}
```
zpChris, line 314: No change necessary here, I just wanted to note something UI-wise to see if this was a minor annoyance for anybody else -- I typically type in numbers into this fields by doing `command + A` -> `delete` -> `[number]`; however, here, the field is auto-populated with a `0` after pressing `delete`, which means I type in the number _then_ delete the `0`. This is clearly a small nit, but I wanted to see if it was easily possible to allow the field to simply have no input (for which we would return an error and set the duration to it's current backend value).

frontend/src/views/Lobby.tsx
```diff
@@ -265,13 +296,33 @@ function LobbyPage() {
...
+        value={duration}
+        disabled={!isHost(currentUser)}
+        onChange={(e) => {
+          const newDuration = Number(e.target.value);
+          if (newDuration >= 0 && newDuration <= 60) {
+            setDuration(Number(e.target.value));
+          }
+        }}
+      />
+      {isHost(currentUser) ? <SmallButton onClick={updateRoomDuration}>Save</SmallButton> : null}
```
zpChris, line 324: Just wanted to bring up that we could potentially not use the "Save" button, and just auto-save on every change, **or** just take the current value of the field when "Start Game" is pressed.

Another idea would be to auto-save the value whenever the user clicks away from the input again. Thoughts on these alternative approaches? Obviously it's no big deal for the MVP, but might be nice in the future to have a cleaner frontend system from which to build & design more.

frontend/src/views/Lobby.tsx
```diff
@@ -59,10 +61,15 @@ function LobbyPage() {
     setRoomId(room.roomId);
     setActive(room.active);
     setDifficulty(room.difficulty);
+    setDuration(room.duration / 60);
   };
 
+  // Function to determine if the given user is the host or not
+  const isHost = useCallback((user: User | null) => user?.nickname === host?.nickname, [host]);
```
zpChris, line 68: On PR #88, I [replaced the nickname comparison with `userId` comparison](https://github.com/rocketden/main/pull/88/files#diff-0e48089598b398a1fce461bcccf5ff61693ec954880e97a02b2ae39c7cb8a31cR334) -- just heads up, since there will be merge conflict issues here, and I might suggest comparing `userId`s instead because that seems like a more certain system long-term.

This does obviously come with the caveat that we have to determine how to hide `userId`s on the frontend, so there's a good argument for comparing by nickname, but either way we should keep it consistent by modifying #88 (the other PR) or #92 (this PR). (This same idea applies elsewhere for "nickname vs. `userId`" in the PR.)

src/main/java/com/rocketden/main/dto/game/GameTimerDto.java
```diff
@@ -10,8 +14,15 @@
...
 @EqualsAndHashCode
 public class GameTimerDto {
+
+    @JsonDeserialize(using = LocalDateTimeDeserializer.class)
+    @JsonSerialize(using = LocalDateTimeSerializer.class)
     private LocalDateTime startTime;
-    private Long duration;
+
+    @JsonDeserialize(using = LocalDateTimeDeserializer.class)
+    @JsonSerialize(using = LocalDateTimeSerializer.class)
```
zpChris, line 23: What do these fields do?

src/main/java/com/rocketden/main/dto/room/RoomDto.java
```diff
@@ -19,4 +19,5 @@
     private List<UserDto> inactiveUsers;
     private boolean active;
     private ProblemDifficulty difficulty;
+    private long duration;
```
zpChris, line 22: nit: Use `Long` instead of `long` here, _or_ replace the fields elsewhere to use `long`?

src/main/java/com/rocketden/main/exception/TimerError.java
```diff
@@ -8,6 +8,7 @@
 @Getter
 public enum TimerError implements ApiError {
 
+    INVALID_DURATION(HttpStatus.BAD_REQUEST, "Please enter a valid duration between 1-60 minutes."),
```
zpChris, line 11: nit: Update this field if necessary depending on how you handle the 0-1 minute category.

zpChris (changes requested):
This looks great overall! Although I haven't reviewed the full PR, I'm requesting changes because I'll be out most of today -- I've looked at eight files so far, I'll look at the remaining (and any additional changes if those have been made) when I get back.

frontend/src/views/Lobby.tsx
alankbi: Good catch - I'm okay with doing any of those options (1. allowing between (0, 60], 2. allowing between [1, 60], as well as a slightly unrelated 3. limiting to only whole numbers) - what are your preferences?

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -188,6 +189,7 @@ public void startAndGetGameSuccess() throws Exception {
 
         assertEquals(actual, gameDto.getRoom());
         assertEquals(1, gameDto.getPlayers().size());
+		assertEquals(GameTimer.DURATION_15, gameDto.getGameTimer().getDuration());
```
zpChris, line 192: nit: Fix tab vs. spacing indentation.

src/test/java/com/rocketden/main/socket/GameSocketTests.java
```diff
@@ -0,0 +1,145 @@
...
+
+            @Override
+            public void handleFrame(StompHeaders headers, Object payload) {
+                blockingQueue.add((GameDto) payload);
+            }
+        });
+    }
+
+    @Test
+    public void socketReceivesMessageOnGameOver() throws Exception {
```
zpChris, line 135: Thanks for making this! I should've done it in a previous PR, sorry about that -- appreciate it! 👍

zpChris (commented):
The rest of the PR looks great! I would approve, but I split this review up into two parts and on the first I mentioned some changes, so I'll just comment on this review.

frontend/src/views/Join.tsx
alankbi: Yeah that's what I was mainly thinking, so that once they start typing again (or submit/click a button again for the other places) the error would go away without having to refresh - open to reverting this given how small a change it is

frontend/src/views/Lobby.tsx
alankbi: Also, if we go with 3, should we also maybe change the backend to use minutes rather than seconds?

frontend/src/views/Lobby.tsx
alankbi: Good catch, I updated all instances to use userId now

src/main/java/com/rocketden/main/dto/game/GameTimerDto.java
alankbi: The GameSocketTests had issues converting messages to Java objects without these to help it serialize and deserialize the dates (I guess it has something similar to do with the `toObjectLocalDateTime` method you implemented in an earlier PR)

Not sure if this actually affects the behavior on the frontend (since it seemed like it worked fine in a non-test environment)

frontend/src/views/Lobby.tsx
alankbi: Good idea on that - I ended up going with the autosave when you click anywhere away from it (onBlur) for a balance between too often and too little saving

frontend/src/views/Game.tsx
```diff
@@ -60,6 +60,7 @@ function GamePage() {
       setSocketSubscribed(true);
 
       // Check if end game.
+      // TODO: after deconstructing game, move this to useEffect on timeUp
```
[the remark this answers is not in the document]
alankbi: The deconstructing part should be already implemented (either in master or one of the many open PRs haha) 

Sounds good tho for the useEffect part - since it's relatively easy (1-commit change) I just added it as part of the [end game when everyone submits story](https://trello.com/c/ilAQWiO8/225-after-everyone-submits-end-game-add-useeffect-variable) since I figure it groups nicely with that one - feel free to modify this tho

frontend/src/views/Lobby.tsx
alankbi: That makes total sense - made a commit to allow it to be empty

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
zpChris: That sounds good!

src/main/java/com/rocketden/main/dto/game/GameTimerDto.java
zpChris: Ah nice, I wonder if I could've just used that instead of the `toObjectLocalDateTime` method, as this is a lot simpler -- either way, sounds good.

frontend/src/views/Lobby.tsx
zpChris: I'm OK with any of that really -- I think I might lean towards 2 and 3; however, I'm still thinking that we keep the backend in seconds, as it'd still be nice to trigger different notifications and powerups on a second-by-second basis.

frontend/src/views/Join.tsx
zpChris: Nah, this seems like a good change, let's keep it.

src/main/java/com/rocketden/main/exception/TimerError.java
alankbi: Ended up going with options 2 and 3, so this should be good unchanged I think

frontend/src/views/Lobby.tsx
alankbi: Sounds good, implemented

zpChris (approved):
LGTM!

I will note that there is a small UI nit where [clicking "start game" immediately after changing the duration doesn't start the game](https://www.loom.com/share/e019830c9c364173a87b8451b4911640), but unless that's an easy fix, this is totally good for the MVP.

zpChris:
Actually, I'm getting the below error on testing, so don't merge quite yet -- running again.

Terminal Timeout Error

PR #93, opened 2021-01-15
Create leaderboard functionality on game page

alankbi:
### List of Changes:

* Change submission code to use CodeLanguage
* Send sorted leaderboard of players to frontend through sockets
* Populate submission with number of test cases from game's problem
* Display ranked players on game page by numCorrect and latest submission time
* Miscellaneous backend tests and UI improvements on the game page
* Separate lobby and game subscription endpoints to fix [disconnect bug](https://trello.com/c/gguR59gg/237-when-player-disconnects-on-the-game-page-the-socket-sends-a-roomdto-to-the-subscription-endpoint)

### Future Changes:

* ~Add socket test in GameSocketTests~ (implemented)
* ~Move leaderboard sorting to GameMapper class~ (implemented)
* Send socket message through notification pipeline on correct submission

### Screencast and Images:

[Game page with functioning leaderboard](https://www.loom.com/share/2d3e797be6dc47e1ae33b7f8d7b51d67)


alankbi:
[Connect submission results with a leaderboard](https://trello.com/c/m09JwAdg/223-connect-submission-results-with-a-leaderboard)

zpChris:
There are still test errors stemming from react-hooks errors & warnings (see image).

React-Hooks Warnings

frontend/src/components/card/LeaderboardCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+      onMouseLeave={() => setShowHover(false)}
+    >
+      <PlayerIcon color={color} />
+      <LowMarginText>{`${place}. ${getDisplayNickname()}`}</LowMarginText>
+
+      {showHover ? (
+        <HoverBar>
+          <SmallText>{getScoreDisplay()}</SmallText>
+          <SmallText>{`Last submitted: ${getSubmissionTime()}`}</SmallText>
+        </HoverBar>
```
zpChris, line 95: nit: Make this HoverBar slightly smaller or shift it somehow so that it is fully within the view of the screen.

HoverBar Shift

frontend/src/views/Game.tsx
```diff
@@ -104,8 +133,28 @@ function GamePage() {
...
+  };
+
+  const displayPlayerLeaderboard = () => {
+    return players.map((player, index) => (
+      <LeaderboardCard
+        player={player}
+        isCurrentPlayer={player.user.userId === currentUser?.userId}
+        place={index + 1}
+        color="blue" // TODO: merge with Chris's color PR
+      />
```
zpChris, line 156: The leaderboard doesn't seem to update correctly? It shifts briefly when the user first submits, then shifts back. See the [attached screencast](https://www.loom.com/share/6310b20e52844c6b9f1fe23c4b683de5).

frontend/src/components/card/LeaderboardCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+    return `${diffMinutes} min ago`;
+  };
+
+  return (
+    <Content
+      onMouseEnter={() => setShowHover(true)}
+      onMouseLeave={() => setShowHover(false)}
+    >
+      <PlayerIcon color={color} />
+      <LowMarginText>{`${place}. ${getDisplayNickname()}`}</LowMarginText>
```
zpChris, line 89: nit: Do you think you could set a standard width across all LowMarginText examples, which make up the "full" Player Card? I'm asking because sometimes the player card becomes off-center, and it seems like it'd be prettier design if it was equal width across all player cards, meaning that the row was centered at the center of the page.

Player Card Unequal Width

zpChris (changes requested):
This looks great -- love the new header design! Awesome work.

Just requesting changes because I think there may be a bug in the leaderboard, and because I left a few other minor nits.

frontend/src/api/GameTimer.ts
```diff
@@ -1,7 +1,7 @@
 export type GameTimer = {
   duration: number,
-  startTime: Date,
-  endTime: Date,
+  startTime: string,
+  endTime: string,
```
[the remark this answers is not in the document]
alankbi: Not sure - however, the reason why I changed this to string is because I think that more accurately reflects what is in the axios response (I think axios receives the date in string format but doesn't actually serialize it into a date object)

frontend/src/views/Game.tsx
```diff
@@ -52,11 +65,28 @@ function GamePage() {
...
 
+  const setStateFromGame = useCallback((game: Game) => {
+    setRoom(game.room);
+    setPlayers(game.players);
+    setGameTimer(game.gameTimer);
+    setProblems(game.problems);
+
+    // Print room and currentPlayer to prevent unused variable warning
+    console.log(room);
+    console.log(currentPlayer);
```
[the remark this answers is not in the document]
alankbi: I think this might actually create a circular dependency issue looking at the console haha (adding currentPlayer as a useCallback dependency causes the setCurrentPlayer method to trigger it over and over) - I fixed this in my next branch (which builds on this PR) since idk if it causes any noticeable UI bugs, but lmk and I can also implement that change here

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -36,15 +42,52 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
 
         player.getSubmissions().add(submission);
         player.setSolved(true);
 
+        // Sort list of players by who is winning
+        GameDto gameDto = GameMapper.toDto(game);
+        sortLeaderboard(gameDto.getPlayers());
+
+        // Send socket update with latest leaderboard info
+        // TODO: test socket in GameSocketTests when merged
```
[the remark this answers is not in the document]
alankbi: Yeah that was my plan since the GameSocketTests file hasn't actually been merged into master yet - it looks like that PR has been approved tho so I could also merge that PR and pull the file from master and implement it in this branch too - lmk

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
alankbi: Actually I ended up removing it but I think there's something still going on - see comment chain below

frontend/src/views/Game.tsx
alankbi: Nice catch, at first I thought it had to do with the circular thing mentioned above but I got rid of it and still have some weird subscription behavior going on in the console (which I assume has something to do with it shifting back): 

Screenshot

I'll try to figure it out tmrw but if I can't maybe we can do some pair programming haha

frontend/src/api/GameTimer.ts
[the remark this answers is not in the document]
zpChris: Ah yeah, I think you're right -- thanks!

src/main/java/com/rocketden/main/service/SubmitService.java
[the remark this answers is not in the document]
zpChris: If that's doable, I think that'd be great! The less TODOs in the PR, the better.

frontend/src/views/Game.tsx
zpChris: This sounds good -- just let me know, I'm definitely open to pair programming.

I do remember I got some sort of circular issue as well at some point, but I didn't mention it because I couldn't replicate it. We can try to figure this out sometime today if you want.

frontend/src/components/card/LeaderboardCard.tsx
alankbi: Good catch - ended up setting a width of 7rem on the leaderboard cards themselves

frontend/src/views/Game.tsx
alankbi: Was able to fix it - I got rid of the useCallback for setStateFromGame and simply moved the setCurrentPlayer logic to a separate useEffect block - this seemed to get rid of the constant resubscribing

alankbi: I'll resolve this as I think the leaderboard bug should be fixed, but feel free to verify on your end as well

frontend/src/views/Game.tsx
```diff
@@ -127,20 +176,23 @@ function GamePage() {
...
+        <FlexLeft>
+          Room:
+          {' '}
+          {roomId || 'N/A'}
+        </FlexLeft>
+        <FlexCenter>
+          <GameTimerContainer gameTimer={gameTimer || null} />
+        </FlexCenter>
+        <FlexRight>
+          <TextLink to="/">Exit Game</TextLink>
```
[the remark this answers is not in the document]
alankbi: Good idea - implemented this with a confirmation popup. One thing to note: the logic I wrote is only on the frontend, so on the backend they'd still be a part of the game. So like if someone submitted and got first place but decided to leave before the game officially ends, they'd still be on the leaderboard instead of removed. Personally I'm okay with that design choice, but we can also create a card in the backlog if we think they should also be removed on the backend too

src/main/java/com/rocketden/main/service/SubmitService.java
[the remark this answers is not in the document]
alankbi: Sounds good, I'll do that - also, since that knocks one of the items off the future changes list, I'll also go ahead and do the "move sort leaderboard to game mapper" feature in this PR too then (since the remaining changes would be small enough where it doesn't warrant a separate PR) - it means this PR will again be slightly bigger haha but hopefully it's not too much more (since it's just moving the logic to a different file)

Note: moving the sorting to game mapper is necessary because as it stands only the submit endpoint will sort the leaderboard before sending the socket update - moving the sorting to the toDto function will ensure all of them have the correct leaderboard (which I only realized once I already finished coding it)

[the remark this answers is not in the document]
zpChris: This all makes sense -- thanks for adding this!

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
zpChris: I think that choice makes sense -- we can revisit that in the future if it becomes a problem, but sounds good for now.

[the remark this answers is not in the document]
zpChris: I think it might make sense to disconnect at least -- still keeping the player's history and whatnot in the game, but making them inactive. This means that if they try to create a new room, they will connect to that instead; right now, this doesn't work as the user is still connected to the other room.

frontend/src/views/Game.tsx
zpChris: The bug is fixed, but there's another note -- this might depend on the game type or settings we choose to implement, but if Player 1 submits correctly, then Player 2 submits correctly, then Player 1 submits correctly _again_, the leaderboard puts Player 1 **behind** Player 2. I'm not sure if this is intentional or not because Player 1 already got the best solution, but I wanted to note this down and see if we should change this @alankbi.

frontend/src/views/Game.tsx
```diff
@@ -140,22 +173,48 @@ function GamePage() {
...
+      .then((res) => {
+        setSubmission(res);
+        checkSendTestCorrectNotification(res);
+      })
+      .catch((err) => setError(err));
+  };
+
+  const exitGame = () => {
+    if (window.confirm('Exit the game? You will not be able to rejoin.')) {
+      history.replace('/');
```
zpChris, line 205: Per [this discussion](https://github.com/rocketden/main/pull/93#discussion_r560563908), I think we may want to disconnect from the socket here as well.

zpChris (changes requested):
Requesting changes because I just left a few potential changes (only one on this review, but another on the general comments of the last review). Reviewing because I want to check out the PR after these changes, but I'm fairly certain I'll approve quickly on the next review.

zpChris:
Sorry for requesting changes again, I just did that because I think there may be more discussion, but I'm fairly certain we'll merge this through on the next review.

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
alankbi: 👍

frontend/src/views/Game.tsx
alankbi: Yeah so the current implementation is by the latest submission only - so if someone gets 100% but decide to submit again then their latest submission would count (and penalize them for time of submission). If we want to change this I can think of two alternatives (although there might be better ones out there): 1. change implementation to rank people by their best solutions, 2. prevent additional submissions once people finish solving the problem. Thoughts?

zpChris: Those are good alternatives -- I think I might lean towards #1? Since I can see preventing additional submissions in some formats but I think it'd be good to stay consistent across most formats that the best solution is the one that is preserved.

alankbi:
After merging with the color PR: 

Game page with colored leaderboard icons

frontend/src/views/Game.tsx
alankbi: 👍

frontend/src/views/Game.tsx
alankbi: Sounds good - implemented

zpChris:
> After merging with the color PR:
> 
> Game page with colored leaderboard icons

Maybe I'll choose some better color combinations in a future PR haha

zpChris (approved):
LGTM! Thanks for making all of those subsequent changes.

PR #94, opened 2021-01-18
Create backend notification service, send "time left" updates.

zpChris:
### List of Changes:

- Create infrastructure for backend notification service. (Example in Postman.)
  - Add error checking for a variety of exceptions, to ensure valid notifications are provided.
- Add tests for the notification service API.
- Trigger "time left" notifications from the backend, which updates users when there are:
  - 60 minutes left,
  - 30 minutes left,
  - 15 minutes left,
  - 5 minutes left,
  - 1 minute left,
  - 30 seconds left, and
  - 10 seconds left.
- Remove unused autowired fields from `LiveGameService.java` and other classes.

### Notes:

- A "TODO" is left in the `GameTests.java` class under the `sendNotificationSuccess` method (see below). I've tried different solutions shown [here](https://stackoverflow.com/questions/40327970/deserialize-java-8-localdatetime-with-jacksonmapper/40330249), but they don't seem to work. I haven't tested putting a time in Postman either, since I'm uncertain what the format looks like.

> If time is replaced with `LocalDateTime.now()`, 400 error, as it cannot convert the time field into a JSON string. We could simply not have this field be set, and then create it on the backend, but if this could be figured out that'd be great.

- The `GameNotification` class is never actually used -- only `GameNotificationDto`. Should I keep the `GameNotification` class?

### Screencast and Images:

[Screencast of the "Time Left" Notification](https://www.loom.com/share/c04e2cb9a0e042f59bb99615635ee09c): This screencast was recorded when the user could set decimal increments of time on the frontend, which is used to show the "one minute left" time notification.

**Image of the "30 Seconds Time Left" Notification**
Image of the '30 Seconds Time Left' Notification


zpChris:
[Set up notifications in backend](https://trello.com/c/j4ErHxIx/222-set-up-notifications-in-backend)

zpChris:
Sorry this is such a long PR -- I was very surprised by its length, considering that this is basically a backend-only change, but over half of the changes are just tests. Let me know if I should split it up in any way, or walk through the PR to help explain it.

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -98,12 +98,63 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
+                new NotificationTimerTask(socketService, room.getRoomId(), "are thirty seconds");
+            timer.schedule(notificationTimerTask, (time - GameTimer.DURATION_30_SEC) * 1000);
+        }
+
+        if (GameTimer.DURATION_10_SEC < time) {
+            Timer timer = new Timer();
+            NotificationTimerTask notificationTimerTask =
+                new NotificationTimerTask(socketService, room.getRoomId(), "are ten seconds");
+            timer.schedule(notificationTimerTask, (time - GameTimer.DURATION_10_SEC) * 1000);
+        }
```
alankbi, line 157: Is it possible to simplify this with loops? Not sure the exact way it would work, but I could see like having a static list of the `[DURATION_X]` values (which we could then easily customize if we wanted to change notification frequencies). There might be some issues with the "are X min" messages, but I feel maybe we could get around that by changing the message to some other less rigid form like "Time remaining in this game is Xmin" (X might have to be in number format to allow it to work with the loop, e.g. `DURATION / 60` to get the correct minute count). Although this would only work with Xmin and not Xsecs notifications (but I think down to the last minute we could also just have the timer flashing red or something rather than a notification with only like 10 seconds left)

That paragraph might make it seem very complicated tho haha, so I'm open to talking about this to see whether it actually makes sense or not

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -133,12 +184,32 @@ public SubmissionDto submitSolution(String roomId, SubmissionRequest request) {
...
+        } else if (notificationDto.getInitiator() != null
+            && !game.getPlayers().containsKey(notificationDto.getInitiator().getUserId())) {
+            throw new ApiException(GameError.USER_NOT_IN_GAME);
+        } else if (notificationDto.getInitiator() == null
+            && Utility.initiatorNotifications.contains(notificationDto.getNotificationType())) {
+            throw new ApiException(GameError.NOTIFICATION_REQUIRES_INITIATOR);
+        } else if (notificationDto.getContent() == null
+            && Utility.contentNotifications.contains(notificationDto.getNotificationType())) {
+            throw new ApiException(GameError.NOTIFICATION_REQUIRES_CONTENT);       
+        }
```
alankbi, line 210: This is a personal preference so feel free to just resolve and ignore this:

nit: Change `else if` to `if` on new lines, since the else if is redundant (I feel like this helps with readability but again it's by personal preference so it's cool if you ignore)

src/main/java/com/rocketden/main/util/Utility.java
```diff
@@ -7,13 +7,24 @@
...
 
+    // List notifications that require an initiator.
+    public static final List<NotificationType> initiatorNotifications =
+        Arrays.asList(NotificationType.SUBMIT_CORRECT,
+            NotificationType.SUBMIT_INCORRECT, NotificationType.CODE_STREAK,
+            NotificationType.TEST_CORRECT);
+
+    public static final List<NotificationType> contentNotifications = 
+        Arrays.asList(NotificationType.SUBMIT_CORRECT,
+            NotificationType.TEST_CORRECT, NotificationType.TIME_LEFT);
```
alankbi, line 26: One thing I wanted to note is I feel like all of these notifications save `CODE_STREAK` would be generated on the backend (also to prevent random users from just spamming any notification they want) - that might also make it easier cause we wouldn't really need to worry about initiatorNotifications and contentNotifications and differentiating them with different API errors

This is just me jotting down my thoughts tho (feel free to add or refute or anything) - not sure if I'm really requesting any changes here

src/main/java/com/rocketden/main/game_object/GameNotification.java
```diff
@@ -1,19 +1,20 @@
...
-    // Convert a matching string (ignoring case) to enum object
-    @JsonCreator
-    public static GameNotification fromString(String value) {
-        try {
-            return GameNotification.valueOf(value.toUpperCase());
-        } catch (IllegalArgumentException e) {
-            throw new ApiException(GameError.BAD_SETTING);
-        }
-    }
+public class GameNotification {
```
alankbi, line 7: Is this class used anywhere or is it mainly the DTO? Mainly curious - I'm fine keeping it either way tho since we might need it later anyways

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -294,4 +304,39 @@ public void submitSolutionSuccess() throws Exception {
...
+        RoomDto roomDto = RoomTestMethods.setUpRoomWithOneUser(this.mockMvc, host);
+        startGameHelper(roomDto, host);
+
+        /**
+         * TODO: If time is replaced with LocalDateTime.now(), 400 error,
+         * as it cannot convert the time field into a JSON string. We could 
+         * simply not have this field be set, and then create it on the
+         * backend, but if this could be figured out that'd be great.
+         */
+
```
alankbi, line 323: Does the `toObjectLocalDateTime` not work in this scenario?

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -369,12 +369,11 @@ public void updateRoomSettingsInvalidDuration() {
...
+        request.setNumProblems(-1);
 
-        request.setDifficulty(ProblemDifficulty.EASY);
-        request.setDuration(-1L);
-
-        ApiException exception = assertThrows(ApiException.class, () -> roomService.updateRoomSettings(ROOM_ID, request));
-        assertEquals(TimerError.INVALID_DURATION, exception.getError());
+        ApiException exception = assertThrows(ApiException.class, () ->
+                roomService.updateRoomSettings(ROOM_ID, request));
+        assertEquals(ProblemError.INVALID_NUMBER_REQUEST, exception.getError());
```
alankbi, line 376: nit: Change this back (I also had this issue when fixing giant merge conflicts lol)

src/test/java/com/rocketden/main/socket/GameSocketTests.java
```diff
@@ -47,25 +51,23 @@
...
     // Predefine user and room attributes.
     private static final String NICKNAME = "rocket";
     private static final String USER_ID = "012345";
     private static final String NICKNAME_2 = "rocketrocket";
     private static final String USER_ID_2 = "098765";
-    private static final String INPUT = "[1, 8, 2]";
-    private static final String OUTPUT = "[1, 2, 8]";
-    private static final long DURATION = 5;
+    private static final String TIME_LEFT = "are ten seconds";
+    private static final long DURATION = 15;
```
alankbi, line 69: Note for future: If we continue expanding this file with more tests, we might want to consider testing the time ran out socket update in a separate PR (that way we don't have to wait for multiple seconds/minutes)

alankbi (commented):
Wow 😮   - good job on pushing all this out haha

Thanks for fleshing out the notification system - I left a few comments but they're mostly broader suggstions/questions so we can leave some of that discussion for a later PR if it's too much/annoying to deal with right now haha

alankbi (changes requested):
(forgot to click request changes lol)

src/main/java/com/rocketden/main/game_object/NotificationType.java
```diff
@@ -0,0 +1,19 @@
+package com.rocketden.main.game_object;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.rocketden.main.exception.NotificationError;
+import com.rocketden.main.exception.api.ApiException;
+
+public enum NotificationType {
+    SUBMIT_CORRECT, SUBMIT_INCORRECT, TEST_CORRECT, CODE_STREAK, TIME_LEFT;
```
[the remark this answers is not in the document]
zpChris: I think the first component makes sense -- the fact that it's split up on the backend and frontend is strange, though I tried to separate the dynamically changing content (on the backend) and the static content (on the frontend). The reason I didn't try to programmatically do this was because separating the "is X minutes" from "are X minutes" depending on X would require a decent number of checks, but I'm open to implementing this.

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -98,12 +98,63 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
 
         currentGameMap.put(room.getRoomId(), game);
+
+        // Create notifications for different "time left" milestones.
+        if (GameTimer.DURATION_60 < time) {
+            Timer timer = new Timer();
+            NotificationTimerTask notificationTimerTask =
+                new NotificationTimerTask(socketService, room.getRoomId(), "are sixty minutes");
+            timer.schedule(notificationTimerTask, (time - GameTimer.DURATION_60) * 1000);
+        }
```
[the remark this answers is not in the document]
zpChris: This is a good idea -- done! I also turned the repeated lines into a function to cut down on repeated code.

src/main/java/com/rocketden/main/game_object/NotificationType.java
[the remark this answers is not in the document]
zpChris: Ok, I made a map in the `GameTimer` class that handles the content for the `TIME_LEFT` notifications, since they are the main one with this problem. I'm open to further changes though, so I won't resolve this quite yet -- let me know.

GameTimer Duration-Content Map

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -98,12 +98,63 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
+                new NotificationTimerTask(socketService, room.getRoomId(), "are thirty minutes");
+            timer.schedule(notificationTimerTask, (time - GameTimer.DURATION_30) * 1000);
+        }
+        
+        if (GameTimer.DURATION_15 < time) {
+            Timer timer = new Timer();
+            NotificationTimerTask notificationTimerTask =
+                new NotificationTimerTask(socketService, room.getRoomId(), "are fifteen minutes");
+            timer.schedule(notificationTimerTask, (time - GameTimer.DURATION_15) * 1000);
+        }
```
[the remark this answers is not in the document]
zpChris: 1: You're right, I'll remove that 60-min notification.

2. Tbh I feel like that's a confusing mechanism -- I'm happy to implement it if you think it'd bother users though.

3. Sure, I've made this change 👍

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Sure I'm fine with that -- updated!

(I think I told you the exact opposite on another PR lol, to each their own -- it's not a huge preference by me though so I'll go by this from now on.)

src/main/java/com/rocketden/main/game_object/GameNotification.java
zpChris: Yeah good question, I put this in the description:

> The GameNotification class is never actually used -- only GameNotificationDto. Should I keep the GameNotification class?

I'm open to whatever for the same reasoning. I don't think we've ever actually just had the DTO version of the object, but it might make sense if we never use `GameNotification`.

src/test/java/com/rocketden/main/api/GameTests.java
zpChris: No it doesn't, because that converts a JSON string to an object, while this converts an object to a JSON string. I've tried putting the same Gson setup in the `toObjectLocalDateTime`, but it didn't seem to fix the issue -- so I'm stuck here. Any advice?

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: Ah thank you for catching this 👍 

Changed back!

src/test/java/com/rocketden/main/socket/GameSocketTests.java
zpChris: Makes sense, sounds good 👍

src/main/java/com/rocketden/main/util/Utility.java
zpChris: This is all fair, I'm mostly checking for things that we control ourselves in the code, so there's an argument to be made for "why check that". I figure the extra security can't hurt, but it may if it is excessive, so I'll leave any changes on this up to you -- I'm fine with leaving it or removing it.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: I updated this to use a for loop before this comment! 🎉  You can see my implementation, I used a map between the duration and content -- let me know if you think I should change anything there.

zpChris:
Re-requesting review @alankbi! Left some comments open above since further changes could be made based on those discussions -- lmk 👍

src/main/java/com/rocketden/main/game_object/GameNotification.java
alankbi: Same here - I'll leave it up to your discretion then

src/main/java/com/rocketden/main/service/GameManagementService.java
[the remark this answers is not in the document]
alankbi: Sounds good on 1 and 3 - for 2 yeah that's fine to leave it

src/main/java/com/rocketden/main/util/Utility.java
alankbi: Sounds good - I see no harm in leaving it for now so all good

src/test/java/com/rocketden/main/api/GameTests.java
alankbi: So right now it's just failing in the tests right?

src/test/java/com/rocketden/main/socket/GameSocketTests.java
```diff
@@ -147,6 +165,20 @@ public void socketReceivesMessageOnGameOver() throws Exception {
...
+    public void socketReceivesNotificationMessageTenSecondsLeft() throws Exception {
+        GameNotificationDto notificationDto = notificationBlockingQueue.poll(TIME_UNTIL_TEN_LEFT + 5, SECONDS);
+
+        assertNotNull(notificationDto);
+        
+        assertNull(notificationDto.getInitiator());
+        assertEquals(NotificationType.TIME_LEFT,
+            notificationDto.getNotificationType());
+        assertTrue(LocalDateTime.now().isAfter(notificationDto.getTime())
+            || LocalDateTime.now().minusSeconds((long) 1).isBefore(notificationDto.getTime()));
```
alankbi, line 178: Just curious what this is checking?

alankbi (approved):
Looks good 👍  thanks for those updates

src/test/java/com/rocketden/main/api/GameTests.java
alankbi: If so I'd be happy with removing the comment/ignoring it

zpChris: Ok -- I think I'll update the comment to be one line and not have a TODO, but still indicate that if time is replaced with `LocalDateTime.now()`, that gets a 400 error -- it'd be good to note that for the future.

src/main/java/com/rocketden/main/game_object/GameNotification.java
zpChris: Ok, I'll actually remove the `GameNotification` class, since it's not really doing anything.

src/test/java/com/rocketden/main/socket/GameSocketTests.java
zpChris: It's checking that the `notificationDto`'s time, since the time is set internally, is between 1 seconds before or after this moment. If you do an exact check the time's don't match, as the `notificationDto` was set up some number of milliseconds ago.

zpChris:
> So right now it's just failing in the tests right?

Actually, it's failing in Postman too -- since I can't figure out how to fix this, I'm just going to set the time on the backend. I'll work on making this change tomorrow @alankbi.

src/test/java/com/rocketden/main/api/GameTests.java
zpChris: Actually, it was failing in Postman too, so I've restructured this code and added a `GameNotificationRequest` that does not include the `time` variable; then, I set the `time = LocalDateTime.now()` on the backend. This is a fairly significant change, so although I think it looks good, I've re-requested review -- so lmk!

zpChris:
@alankbi Dealing with some test failures right now, I'll let you know on here when it's working again.

zpChris:
> @alankbi Dealing with some test failures right now, I'll let you know on here when it's working again.

I found the issue -- I had removed the ten seconds left notification, meaning the `socketReceivesNotificationMessageTenSecondsLeft` test was failing. I just added back that ten seconds left notification, because I still think it's useful as a last-second notification and it shortens the runtime for / simplifies testing.

alankbi (approved):

PR #96, opened 2021-01-19
Send socket message / end game once everybody solves the problem.

zpChris:
### List of Changes:

- Add a `boolean allSolved` field to the `Game` class and end the game once everybody has solved the problem.
- Check whether to update the `allSolved` variable on submission; when this variable first changes from `false` to `true`, send a socket notification.
- Test the mapping and socket message related to the `allSolved` variable.

### Notes:

- I use the `Mockito.doAnswer` method outlined [here](https://stackoverflow.com/questions/2276271/how-to-mock-void-methods-with-mockito) to simulate the method call and return of a mocked object (the `submitService.submitSolution` method).
- In the future, the game may not end if everybody has solved the problem, so we can design for different reactions based on the game type.

### Screencast and Images:

[Screencast of the Game End Once Everybody Solves the Problem](https://www.loom.com/share/b5ca00e144974f2d903d25329be47088)

**Image of the All Solved Socket Message Once Everybody Solves the Problem**
Image of the All Solved Socket Message Once Everybody Solves the Problem


src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -129,7 +129,17 @@ public SubmissionDto submitSolution(String roomId, SubmissionRequest request) {
...
+        SubmissionDto submissionDto = submitService.submitSolution(game, request);
+
+        /**
+         * Check allSolved if the user solved the problem and not everybody
+         * has solved the problem yet.
+         */
+        if (submissionDto != null && submissionDto.getNumCorrect().equals(submissionDto.getNumTestCases())&& Boolean.FALSE.equals(game.getAllSolved())) {
+            conditionalSolvedSocketMessage(game);
+        }
+
```
alankbi, line 141: Thoughts on removing this if statement check? Since the method handles the conditional update step as well (so this check might just be saving a few ms of time)

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
```diff
@@ -216,7 +242,105 @@ public void submitSolutionSuccess() {
...
+        gameService.submitSolution(ROOM_ID, request);
+
+        verify(submitService).submitSolution(eq(game), eq(request));
+
+        // Confirm that socket sent updated GameDto object.
+        verify(socketService).sendSocketUpdate(eq(GameMapper.toDto(game)));
+    }
+
+    @Test
+    public void sendAllSolvedNoSocketUpdate() {
```
alankbi, line 300: Possible idea: instead of enforcing the socket update only goes out once, is it possible if we just throw a `GameError.GAME_OVER` error if people try to submit after everyone is done/time is up?

frontend/src/views/Game.tsx
```diff
@@ -113,8 +114,7 @@ function GamePage() {
       setStateFromGame(updatedGame);
 
       // Check if end game.
-      // TODO: after deconstructing game, move this to useEffect on timeUp
-      if (updatedGame.gameTimer.timeUp) {
+      if (updatedGame.gameTimer.timeUp || updatedGame.allSolved) {
         history.push('/game/results', {
           game: updatedGame,
         });
```
alankbi, line 120: Suggestion: save `allSolved` and `timeUp` state variables in the `setStateFromGame` function and move to a separate useEffect block (that way the redirect will occur even if refreshing at the perfect time)

src/main/java/com/rocketden/main/game_object/Game.java
```diff
@@ -24,4 +24,7 @@
 
     private GameTimer gameTimer;
 
+    // Boolean to hold whether all users have solved the problem.
+    private Boolean allSolved = false;
```
alankbi, line 28: Thoughts on changing this to something like `isGameOver` and also setting it to true when the GameTimer `timeUp` is hit? That way we only need to check one variable for whether the game is over esp. if we add more end-game conditions in the future - although is there any situation where we would need to distinguish between game over by all solved or by time up?

alankbi (changes requested):
Just a couple idea discussions but otherwise great 👍

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: The method `conditionalSolvedSocketMessage` doesn't actually check for the conditions that `submissionDto` exists, the player got full credit, and not everybody has already solved the game -- these are the conditions I'm checking for here. I could move those checks within the `conditionalSolvedSocketMessage` method though -- should I do that?

src/main/java/com/rocketden/main/game_object/Game.java
zpChris: I think in the future, we may have different game modes -- for example, users compete to get the fastest solution, or one that uses the least memory -- which means we'll want to distinguish between `timeUp` and `allSolved`. It may also be useful for players viewing their past games, or for us curious about statistics, to see if the game ended by everybody solving the problem, or by the timer running out.

frontend/src/views/Game.tsx
zpChris: Good suggestion! Done ✅ 

I did not pass any state variables through `history.push`, as I saw you handled this logic in #97 and there's no use for it in this PR.

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
zpChris: I was initially going to implement this, but had another thought -- will this error actually ever exist, or will it just be `GameError.NOT_FOUND`? For example, if the game finishes, wouldn't the game be removed from the `currentGameMap`? Is there ever any intermediary state?

zpChris:
Re-requesting review, I left replies on the above discussions, so we can discuss those potential changes.

src/main/java/com/rocketden/main/game_object/Game.java
alankbi: Good points, I hadn't though of that - this works for me

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
alankbi: That's a good point (I think you commented about it in my other PR which I can implement that clean up stuff) - in that scenario though, would we need this test? Assuming the game is removed when all players finish submitting, then someone submitting again should throw a not found error

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: The main reason I brought it up was cause 1. the submission shouldn't ever return null based on our implementation 2. `conditionalSolvedSocketMessage` already loops through each player to check if each `player.solved` is true (tho I can see why that could be there if we wanted to save a bit on runtime to prevent running the check on all submissions), and 3. seems like the game error not found thing you brought up in a different comment

But yeah if you decide to keep it I think it'd be best to leave here that way you don't have to pass in the submission to that method

alankbi (approved):
A few unresolved comment chains are still left but otherwise looks good 👍

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -154,4 +164,26 @@ public void updateCode(String roomId, String userId, PlayerCode playerCode) {
         liveGameService.updateCode(game.getPlayers().get(userId), playerCode);
     }
 
+    /**
+     * If all players have solved the problem, update the game and send
+     * socket message indicating as such.
+     * (Depending on the game setting, this may or may not end the game).
+     */
+    public void conditionalSolvedSocketMessage(Game game) {
```
alankbi, line 172: On second thought, what do you think about moving this method to the submit service? I was thinking and after each submission, the submit service already sends a socket update, so this would cause 2 messages to be sent on every all solved game end. Maybe instead after the user code is judged/dummy submission is created, it runs a check for whether to set allSolved to true, and then proceeds as normal

Might also help to keep GameManagementService less cluttered

zpChris: I think it makes sense to move this method to submit service -- done! And yeah, that also makes sense not to actually send the socket message through this method, since it will be sent through the latter socket message -- I'll update this to reflect those changes.

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
zpChris: Not in our current setup -- the situation in which it would be useful is if the game did not end when everybody solves the problem, but rather when the time runs out if players are competing to make the best solution (fastest or least memory). Do you think I should keep this in for that reason, or no?

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: I can address both of these:
1. Yeah this is true, this is actually used for kind of a hacky purpose: in the test of this method in `GameManagementServiceTests.java`, the `submissionDto` actually is `null` if you don't mock it, so I used this check to prevent an error from occurring. I'm happy to instead mock the `submissionDto` return, just let me know.
2. That's a good point, because that only changes
3. This is fair, though this was my response to that comment:
> Not in our current setup -- the situation in which it would be useful is if the game did not end when everybody solves the problem, but rather when the time runs out if players are competing to make the best solution (fastest or least memory). Do you think I should keep this in for that reason, or no?

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Just to clarify a little bit - I was thinking the call to the conditionalSolvedSocketMessage (which could probably be renamed now that it's just checking allSolved) could happen perhaps at submit service line 44 (after submission is graded/determined to be success but before socket is sent) and then the method proceeds to send the socket update and return submission as before (meaning game management service 132-145 would no longer be necessary) - mainly to make submitService contain all submission-related logic and to keep the game management service only responsible for checking API errors beforehand and delegating responsibility to the correct class

This might also be related to the other two open comments: 

> Not in our current setup -- the situation in which it would be useful is if the game did not end when everybody solves the problem, but rather when the time runs out if players are competing to make the best solution (fastest or least memory). Do you think I should keep this in for that reason, or no?

After the recent changes would this still be necessary? Assuming we continue the pattern of sending a socket update after each user submission, then even if everyone solves the problem we'd still be sending/expecting a socket update after another person submits

> 1. Yeah this is true, this is actually used for kind of a hacky purpose: in the test of this method in GameManagementServiceTests.java, the submissionDto actually is null if you don't mock it, so I used this check to prevent an error from occurring. I'm happy to instead mock the submissionDto return, just let me know.
> 2. That's a good point, because that only changes

For 1, if the call to the conditional allSolved method is moved to the submit service method, would that remove the null mocking issue? Although I'm not sure exactly what the test covers or if this would require additional testing elsewhere to make up for it. For 2, moving that call could at least I guess simplify that if statement, since submit service will have logic/an if statement in the future to determine whether to set player.getSolved to true, and that call to the conditional check could be made at the same place

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
alankbi: Continued in the other chain

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Continued in the other chain

zpChris: I've addressed (1) as I've removed the `submissionDto != null` check, so now only the second two are applicable.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Ah, that makes sense -- OK, I added that code directly into the `submitSolution` method in `SubmitService.java`!

alankbi (approved):
Approved 👍

PR #97, opened 2021-01-23
Set up results page and play again functionality

alankbi:
### List of Changes:

* Set up results page to display list of players/winners
* Create play again functionality 
  * Creat REST endpoint for host to trigger play again
  * Connect frontend and backend to listen to the playAgain status through socket messages
  * Implement unit and integration tests
* Various refactorings to improve the codebase/get things properly working

### Notes:

* The play again functionality is triggered by the host and takes people back to the lobby where they can add new people or adjust the room settings. On the backend it reuses the room object rather than creating a new room. Lmk if we want to do this any differently or anything 

### Screencast and Images:

[Results page and play again functionality](https://www.loom.com/share/6c7f482f247044eea439080128ffade8)

Results page showing winners/players


alankbi:
[Display leaderboard on results page](https://trello.com/c/5raZbF1a/224-display-leaderboard-on-results-page)

frontend/src/views/Results.tsx
```diff
@@ -1,10 +1,99 @@
...
+          player={player}
+          place={index + 1}
+          isCurrentPlayer={currentUser?.userId === player.user.userId}
+          color={player.color}
+        />
+      ))}
+
+      {currentUser?.userId === room?.host.userId
+        ? <PrimaryButton onClick={playAgainAction}>Play Again?</PrimaryButton>
+        : <Text>Waiting for the host to choose whether to play again</Text>}
```
zpChris, line 95: nit: Add `...` to the end of the sentence.

frontend/src/views/Results.tsx
```diff
@@ -1,10 +1,99 @@
...
+  const playAgainAction = () => {
+    setError('');
+    setLoading(true);
+
+    playAgain(room!.roomId, { initiator: currentUser! })
+      .catch((err) => {
+        setLoading(false);
+        setError(err.message);
+      });
+  };
```
zpChris, line 77: This would require some redesign, but perhaps now or in the future, the results page _is_ the lobby page in a sense? Like maybe it is a modal on top of the lobby page? My reasoning is that if the host unexpectedly leaves, the rest of the players might want to continue the game.

Thinking on this further though, I think the current design is fine for the MVP, but we might want to revamp it in the future.

frontend/src/views/Results.tsx
```diff
@@ -1,10 +1,99 @@
...
+        <PlayerResultsCard
+          player={player}
+          place={index + 1}
+          isCurrentPlayer={currentUser?.userId === player.user.userId}
+          color={player.color}
+        />
+      ))}
+
+      {currentUser?.userId === room?.host.userId
+        ? <PrimaryButton onClick={playAgainAction}>Play Again?</PrimaryButton>
```
zpChris, line 94: nit:
```suggestion
        ? <PrimaryButton onClick={playAgainAction}>Play Again</PrimaryButton>
```

src/main/java/com/rocketden/main/exception/GameError.java
```diff
@@ -10,10 +10,11 @@
...
     BAD_LANGUAGE(HttpStatus.BAD_REQUEST, "An invalid language was chosen."),
     BAD_SETTING(HttpStatus.BAD_REQUEST, "An invalid game setting was provided."),
-    USER_NOT_IN_GAME(HttpStatus.BAD_REQUEST, "The requested action could not be completed because the user is not in the game."),
-    INVALID_PERMISSIONS(HttpStatus.FORBIDDEN, "You do not have permission to perform this action."),
     EMPTY_FIELD(HttpStatus.BAD_REQUEST, "Please ensure a value is provided for each required field."),
-    NOT_FOUND(HttpStatus.NOT_FOUND, "A game could not be found with the given id.");
+    GAME_NOT_OVER(HttpStatus.FORBIDDEN, "You may not perform this action because the game is not over."),
+    INVALID_PERMISSIONS(HttpStatus.FORBIDDEN, "You do not have permission to perform this action."),
+    NOT_FOUND(HttpStatus.NOT_FOUND, "A game could not be found with the given id."),
+    USER_NOT_IN_GAME(HttpStatus.BAD_REQUEST, "The requested action could not be completed because the user is not in the game.");
```
zpChris, line 17: Did you change these to arrange alphabetically? Should I keep that system on future enums?

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -95,6 +99,33 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
         return roomDto;
     }
 
+    public RoomDto playAgain(String roomId, PlayAgainRequest request) {
+        Game game = getGameFromRoomId(roomId);
+
+        if (game.getGameTimer() == null || !game.getGameTimer().isTimeUp()) {
+            throw new ApiException(GameError.GAME_NOT_OVER);
+        }
```
zpChris, line 107: When we add #96, will we add `allSolved` to the if statement?

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -95,6 +99,33 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
...
+
+        Room room = game.getRoom();
+        User initiator = UserMapper.toEntity(request.getInitiator());
+        if (!room.getHost().equals(initiator)) {
+            throw new ApiException(GameError.INVALID_PERMISSIONS);
+        }
+
+        // Set all users to be disconnected
+        room.getUsers().forEach((user) -> user.setSessionId(null));
```
zpChris, line 116: Does this actually disconnect the users on the frontend? If not, what is this used for?

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
```diff
@@ -263,6 +264,94 @@ public void submitSolutionEmptyField() {
...
+        Thread.sleep(1000);
+
+        UserDto initiator = new UserDto();
+        initiator.setNickname(NICKNAME_2);
+        PlayAgainRequest request = new PlayAgainRequest();
+        request.setInitiator(initiator);
+
+        ApiException exception = assertThrows(ApiException.class, () -> gameService.playAgain(ROOM_ID, request));
+        assertEquals(GameError.INVALID_PERMISSIONS, exception.getError());
+
```
zpChris, line 329: really minor nit: Remove line.

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
```diff
@@ -263,6 +264,94 @@ public void submitSolutionEmptyField() {
...
+        room.setActive(true);
+        room.addUser(host);
+        room.setDuration(1L);
+
+        StartGameRequest request = new StartGameRequest();
+        request.setInitiator(UserMapper.toDto(host));
+
+        Mockito.doReturn(room).when(repository).findRoomByRoomId(ROOM_ID);
+        gameService.startGame(ROOM_ID, request);
+        Thread.sleep(1000);
```
zpChris, line 287: nit: Replace `Thread.sleep` with a different testing implementation (in all instances, there is another below for example):

Screen Shot 2021-01-23 at 1 40 57 PM

For the `verify` check, I believe you can do `verify(socketService, timeout(1000)).sendSocketUpdate(eq(gameDto));` -- I do this on `EndGameTimerTaskTests.java`.

frontend/src/views/Results.tsx
```diff
@@ -1,10 +1,99 @@
...
+                user: location.state.currentUser,
+                roomId: updatedGame.room.roomId,
+              });
+            })
+            .catch((err) => setError(err.message));
+        }
+      };
+
+      subscribe(routes(location.state.game.room.roomId).subscribe_game, subscribeCallback)
+        .catch((err) => setError(err.message));
```
zpChris, line 60: Before subscribing the users to the "subscribe game" route again, unsubscribe them from the current "subscribe game" route, and from the "notifications" route. This can be done by passing the subscription through to the `Results` page in location / history, then calling `unsubscribe()` on that object.

frontend/src/views/Results.tsx
```diff
@@ -1,10 +1,99 @@
...
+
+        // Disconnect users from socket and then redirect them to the lobby page
+        if (updatedGame.playAgain) {
+          disconnect()
+            .then(() => {
+              history.replace(`/game/lobby?room=${updatedGame.room.roomId}`, {
+                user: location.state.currentUser,
+                roomId: updatedGame.room.roomId,
+              });
+            })
```
zpChris, line 54: What exactly is happening here -- is a new game being constructed and added to the `currentGameMap`, or does the object just consist of the current game object. It seems like it is the latter, and in that case, how is the game "refreshed" so the information from the previous game does not stay there?

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -95,6 +99,33 @@ public RoomDto startGame(String roomId, StartGameRequest request) {
         return roomDto;
     }
 
+    public RoomDto playAgain(String roomId, PlayAgainRequest request) {
```
zpChris, line 102: At some point, we should remove the `game` from `currentGameMap`. I thought maybe this would be done after the game ends, or maybe it could be done after a set amount of time or when all the users leave. Thoughts?

zpChris (changes requested):
Requesting changes because there are some design discussions around the PR I wanted to bring up -- but overall, this looks great! I love the play again functionality, great job 😃

frontend/src/api/Socket.ts
```diff
@@ -84,16 +84,16 @@ export const send = (sendUrl: string, headers?: {}, body?: string):
...
     if (stompClient && stompClient.connected) {
       // Stomp client is already set from last connection.
       stompClient.disconnect(() => {
         resolve();
       });
     } else {
-      reject(errorHandler('The socket is not connected.'));
+      resolve();
     }
   });
```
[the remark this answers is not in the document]
alankbi: Sure - do you mean like to connect them if they aren't connected already? In which case that might also make the connect function redundant 

As for this place, the main reason I did this change was so that disconnect would basically ensure you're not connected to the socket (regardless of whether you were connect or not in the first place)

frontend/src/views/Game.tsx
```diff
@@ -106,19 +92,21 @@ function GamePage() {
     }
   }, [currentUser]);
 
+  // Check if game is over or not (TODO: include check for all solved)
+  useEffect(() => {
+    if (gameTimer?.timeUp) {
+      history.replace('/game/results', {
+        game,
```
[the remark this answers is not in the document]
alankbi: Good question - there might be stuff in the future, although as of now I can't think of any results page specific variables beyond the play again variable (which might not even be necessary if we change the results page to basically be a lobby page)

frontend/src/views/Results.tsx
alankbi: That's a good idea - if it was a modal-type thing on the lobby page, I could see it being pretty simple to change the playAgain function to simply be a onGameEnd function called within the `Game.tsx` file, which redirects you back to the lobby page with like a game/results object. Then on the lobby page it can check for that in `location` and if so display the modal. 

One question about this - if it's a modal with like a close button, what happens if people close it but want to view it again later? Or if they have the modal open and then the host clicks start game, would that catch them off guard (they weren't expecting a new game to begin)?

src/main/java/com/rocketden/main/exception/GameError.java
alankbi: Yeah - I figured just whenever we get around to adding/modifying the errors, we can just reorder them to be in alphabetical order (or maybe order them by HTTP status might be better, either way)

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Yeah probably (maybe we could add an isGameOver helper method or even a gameOver variable to Game.java assuming in the future the end game criteria might grow more complicated)

Another note is after #96 is merged I'll also update the Game.tsx file to pass in the correct objects to the results page

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: This is basically to ensure the list of active and inactive users are correct and re-trigger the WebSocketConnectionEvents code - without it, I think everyone would show up to the lobby page as connected even if they've already closed the browser

frontend/src/views/Results.tsx
alankbi: Sounds good - instead of passing those over location though, I just unsubscribed them on the game page before redirecting (I also changed the boolean `xxxSubscribed` to subscription objects using a null check to determine the same result)

frontend/src/views/Results.tsx
alankbi: It's only passing roomId and the currentUser object to the lobby page (location requirements), so none of the actual game info is carried over (and in fact I think once the other PR is merged in I'll implement the endGame functionality to delete the old game - tho in the future endGame might be to save to the database). And then once on the lobby page it behaves as normal (fetching the RoomDto with the same settings as before, etc.)

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Yeah sounds good - I'll wait for the other PR to be merged first, that way I can pull those changes and add the calls to a new endGame method at the correct places.

alankbi: Actually, this might present some issues - the play again method requires the game to still exist by setting the play again variable and sending a game socket message, so if games were removed, it likely couldn't be immediately after the game ends. Thoughts? We could do delete after a certain amount of time, although tbh that doesn't feel desirable to be too reliant on timers and stuff

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
alankbi: Sounds good

frontend/src/api/Socket.ts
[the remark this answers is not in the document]
zpChris: Oh I meant that you replaced the `disconnect` function's `reject(errorHandler('The socket is not connected.'));` with `resolve();`, whereas `subscribe` and `send` still have the `reject` lines; so I'm thinking either change all to `resolve();` or leave everything as-is.

frontend/src/views/Results.tsx
zpChris: That's a good question -- we should think about design by checking out other sites, since I'm not sure how this would work off the top of my head. We could look at other sites like [colonist.io](http://colonist.io/), [binarysearch.com](https://binarysearch.com/), and [Clash of Code](https://www.codingame.com/multiplayer/clashofcode) for inspiration. One idea is that even if you close the modal, you can re-open it through some tab or something; then, every time you "Play Again", that previous game history is added to the list that you can view on reopening the modal.

We maybe should talk more about this sometime, either on a call or just when playing basketball.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Maybe we could discuss different ideas on a call, but there could be a countdown for rejoining the game (visible to the users, and if the game is not restarted, then you can't restart it?), or a new Game could be created every time (we'd have to figure out another way to restart, perhaps by creating a different subscription especially for the "restart and create new game" purpose). You could potentially make a new "create room" method that allows you to specify the room ID by making it the previous one, and the game can be stored in the database under a different ID (to prevent duplicates, we'd likely want to do this anyway).

zpChris: That's kind of just a braindump of ideas, we can talk through this more another time.

frontend/src/api/Socket.ts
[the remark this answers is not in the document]
alankbi: The reason I went with this was cause calling disconnect when already disconnected doesn't change anything (and thus I didn't think it warranted an error message) whereas calling subscribe/send when not connected would definitely cause an error to come up - however, the disconnect reject error that originally prompted this change I think was fixed elsewhere so I'd be happy to change them all back to reject if we wanted to prioritize consistency

frontend/src/views/Results.tsx
alankbi: Sounds good -  I added all this info to a card in the backlog so I'll resolve this convo for now (since it's probably not ready for this PR yet)

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Sounds good - given that this might also be dependent on what we do with the results page, I added another card for it in the backlog with the thought that it could be dealt with in its own PR (in addition to all the saving game info to the database stuff which could be relevant to an endGame method)

frontend/src/api/Socket.ts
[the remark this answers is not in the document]
zpChris: Yeah, I think that'd make sense to change it back, if that's alright 👍

zpChris (approved):
LGTM! Thanks for making the relevant changes 👍 

I do think a future task would be remove the game from the `currentGameMap` (and then, at some point, the room from the database as well). This could be done by simply requiring players to restart the game within some amount of time, or closing the game once everybody leaves the page (disconnects).

One more thing -- I want to note that if you refresh on the "Results" page, it disconnects you from the socket. However, a lot of work has been on this PR, so I've added a [new card in Trello for this](https://trello.com/c/n6FZB2w3/248-refresh-on-results-page-causes-socket-disconnect). It also doesn't seem to update the host correctly in this case because they never connect in the first place, which is another issue.

That is all for a future PR though, this looks great!

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
```diff
@@ -411,6 +411,95 @@ public void submitSolutionEmptyField() {
...
+        initiator.setNickname(NICKNAME_2);
+        PlayAgainRequest request = new PlayAgainRequest();
+        request.setInitiator(initiator);
+
+        ApiException exception = assertThrows(ApiException.class, () -> gameService.playAgain(ROOM_ID, request));
+        assertEquals(GameError.INVALID_PERMISSIONS, exception.getError());
+    }
+
+    @Test
+    public void playAgainGameNotOver() {
```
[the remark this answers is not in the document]
alankbi: The first call, `getGameFromRoomId`, should trigger a game not found error (since the game hasn't started yet) - so yeah nothing bad should happen I wouldn't think

PR #100, opened 2021-01-26
Create pipeline for calling the tester service

alankbi:
### List of Changes:

* Add method in SubmitService that sends a POST request to the tester service
* Create URL and debug mode config options to support live and not-yet-ready stages (we're not ready yet)
* Write tests to cover new endpoint to best of ability 
* Create associated Request/Response files for sending/receiving payload from tester

### Notes:

* The tester.url and tester.debugMode parameters in the application.properties files (different for main and test environments) allow us to specify the URL and also debugMode 
  * Debug mode specifies whether to return the default dummy submission (which we currently always return) if the call to tester fails (which it currently always does)
* The URL and debug mode options do not work with mocked test classes (MockMVC works fine), so there are workaround method to those (which are admittedly ugly design imo, so feel free to leave suggestions for better solutions)
* The test cases are very weak for the actual tester call itself, since I couldn't get it to mock correctly so it's basically just letting it fail and asserting that the dummy response comes back - we might need improvements later
* Once the service is live, we'll probably need logic to determine if the response is like a 40X bad request/etc. response, since I think that won't trigger an exception on main but instead we'll have to check the status and convert the JSON to the proper response object (right now, it just defaults to the TesterResponse/RunDto object)


alankbi:
[Call tester service on submit solution to judge user code](https://trello.com/c/LoRcoGAv/244-call-tester-service-on-submit-solution-to-judge-user-code)

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -58,4 +105,58 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
+            submission.setNumCorrect(request.getProblem().getTestCases().size());
+            submission.setNumTestCases(request.getProblem().getTestCases().size());
+            submission.setPlayerCode(new PlayerCode(request.getCode(), request.getLanguage()));
+
+            return submission;
+        } catch (Exception e) {
+            // If in debug mode (tester is unavailable), return a dummy submission
+            if (getDebugMode()) {
+                return getDummySubmission(request);
+            }
```
zpChris, line 134: Why not put this if statement at the start of the method, outside of the try-catch block? Is the testerService still supposed to return something if you are in debug mode?

src/main/java/com/rocketden/main/dto/game/TesterResponse.java
```diff
@@ -0,0 +1,12 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+// This class is mapped directly to the tester service's RunDto class.
+public class TesterResponse {
+    private Boolean status;
+    private String output;
+}
```
zpChris, line 12: I want to note that I think the `TesterResponse` will end up being different than this because I'm predicting that we do the process of parsing the `output` and gathering the results for each test, errors, and # correct -- and passing that through here. So I can envision us essentially returning a more complete Submission object, which means the TODO shown [here](https://github.com/rocketden/main/pull/100/files#diff-df30ccd300c4cf9576689b18935e50bcc164e2387d5485a5c3963827c86f5532R123) in my mind is really more relevant for the tester repo.

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -58,4 +105,58 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
+            HttpResponse response = httpClient.execute(post);
+            String jsonResponse = EntityUtils.toString(response.getEntity());
+
+            TesterResponse testerResponse = getResponseFromJson(jsonResponse);
+
+            // TODO: logic to convert TesterResponse into Submission object
+            Submission submission = new Submission();
+            submission.setNumCorrect(request.getProblem().getTestCases().size());
+            submission.setNumTestCases(request.getProblem().getTestCases().size());
+            submission.setPlayerCode(new PlayerCode(request.getCode(), request.getLanguage()));
```
zpChris, line 127: So to be clear, right now, we are getting the results from the tester service but not doing anything with them because that process needs more work?

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -58,4 +105,58 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
+            HttpPost post = new HttpPost(getTesterUrl());
+
+            StringEntity stringEntity = new StringEntity(gson.toJson(request));
+            post.setEntity(stringEntity);
+            post.setHeader("Content-type", "application/json");
+
+            HttpResponse response = httpClient.execute(post);
+            String jsonResponse = EntityUtils.toString(response.getEntity());
+
+            TesterResponse testerResponse = getResponseFromJson(jsonResponse);
```
zpChris, line 121: How long did this process seem to take on your computer?

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -58,4 +105,58 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
+            }
+
+            throw new ApiException(GameError.TESTER_ERROR);
+        }
+    }
+
+    // Is null in certain testing environments; if so, return a default value
+    private String getTesterUrl() {
+        if (testerUrl == null) {
+            return "http://localhost:8080";
```
zpChris, line 143: Will this be accurate, or is this just a default URL that may or may not connect to the tester service? If it's the latter, I may favor returning null and doing a check? I'm not sure -- what do you think.

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -58,4 +105,58 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
+        }
+        return testerUrl;
+    }
+
+    // Is null in certain testing environments; if so, return true by default
```
zpChris, line 148: When is this value null?

src/main/java/com/rocketden/main/service/SubmitService.java
```diff
@@ -58,4 +105,58 @@ public SubmissionDto submitSolution(Game game, SubmissionRequest request) {
...
+        return debugMode == null || debugMode;
+    }
+
+    // Method that can be mocked to not return an error for SubmitServiceTests
+    protected TesterResponse getResponseFromJson(String json) {
+        return gson.fromJson(json, TesterResponse.class);
+    }
+
+    // This method should only be called for testing purposes
+    protected void toggleDebugModeForTesting(boolean debugMode) {
```
zpChris, line 159: nit:
```suggestion
    protected void setDebugModeForTesting(boolean debugMode) {
```
(I think it'd be "toggle" if it switched whatever value the variable was currently set at.)

zpChris (approved):
LGTM! Just left a good amount of nits and general questions -- before you merge, can you resolve those? I'm approving instead of commenting because I don't think the discussions will impact this PR, but just want to verify the answers either way.

src/main/java/com/rocketden/main/exception/GameError.java
```diff
@@ -10,12 +10,13 @@
...
     BAD_SETTING(HttpStatus.BAD_REQUEST, "An invalid game setting was provided."),
-    NOT_FOUND(HttpStatus.NOT_FOUND, "A game could not be found with the given id."),
     USER_NOT_IN_GAME(HttpStatus.BAD_REQUEST, "The requested action could not be completed because the user is not in the game."),
     NOTIFICATION_REQUIRES_INITIATOR(HttpStatus.BAD_REQUEST, "The request action could not be completed because an initiator is required for this notification."),
     NOTIFICATION_REQUIRES_CONTENT(HttpStatus.BAD_REQUEST, "The request action could not be completed because content is required for this notification."),
     INVALID_PERMISSIONS(HttpStatus.FORBIDDEN, "You do not have permission to perform this action."),
-    EMPTY_FIELD(HttpStatus.BAD_REQUEST, "Please ensure a value is provided for each required field.");
+    EMPTY_FIELD(HttpStatus.BAD_REQUEST, "Please ensure a value is provided for each required field."),
+    NOT_FOUND(HttpStatus.NOT_FOUND, "A game could not be found with the given id."),
+    TESTER_ERROR(HttpStatus.INTERNAL_SERVER_ERROR, "An internal error occurred connecting to the tester service.");
```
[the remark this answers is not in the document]
alankbi: Personally, I'd be okay to just have this one tester error on the main side - if an internal error occurs, the message prob isn't relevant to the user - only to us for debug purposes, in which case we'd probably want to debug the error on the tester repo anyways where it originated

Might also help keep the amount of copied/mirrored code smaller between main and tester

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: My main thinking behind it was that the dummy submission would be an optional fallback for the actual request. Obviously rn the request is not yet ready so it always fails/falls back - if you think this might be better suited to be put at the start, lmk and I can do that

src/main/java/com/rocketden/main/dto/game/TesterResponse.java
alankbi: Makes sense yeah - I can imagine most of that will be done on tester but there might still be a few missing pieces (potentially isSolved) to clean up here (also having a separate response class might be useful for the same reasoning behind our current DTO/model structure

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: Yeah - I think kind of tied to your other comment, once the tester service returns a more complete TesterResponse, then we would actually be able to use it

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: I didn't pay too much attention but definitely less than a few seconds (and less than how long the POST request on the tester service takes to finish with all the docker stuff)

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: This is just returning a default URL (that doesn't work, since for our tests we won't have access to the tester service unless we link it with the live site which probably isn't good practice haha) for testing purposes - there's a few tests in which it can't read in the value for testerUrl, so this is just checking that and setting it to the same value as in the test application.properties

About the null check - could you clarify what that might look like?

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: Only when running tests that have mocked the submit service class (for now that's just SubmitServiceTests I think)

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: Sounds good - I made a separate commit for this (since I also had to refactor the usages of the method too)

src/main/java/com/rocketden/main/service/SubmitService.java
zpChris: Yeah I think it makes more sense for it to be at the start; if you're testing the actual request, you'd want it to show the error to troubleshoot rather than a dummy submission imo.

src/main/java/com/rocketden/main/service/SubmitService.java
zpChris: I guess not actually returning null, but maybe throwing an error as the application has no chance of working in those testing environments (if I understand correctly). I don't have a huge preference on this though.

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: Sounds good - implemented

src/main/java/com/rocketden/main/service/SubmitService.java
alankbi: I'll probably leave it as is then - it's a little gimmicky but I think those unit tests are still valid/useful to have

alankbi:
Re-requesting review since there were a decent number of potential changes 

Also, it looks like GitHub Actions is down right now so might wanna rerun the tests when it's back up

zpChris (approved):
LGTM!

PR #101, opened 2021-01-29
Create Backend Setup for Creating the Problem Default Code.

zpChris:
### List of Changes:

- Add the fields to store input and output types to the `Problem` class.
- Create the `ProblemIOType` and `ProblemInput` objects, as moved from the tester repository.
- Update the functions that create and map `Problem` objects.
- Update the current tests to reflect the change in creating problems.
- Add tests to confirm that these fields work as expected.

### Notes:

**Issue with `LazyInitializationException`** 

tl;dr: We may want to redo how we fetch some entities.

Introducing a separate entity to the `Problem` class (`ProblemInput`) caused a `LazyInitializationException`, as [only one `FetchType.EAGER` List](https://stackoverflow.com/questions/30122786/can-a-jpa-entity-have-multiple-onetomany-associations) entity is allowed in a single entity. This can be [sidestepped by adding `@IndexColumn(name="INDEX_COL")`](https://blog.eyallupu.com/2010/06/hibernate-exception-simultaneously.html), but that is deprecated (and thus likely not recommended). I also tried using [`Hibernate.initialize`](https://howtodoinjava.com/hibernate/use-hibernate-initialize-to-initialize-proxycollection/), but this didn't seem to work.

The other solution is to add [`spring.jpa.properties.hibernate.enable_lazy_load_no_trans=true`](https://stackoverflow.com/questions/36583185/spring-data-jpa-could-not-initialize-proxy-no-session-with-methods-marke) to `application.properties`; _**this is what I ended up doing**_. However, this (as well as open-session-in-view, which we use) seems to be an anti-pattern: see [[1]](https://vladmihalcea.com/the-hibernate-enable_lazy_load_no_trans-anti-pattern/) [[2]](https://vladmihalcea.com/the-open-session-in-view-anti-pattern/).

This same blogger has a recommendation for [how to best handle `LazyInitializationException`](https://vladmihalcea.com/the-best-way-to-handle-the-lazyinitializationexception/). It seems `FetchType.EAGER` may [also be an anti-pattern](https://stackoverflow.com/questions/3238094/fetch-multiple-onetomany-relationships-hibernate-jpa), so I added a [card to the Trello board with all of this info](https://trello.com/c/bqaSq8db/252-issue-with-lazyinitializationexception).

-----

- I parameterized a test in `UserTests.java` to reduce duplicate code.
- Since the `problemInputs` and `outputType` are inherent to the problem, they are set when you first make the problem, unlike the test cases.
- The Problem ID field is now fairly different -- the current example in Postman is `"3647fc2e-2bfd-467a-8309-150809ef6dcc"`.
- Since there is already a good amount of code in this PR, I will introduce the code that creates the base problem default code in the next PR.

### Screencast and Images:

There are no changes to the UI (yet, that will happen next PR), thus no screencasts or images are attached.


src/test/java/com/rocketden/main/util/UtilityTestMethods.java
```diff
@@ -33,9 +37,15 @@ public static String convertObjectToJsonString(Object o) {
...
-                    DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSSSS"));
+            public LocalDateTime deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+                    DateTimeFormatter formatter = new DateTimeFormatterBuilder()
+                        .appendPattern(DATE_TIME_FORMAT_PATTERN)
+                        // optional decimal point followed by 1 to 6 digits
+                        .optionalStart()
+                        .appendFraction(ChronoField.MICRO_OF_SECOND, 1, 6, true)
+                        .optionalEnd()
+                        .toFormatter(); 
+                    return LocalDateTime.parse(json.getAsString(), formatter);
```
alankbi, line 48: Does this fix those constant test failiures? If so that's awesome - how does this work?

src/main/java/com/rocketden/main/model/problem/Problem.java
```diff
@@ -52,4 +60,9 @@ public void addTestCase(ProblemTestCase testCase) {
     public boolean removeTestCase(ProblemTestCase testCase) {
         return testCases.remove(testCase);
     }
+
+    public void addProblemInput(ProblemInput problemInput) {
+        problemInputs.add(problemInput);
+        problemInput.setProblem(this);
+    }
```
alankbi, line 67: nit: This is pretty small but could you add a quick test for this in ProblemEntityTests?

alankbi (approved):
Looks great, approved! Only like 1-2 minor nits

alankbi:
[Create backend setup for creating the problem default code](https://trello.com/c/7b7cjtgT/254-create-backend-setup-for-creating-the-problem-default-code)

src/main/java/com/rocketden/main/model/problem/ProblemInput.java
```diff
@@ -0,0 +1,37 @@
...
+@EqualsAndHashCode
+public class ProblemInput {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Integer id;
+
+    private String name;
+    private ProblemIOType type;
+
+    @ManyToOne(fetch = FetchType.EAGER)
```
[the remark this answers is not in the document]
zpChris: Ah -- the value here doesn't seem to make as much of an impact. From one of the links I read, I think you can put these as `EAGER`, they I don't remember where I found that.

The value that matters is in the `Problem` class, currently the `FetchType` values are `LAZY`, and if they are both `EAGER` than an error arises.

src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
```diff
@@ -10,4 +15,6 @@
     private String name;
     private String description;
     private ProblemDifficulty difficulty;
+    private List<ProblemInput> problemInputs;
```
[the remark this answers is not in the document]
zpChris: Done!

src/test/java/com/rocketden/main/util/UtilityTestMethods.java
zpChris: Yes I think it does (I should've figured this out earlier tbh), here's how it works:

The Date used to be parsed by the following format: `yyyy-MM-dd'T'HH:mm:ss.SSSSSS". It is notable that there are exactly six microseconds there -- if there are any less, than the test fails because it cannot parse it. So I believe what was happening (I could be wrong, but something like this) was occasionally the last `S` would be a `0`, which means the format was only five `S`, and thus the test failed.

This setup correctly parses correctly regardless of how many microseconds there are, so it should parse as expected now.

src/main/java/com/rocketden/main/model/problem/Problem.java
zpChris: Sure, no problem -- done!

PR #102, opened 2021-01-30
Generate default code on the backend.

zpChris:
### List of Changes:

- Add an interface and generator implementing classes (Java, Python) for writing default code.
- Create the `Language` enum and relevant variables, from the tester repository.
- Write tests in single file for the `writeDefaultCode` implementing methods.

### Notes:

- This is the second of a three-part PR series to write the default code on the frontend. This PR implements the functionality on the backend.
- I struggled to implement the tests for `problemService.getDefaultCode`, as I could not mock the list of `DefaultCodeGeneratorService` implementing classes -- @alankbi, if you know how to do this, let me know!
- I tried making the tests for `writeDefaultCode` a single parameterized test, but struggled because I could not pass in the list of service classes into the `@ValueSource` annotation.
- We can add the other default code languages as we implement that here and on the tester repository.

### Screencast and Images:

There are no changes to the UI, and thus no screencasts or images are attached.


zpChris:
@alankbi, this is in the Notes section, but I wanted to place it here as well for visibility:

> I struggled to implement the tests for problemService.getDefaultCode, as I could not mock the list of DefaultCodeGeneratorService implementing classes -- @alankbi, if you know how to do this, let me know!

zpChris:
[Implement generator classes on the backend](https://trello.com/c/E2TvAsRS/255-implement-generator-classes-on-the-backend)

src/main/java/com/rocketden/main/model/Language.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.commons.lang.WordUtils;
+
+import lombok.Getter;
+
+@Getter
+public enum Language {
+
+    PYTHON,
+    RUBY,
+    SWIFT,
```
alankbi, line 16: There's already a CodeLanguage enum class - could you add this driverGeneratorName logic to that class and use that one instead? Or delete the old one and refactor the codebase to use this class, whichever is easier

src/main/java/com/rocketden/main/service/DefaultCodeGeneratorService.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.List;
+
+import com.rocketden.main.model.Language;
+import com.rocketden.main.model.problem.ProblemIOType;
+import com.rocketden.main.model.problem.ProblemInput;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public interface DefaultCodeGeneratorService {
```
alankbi, line 12: nit: Move this into the generators folder

alankbi (commented):
Looks good - only left like 1 legitimate nit/suggestion haha

Before approving though, I wanted to address this:

>I struggled to implement the tests for problemService.getDefaultCode, as I could not mock the list of DefaultCodeGeneratorService implementing classes -- @alankbi, if you know how to do this, let me know!

I haven't looked too much into it but does something like [this](https://stackoverflow.com/questions/33302340/how-to-mock-a-autowired-list-of-spring-beans) help (altho it's possible you've seen this link already)? 

Since you're unit testing the actual Java/Python generator services themselves though, I think it might be ok not to have too many tests covering that method - however, if you can, I would add an API test (with MockMVC) covering that endpoint as a whole (which shouldn't require you to mock anything)

src/main/java/com/rocketden/main/model/Language.java
zpChris: Whoops, that was silly of me, I forgot about CodeLanguage; thanks for catching this, I'll use that instead.

zpChris: Done! I accidentally changed on #103 at first instead, but the change is on both now.

src/main/java/com/rocketden/main/service/DefaultCodeGeneratorService.java
zpChris: Done ✅

zpChris:
> Looks good - only left like 1 legitimate nit/suggestion haha
> 
> Before approving though, I wanted to address this:
> 
> > I struggled to implement the tests for problemService.getDefaultCode, as I could not mock the list of DefaultCodeGeneratorService implementing classes -- @alankbi, if you know how to do this, let me know!
> 
> I haven't looked too much into it but does something like [this](https://stackoverflow.com/questions/33302340/how-to-mock-a-autowired-list-of-spring-beans) help (altho it's possible you've seen this link already)?
> 
> Since you're unit testing the actual Java/Python generator services themselves though, I think it might be ok not to have too many tests covering that method - however, if you can, I would add an API test (with MockMVC) covering that endpoint as a whole (which shouldn't require you to mock anything)

Ok sounds good, I'll work on this now -- thanks for adding that link too, I think I likely will write a test for this (but won't spend _too_ much time on it).

zpChris:
Done! Changes made ✅  Re-requested review since there were a decent number of changes.

alankbi (approved):
Approved 👍

PR #103, opened 2021-01-31
Add the default code setup to the frontend.

zpChris:
### List of Changes:

- Add the default code to the frontend (implementing for Java and Python).
- Add API call for getting the default code associated with a problem.

### Notes:

- Not all of the languages on the backend have syntax highlighting; the [supported list](https://github.com/suren-atoyan/monaco-react#onvalidate) does not include Swift, Ruby, and Bash.
- I designed the API call and function so that it could handle multiple problems (as the default code will depend on the problem). Eventually, if we have multiple problems in the same game, once you change problems the code should change as well.
- I tried to use a `Map<Language, string>` on the frontend, but had an issue with using the `.get(...)` function. However, I think the `DefaultCodeType` type is a good solution -- just wanted to note a blocker I ran into.
- If you choose a language with unimplemented default code, it throws an error on the frontend.

### Screencast and Images:

[Screencast of Get Default Code on Game Page](https://www.loom.com/share/143a426ab87846a4b70f783553749152)

**Image of the Default Code in Editor**

Get Default Code on Game Page


zpChris:
[Set up the default code on the frontend](https://trello.com/c/JN6tj6f9/256-set-up-the-default-code-on-the-frontend)

alankbi:
Will review this tomorrow

frontend/src/views/Game.tsx
```diff
@@ -84,6 +85,22 @@ function GamePage() {
...
+        promises.push(getDefaultCodeMap(problem.problemId));
+      }
+    });
+
+    // Get the result of promises and set the default code list.
+    Promise.all(promises).then((result) => {
+      setDefaultCodeList(result);
+    }).catch((err) => {
+      setError(err.message);
+    });
```
alankbi, line 101: If I'm understanding correctly, this waits for all the default code maps for each language to be returned and then the result object is a list of the code maps?

frontend/src/components/game/Editor.tsx
```diff
@@ -70,34 +30,41 @@ function ResizableMonacoEditor(props: EditorProps) {
     });
   };
 
-  const handleLanguageChange = (language: string) => {
+  const handleLanguageChange = (language: Language) => {
     // Save the code for this language
-    languages[currentLanguage].defaultCode = codeEditor!.getValue();
+    if (codeMap != null && codeEditor != null) {
+      codeMap[currentLanguage] = codeEditor.getValue();
+      codeEditor.setValue(codeMap[language]);
```
alankbi, line 37: Note: if language is not in the map, I get this error in the console: 

Screenshot

However, the default behavior is that the code in the editor simply stays as is, so think it's fine for now, just something to note

frontend/src/api/Language.ts
```diff
@@ -0,0 +1,38 @@
...
+  Rust = 'RUST',
+  Bash = 'BASH',
+}
+
+export const fromString = (key: string): Language => Language[key as keyof typeof Language];
+
+export const languageToEditorLanguage = (key: Language): string => {
+  // Swift, Rust, and Bash do not have corresponding editor languages.
+  switch (key) {
+    case Language.Python:
```
alankbi, line 19: nit: Comment out the languages that we aren't currently supporting

alankbi (approved):
No major issues! One last remaining comment that I will mention:

When you type code in the editor, add new lines, click tab/enter, etc., it defaults to using spaces, but the generated code uses tabs, which causes a mixture of the two. This isn't too significant right now, but worth potentially standardizing at some point to make things nicer for the user

alankbi:
Screenshot

👀

zpChris:
> No major issues! One last remaining comment that I will mention:
> 
> When you type code in the editor, add new lines, click tab/enter, etc., it defaults to using spaces, but the generated code uses tabs, which causes a mixture of the two. This isn't too significant right now, but worth potentially standardizing at some point to make things nicer for the user

I agree, I think it'd make sense to change the auto-generated code to spaces at least on this end, or use `String.replace("\t", "    ");`. We can implement this in a future PR (since this one is approved and I'll merge it in).

frontend/src/views/Game.tsx
zpChris: This waits for the default code maps (which have the code for each language) _for each **Problem**_ in the Game object to be returned. Yes, the result is a list of code maps, where each code map corresponds to a different Problem and the code maps have the language as the key and the String of start code as the value.

frontend/src/components/game/Editor.tsx
zpChris: Yes, this error will go away once we implement all of the languages 👍

frontend/src/api/Language.ts
zpChris: Good idea -- done!

PR #104, opened 2021-02-02
Implement remaining CRUD endpoints for problems

alankbi:
### List of Changes:

* Add functionality and endpoints for editing and deleting problems
* Implement test case input and ouptut format verification (must be GSON-parseable)
* Add more stringent checks for null and empty problem input values 
* Write additional test cases for new code and discovered edge cases

### Notes:

* These new endpoints are for developer productivity - when it comes time to create a bunch of problems, this + an admin panel (the next PR) will likely be very useful as opposed to creating a bunch of problems in Postman
* There were a lot of edge cases with all the problem request verification stuff - so please feel free rigorously test this and report any bugs that may have been missed haha

### DIFF
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -1,5 +1,7 @@
 package com.rocketden.main.service;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 import com.rocketden.main.dao.ProblemRepository;
 import com.rocketden.main.dto.problem.CreateProblemRequest;
 import com.rocketden.main.dto.problem.CreateTestCaseRequest;
@@ -26,6 +28,7 @@ public class ProblemService {
 
     private final ProblemRepository repository;
     private final Random random = new Random();
+    private final Gson gson = new Gson();
 
     @Autowired
     public ProblemService(ProblemRepository repository) {
@@ -91,6 +94,8 @@ public class ProblemService {
         if (updatedProblem.getDifficulty() == ProblemDifficulty.RANDOM) {
             throw new ApiException(ProblemError.BAD_DIFFICULTY);
         }
+
+        validateTestCases(updatedProblem.getTestCases());
 
         problem.setName(updatedProblem.getName());
         problem.setDescription(updatedProblem.getDescription());
@@ -193,5 +198,28 @@ public class ProblemService {
         repository.save(problem);
 
         return ProblemMapper.toTestCaseDto(testCase);
+    }
+
+    // Check to make sure test case inputs and outputs are Gson-parsable
+    private void validateTestCases(List<ProblemTestCaseDto> inputs, List<ProblemInputDto> types) {
+        for (ProblemTestCaseDto input : inputs) {
+            validateGsonParseable(input.getInput(), types);
+            validateGsonParseable(input.getOutput(), types);
+        }
+    }
+
+    private void validateGsonParseable(String input, List<ProblemInputDto> types) {
+        String[] inputs = input.trim().split("\n");
+        if (inputs.length != types.size()) {
+            throw new ApiException(ProblemError.INCORRECT_INPUT_COUNT);
+        }
+
+        for (int i = 0; i < types.size(); i++) {
+            try {
+                gson.fromJson(inputs[i], types.get(i).getType().getClassType());
+            } catch (JsonSyntaxException e) {
+                throw new ApiException(ProblemError.INVALID_INPUT);
+            }
+        }
     }
 }
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -105,12 +105,6 @@ public class ProblemService {
             throw new ApiException(ProblemError.BAD_DIFFICULTY);
         }
 
-        // Ensure that the user entered valid inputs and outputs for the problem
-        for (ProblemTestCaseDto input : updatedProblem.getTestCases()) {
-            validateInputsGsonParseable(input.getInput(), updatedProblem.getProblemInputs());
-            validateGsonParseable(input.getOutput(), updatedProblem.getOutputType());
-        }
-
         problem.setName(updatedProblem.getName());
         problem.setDescription(updatedProblem.getDescription());
         problem.setDifficulty(updatedProblem.getDifficulty());
@@ -121,7 +115,12 @@ public class ProblemService {
             problem.addProblemInput(ProblemMapper.toProblemInputEntity(problemInput));
         }
 
+        problem.getTestCases().clear();
         for (ProblemTestCaseDto testCaseDto : updatedProblem.getTestCases()) {
+            // Ensure that the user entered valid inputs and outputs for the problem
+            validateInputsGsonParseable(testCaseDto.getInput(), updatedProblem.getProblemInputs());
+            validateGsonParseable(testCaseDto.getOutput(), updatedProblem.getOutputType());
+
             ProblemTestCase testCase = new ProblemTestCase();
             testCase.setInput(testCaseDto.getInput());
             testCase.setOutput(testCaseDto.getOutput());
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -22,6 +22,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 @Service
 public class ProblemService {
@@ -95,7 +96,7 @@ public class ProblemService {
             throw new ApiException(ProblemError.BAD_DIFFICULTY);
         }
 
-        validateTestCases(updatedProblem.getTestCases());
+        validateTestCases(updatedProblem.getTestCases(), updatedProblem.getProblemInputs());
 
         problem.setName(updatedProblem.getName());
         problem.setDescription(updatedProblem.getDescription());
@@ -183,6 +184,16 @@ public class ProblemService {
         if (request.getInput() == null || request.getOutput() == null) {
             throw new ApiException(ProblemError.EMPTY_FIELD);
         }
+
+        // Verify inputs are of valid form
+        List<ProblemInputDto> inputs = problem.getProblemInputs()
+                .stream()
+                .map(ProblemMapper::toProblemInputDto)
+                .collect(Collectors.toList());
+
+        validateGsonParseable(request.getInput(), inputs);
+        validateGsonParseable(request.getOutput(), inputs);
+
 
         ProblemTestCase testCase = new ProblemTestCase();
         testCase.setInput(request.getInput());
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -101,7 +101,7 @@ public class ProblemService {
         // Ensure that the user entered valid inputs and outputs for the problem
         for (ProblemTestCaseDto input : updatedProblem.getTestCases()) {
             validateGsonParseable(input.getInput(), updatedProblem.getProblemInputs());
-            validateGsonParseable(input.getOutput(), updatedProblem.getProblemInputs());
+            validateGsonParseable(input.getOutput(), updatedProblem.getOutputType());
         }
 
         problem.setName(updatedProblem.getName());
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -117,7 +117,15 @@ public class ProblemService {
     }
 
     public ProblemDto deleteProblem(String problemId) {
-        return null;
+        Problem problem = repository.findProblemByProblemId(problemId);
+
+        if (problem == null) {
+            throw new ApiException(ProblemError.NOT_FOUND);
+        }
+
+        repository.delete(problem);
+
+        return ProblemMapper.toDto(problem);
     }
 
     public List<ProblemDto> getAllProblems() {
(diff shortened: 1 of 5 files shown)


alankbi:
[Implement remaining CRUD endpoints for problems](https://trello.com/c/8sr9Jcfc/261-implement-remaining-crud-endpoints-for-problems)

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -83,6 +86,71 @@ public ProblemDto getProblem(String problemId) {
...
     }
 
+    public ProblemDto editProblem(String problemId, ProblemDto updatedProblem) {
+        Problem problem = repository.findProblemByProblemId(problemId);
+
+        if (problem == null) {
+            throw new ApiException(ProblemError.NOT_FOUND);
+        }
+
+        if (updatedProblem.getName() == null || updatedProblem.getDescription() == null
```
zpChris, line 96: Not sure this is necessary for this PR, but I notice that we do a lot of these null checks, where we iterate over the different fields and check if they are null -- and if so, throwing an error. I had an idea that on a future PR, we could simplify this by making a single utility function that iterates over the different fields to check if all have a value, similar to the checkLocation function we did on the frontend. Thoughts?

zpChris: Also, just on this PR, could we check that `updatedProblem == null` in this condition as well? We do that for the `problem`, but not for `updatedProblem`, meaning the wrong error could be thrown.

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -164,6 +242,38 @@ public ProblemTestCaseDto createTestCase(String problemId, CreateTestCaseRequest
         return ProblemMapper.toTestCaseDto(testCase);
     }
 
+    // Check to make sure test case inputs and outputs are Gson-parsable
+    protected void validateGsonParseable(String input, List<ProblemInputDto> types) {
```
zpChris, line 246: What was your thinking behind making this `protected`?

Also, nit: Change to `validateInputsGsonParseable` or something similar, to differentiate it from the below method. (Not a huge deal either way, but I think it could be helpful to make it a different method name so as to avoid confusion.)

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -164,6 +242,38 @@ public ProblemTestCaseDto createTestCase(String problemId, CreateTestCaseRequest
...
+        if (input == null) {
+            throw new ApiException(ProblemError.INVALID_INPUT);
+        }
+
+        String[] inputs = input.trim().split("\n");
```
zpChris, line 251: really minor nit: Add comment noting that each input is on a new line.

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -83,6 +86,71 @@ public ProblemDto getProblem(String problemId) {
...
+        problem.setDifficulty(updatedProblem.getDifficulty());
+        problem.setOutputType(updatedProblem.getOutputType());
+
+        problem.getProblemInputs().clear();
+        for (ProblemInputDto problemInput : updatedProblem.getProblemInputs()) {
+            if (problemInput != null) {
+                problem.addProblemInput(ProblemMapper.toProblemInputEntity(problemInput));
+            } else {
+                throw new ApiException(ProblemError.BAD_INPUT);
+            }
```
zpChris, line 124:
```suggestion
            problem.addProblemInput(ProblemMapper.toProblemInputEntity(problemInput));
```
(I believe you already have this check in `validateGsonParseable`, so I don't think there's a need for it here.)

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -83,6 +86,71 @@ public ProblemDto getProblem(String problemId) {
...
+        }
+
+        for (ProblemTestCaseDto testCaseDto : updatedProblem.getTestCases()) {
+            ProblemTestCase testCase = new ProblemTestCase();
+            testCase.setInput(testCaseDto.getInput());
+            testCase.setOutput(testCaseDto.getOutput());
+            testCase.setHidden(testCaseDto.isHidden());
+            testCase.setExplanation(testCaseDto.getExplanation());
+
+            problem.addTestCase(testCase);
```
zpChris, line 134: Should we add any checks here, or no need? For example, we could verify that this can be properly converted into the relevant types using the same method `validateGsonParseable`, while still just setting the problem with this the input string, output string, etc.

zpChris: Or, perhaps another solution is to create and save the problem, then call `createTestCase` with the relevant inputs and adding it through that method.

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -333,4 +361,210 @@ public void getRandomProblemNotFound() {
...
+    @Test
+    public void deleteProblemFailure() {
+        ApiException exception = assertThrows(ApiException.class, () -> problemService.deleteProblem("ZZZ"));
+
+        verify(repository).findProblemByProblemId("ZZZ");
+        assertEquals(ProblemError.NOT_FOUND, exception.getError());
+    }
+
+    @Test
+    public void validateGsonParseable() {
```
zpChris, line 529: I'm not sure if I missed it, but do you have a test case where multiple inputs and input types are _correctly_ provided for a problem? You addressed the errors on this, I want to check that `String.split("\n");` is doing its job basically.

zpChris (changes requested):
Requesting changes just because I left some nits and minor changes, but really this PR looks great! I expect I'll approve after these few nits are looked at 👍

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: Yeah I def think that could be helpful in the future - also sounds good, I added that null check

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: Mainly so that the method could directly be tested (not sure if there's a way to do that without using something new like Powermock)

Also sounds good on the nit

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: I think it should already be verified above at line 109, although lmk if I missed any situation or I should move that block down or something (maybe that helps with readability?)

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
alankbi: I think the first three lines here should verify that they work - although were you referring to like a full unit/API test that covers multi-line inputs? If so, I can def add that too, just lmk

zpChris: Ah I see -- I didn't connect in my mind the idea that the method shown only throws an error if it _doesn't_ work, and otherwise just does nothing. Is there any case where that could fail but the exception is not checked, so nothing happens? I guess I'm wondering if you should just stick a `return true` at the end of something and check that is returned.

I don't think this is necessary, so I'm guessing you can just say so and resolve this comment, but I'm leaving this comment anyway just in case.

src/main/java/com/rocketden/main/service/ProblemService.java
zpChris: Ah, I see -- makes sense thank you!

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -242,6 +248,28 @@ public void createTestCaseFailure() {
...
+        expected.setOutputType(IO_TYPE_2);
+
+        CreateTestCaseRequest request = new CreateTestCaseRequest();
+        request.setInput(INPUT);
+        request.setOutput(OUTPUT);
+
+        Mockito.doReturn(expected).when(repository).findProblemByProblemId(expected.getProblemId());
+
+        ApiException exception = assertThrows(ApiException.class, () ->
+                problemService.createTestCase(expected.getProblemId(), request));
```
zpChris, line 267: **Note:** There are four other examples with this same fix below.

```suggestion
        String problemId = expected.getProblemId();
        ApiException exception = assertThrows(ApiException.class, () ->
                problemService.createTestCase(problemId, request));
```
This suggestion is based off the attached screenshot; since `expected.getProblemId()` could throw an error / exception as well, we should calculate that before so we can ensure that the error thrown comes from `createTestCase`.

Exception of Inner Methods in Assert Throws

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -83,6 +86,68 @@ public ProblemDto getProblem(String problemId) {
...
+
+        if (updatedProblem.getDifficulty() == ProblemDifficulty.RANDOM) {
+            throw new ApiException(ProblemError.BAD_DIFFICULTY);
+        }
+
+        // Ensure that the user entered valid inputs and outputs for the problem
+        for (ProblemTestCaseDto input : updatedProblem.getTestCases()) {
+            validateInputsGsonParseable(input.getInput(), updatedProblem.getProblemInputs());
+            validateGsonParseable(input.getOutput(), updatedProblem.getOutputType());
+        }
```
zpChris, line 112: This for loop checks for the test cases provided in the `updatedProblem`, but it does not check if this fits with the test cases already existing in the `problem` object in the database. This presents an issue where a user can create a problem that accepts an **ARRAY_INTEGER** input with the appropriate test case, and then the same user can send an edit request to the problem with an **INTEGER** input an a _new_ appropriate test case. Then, both test cases will exist in the database, where one of them does not fit the current problem schema. I see a few potential solutions to this:

1. Do not allow updates to the `problemInputs` and `outputType` fields. This would restrict editing capabilities but effectively remove this problem.
2. Allow updates to `problemInputs` and `outputType` fields, and remove the existing test cases that _conflict_ with this change (many input types can be seen as another, like STRING and INTEGER in certain cases).
3. Allow updates to `problemInputs` and `outputType` fields, and throw an error if _any_ test case (in existing or updated) conflict with the change.
4. Allow updates to `problemInputs` and `outputType` fields, then remove all existing test cases if these fields are updated.
5. Add a `boolean` value called `rewriteTests` to the `editProblem` request (preferably in the request body, though this would require a new object) that always removes any tests in the problem and replaces them with the new tests in `updatedProblem`.
6. Add a `boolean` value called `rewriteTests` to the `editProblem` request (preferably in the request body, though this would require a new object) that always removes any tests in the problem _if they conflict_ with any new schema that is set.

There are probably more solutions as well -- let me know if you have any preference!

zpChris (commented):
Not requesting changes, but not approving since I saw a potential problem with updating the problem input types and output type schema. See my comment for possible solutions -- let me know what you think!

I'm also open to just providing a temporary fix in this PR (like removing all existing test cases in the problem), and then opening a new PR for a larger fix, if you really want to merge this through.

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
alankbi: Hm yeah that's a good point, I hadn't thought about a case in which it would fail but not throw an exception. My suspicion is that this scenario wouldn't be possible (since if something bad happens for it to not return true then an exception has to be thrown) so I'll prob leave it as is, but I'm ambivalent either way if you want to unresolved this

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: Good catch - I actually created a test case for this (`editProblemNewProblemInputInvalidatesTestCases`) along the lines of solution 3, but it's a good thing you brought this up because I looked at it again and there's a bug haha - the test cases in updatedProblem are supposed to contain all the test cases (old and new) so it should throw an error if any of the inputs get invalidated, but I forgot to clear/replace the existing test cases in the database before setting the new ones so it's basically just adding a whole bunch of duplicate test cases on every edit

It should be fixed now - lmk if you notice any other bugs and things like that

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
alankbi: Sounds good - I made a separate commit for this since as you mentioned, there were other places that needed to be updated too

src/main/java/com/rocketden/main/service/ProblemService.java
zpChris: Awesome, it's fixed on my end as well! To clarify, if the `updatedProblem` has no test cases while the `problem` has some old test cases, this means that the user is attempting the edit the problem and remove all test cases?

This sounds like a fine solution, I just wanted to make sure that's actually what is happening here.

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -83,6 +86,67 @@ public ProblemDto getProblem(String problemId) {
...
     }
 
+    public ProblemDto editProblem(String problemId, ProblemDto updatedProblem) {
+        Problem problem = repository.findProblemByProblemId(problemId);
+
+        if (problem == null) {
+            throw new ApiException(ProblemError.NOT_FOUND);
+        }
+
+        if (updatedProblem == null || updatedProblem.getName() == null
```
zpChris, line 96: Add check for `updatedProblem.getTestCases() == null` as well, otherwise a 405 method not allowed error is shown, instead of `ProblemError.EMPTY_FIELD`.

OR, if you want users to pass in a `null` field for `testCases` to indicate that the updated problem should have no test cases, you should modify the for loop line `for (ProblemTestCaseDto testCaseDto : updatedProblem.getTestCases()) {` so that it doesn't throw an error.

zpChris (approved):
LGTM! Left one more minor fix, but the changes look good, and I'm excited to see the frontend editing setup 😃

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: Yep - that's the intention

src/main/java/com/rocketden/main/service/ProblemService.java
alankbi: Went with option one 👍

PR #106, opened 2021-02-15
Integrate main to use updated tester response data

alankbi:
### List of Changes:

* Update TesterReponse class to match the RunDto on the tester repo 
* Create new TesterError class to map errors from the tester repo
* Return different error for tester triggered error and internal POST call error
* Add additional test cases for newly refactored SubmitService setup
* Change default code to use spaces to prevent Python tab-space mix error on tester
* Display output from submission on the frontend, in the "output" container.
* Create separate calls in the backend to differentiate between "run code" and "submit".

### Notes:

* The tester API call should have slightly better test coverage now, although not perfect 
* The TesterError class is not an enum, so we can dynamically create them and don't have to copy all the errors from the tester repo over into main
* Right now, the output displayed in a small box, and is definitely not user-friendly. I'm assuming this is a low priority right now though, as we are just trying to get the MVP out.
* If the problem does not have at least one test case, it cannot do the "run code" operation as the output passed through there is the same as the first test case, to ensure that it passes all parsing checks in the tester repository.
* If the user does not have a "return" statement (at least in Java code), this is seen as misformatted output and not as a compilation error. I'm not certain why this is, but I figured troubleshooting this issue could be left to another PR - see [the #15 tester issue](https://github.com/rocketden/tester/issues/15).

### Screencast and Images:

[Screencast of the Run Code and Submit Functionality](https://www.loom.com/share/eb10a65b09174bd69477bb5ae7c47dfa)

**Images of the Correct Submission with Hidden and Open Test Cases**
_Three submissions are shown below, as I scrolled down the "output" container (it could not be enlarged any more)._

Image of the Correct Submission
Image of the Correct Submission
Image of the Correct Submission

**Image of the Test Case Result**
Image of the Test Case Result

### DIFF
--- a/frontend/src/views/Game.tsx
+++ b/frontend/src/views/Game.tsx
@@ -20,7 +20,7 @@ import { User } from '../api/User';
 import { GameNotification, NotificationType } from '../api/GameNotification';
 import Difficulty from '../api/Difficulty';
 import {
-  Game, getGame, Player, SubmissionResult, submitSolution,
+  Game, getGame, Player, SubmissionResult, submitSolution, runSolution,
 } from '../api/Game';
 import LeaderboardCard from '../components/card/LeaderboardCard';
 import GameTimerContainer from '../components/game/GameTimerContainer';
@@ -202,7 +202,30 @@ function GamePage() {
   };
 
   // Callback when user runs code against custom test case
-  const runSolution = () => {
+  const runCode = (input: string) => {
+    setLoading(true);
+    setError('');
+    const request = {
+      initiator: currentUser!,
+      input,
+      code: currentCode,
+      language: currentLanguage,
+    };
+
+    runSolution(roomId, request)
+      .then((res) => {
+        setLoading(false);
+        setSubmission(res);
+        checkSendTestCorrectNotification(res);
+      })
+      .catch((err) => {
+        setLoading(false);
+        setError(err.message);
+      });
+  };
+
+  // Callback when user runs code against custom test case
+  const submitCode = () => {
     setLoading(true);
     setError('');
     const request = {
@@ -316,7 +339,8 @@ function GamePage() {
               <Console
                 testCases={problems[0]?.testCases}
                 submission={submission}
-                onRun={runSolution}
+                onRun={runCode}
+                onSubmit={submitCode}
               />
             </Panel>
           </SplitterLayout>
--- a/frontend/src/views/Game.tsx
+++ b/frontend/src/views/Game.tsx
@@ -20,7 +20,7 @@ import { User } from '../api/User';
 import { GameNotification, NotificationType } from '../api/GameNotification';
 import Difficulty from '../api/Difficulty';
 import {
-  Game, getGame, Player, Submission, submitSolution, runSolution,
+  Game, getGame, Player, Submission, submitSolution, runSolution, SubmissionType,
 } from '../api/Game';
 import LeaderboardCard from '../components/card/LeaderboardCard';
 import GameTimerContainer from '../components/game/GameTimerContainer';
@@ -230,6 +230,9 @@ function GamePage() {
     runSolution(roomId, request)
       .then((res) => {
         setLoading(false);
+
+        // Set the 'test' submission type to correctly display result.
+        res.submissionType = SubmissionType.Test;
         setSubmission(res);
         checkSendTestCorrectNotification(res);
       })
@@ -252,6 +255,9 @@ function GamePage() {
     submitSolution(roomId, request)
       .then((res) => {
         setLoading(false);
+
+        // Set the 'submit' submission type to correctly display result.
+        res.submissionType = SubmissionType.Submit;
         setSubmission(res);
         checkSendSolutionCorrectNotification(res);
       })
--- a/frontend/src/views/Game.tsx
+++ b/frontend/src/views/Game.tsx
@@ -20,7 +20,7 @@ import { User } from '../api/User';
 import { GameNotification, NotificationType } from '../api/GameNotification';
 import Difficulty from '../api/Difficulty';
 import {
-  Game, getGame, Player, SubmissionResult, submitSolution, runSolution,
+  Game, getGame, Player, Submission, submitSolution, runSolution,
 } from '../api/Game';
 import LeaderboardCard from '../components/card/LeaderboardCard';
 import GameTimerContainer from '../components/game/GameTimerContainer';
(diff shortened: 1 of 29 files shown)


alankbi:
[Integrate responses on main to use updated tester endpoint](https://trello.com/c/u53sx421/265-integrate-responses-on-main-to-use-updated-tester-endpoint)

alankbi:
This should now be working for both Python and Java - I think the remaining thing is just whether we wanted to add the displaying of the first user output/wrong answer/error/compilation error @zpChris

zpChris:
Added more to this PR: run code functionality and displaying more specific output on the frontend.

zpChris (approved):
LGTM!

I worked on this PR after @alankbi implemented the connection to the tester repository, so he should review it as well. However, I couldn't figure out how to request review from the PR's owner, and I couldn't determine how to change the PR owner either. @alankbi, are you able to review the PR yourself without me requesting your review on it?

frontend/src/views/Game.tsx
```diff
@@ -199,20 +201,69 @@ function GamePage() {
     }
   };
 
+  // Send notification if solution is correct and currentUser is set.
+  const checkSendSolutionCorrectNotification = (submissionParam: Submission) => {
```
alankbi, line 205: If the other notification method is no longer being used could we delete it?

alankbi: Oh I just saw it's still being called in the other runSolution method - does this mean that if someone checks against a custom test case, other people will see that they tested it locally?

src/main/java/com/rocketden/main/dto/game/SubmissionResultDto.java
```diff
@@ -0,0 +1,17 @@
...
+@Getter
+@Setter
+public class SubmissionResultDto {
+    // This class is mapped with the tester service's ResultDto class.
+    private String console;
+    private String userOutput;
+    private String error;
+    private String input;
+    private String correctOutput;
+    private boolean hidden;
```
alankbi, line 15: Note for future: if we go like a competition route we will prob have to remove some of these fields to prevent people inspecting element/cheating

src/main/java/com/rocketden/main/dto/game/TesterResponse.java
```diff
@@ -3,10 +3,17 @@
...
+import com.rocketden.main.game_object.SubmissionResult;
+
 @Getter
 @Setter
-// This class is mapped directly to the tester service's RunDto class.
 public class TesterResponse {
-    private Boolean status;
-    private String output;
+    // This class is mapped directly to the tester service's RunDto class.
+    private List<SubmissionResult> results;
```
alankbi, line 14: nit: Keep this as a separate class rather than reusing SubmissionResult, that way we can decouple the two services (so if tester changes its response object, we aren't forced to perfectly mirror the response in the main repo database

frontend/src/views/Game.tsx
zpChris: Yeah that's right - I'm thinking that all of these notifications could be incorporated into an "activity feed" similar to [binarysearch](https://binarysearch.com/)'s chat functionality, but just in a separate feed (or perhaps a toggle-able feed).

src/main/java/com/rocketden/main/dto/game/SubmissionResultDto.java
zpChris: Lol hmm good point - we may want to have a sprint dedicated to all the security problems and bug fixes and whatnot. We could have like a UI sprint and a security / bug fix sprint 👍

src/main/java/com/rocketden/main/dto/game/TesterResponse.java
zpChris: Sounds good, will do!

src/test/java/com/rocketden/main/service/SubmitServiceTests.java
```diff
@@ -43,6 +61,43 @@
...
+        request.setInput(INPUT);
+        request.setInitiator(UserMapper.toDto(user));
+
+        SubmissionDto submissionDto = submitService.runCode(game, request);
+
+        assertEquals(CODE, submissionDto.getCode());
+        assertEquals(LANGUAGE, submissionDto.getLanguage());
+        assertEquals(submissionDto.getNumCorrect(), submissionDto.getNumTestCases());
+        assertNotNull(submissionDto.getRuntime());
+        assertFalse(game.getAllSolved());
```
alankbi, line 98: Now that we've added the results object to the Submission classes, could we add checks for those fields? (might require updating the getDummySubmission method to set those fields too)

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -147,6 +147,22 @@ public void setStartGameTimer(Game game, Long duration) {
         gameTimer.getTimer().schedule(endGameTimerTask, duration * 1000);
     }
 
+    // Test the submission, return the results, and send a socket update
+    public SubmissionDto runCode(String roomId, SubmissionRequest request) {
+        Game game = getGameFromRoomId(roomId);
+
+        if (request.getInitiator() == null || request.getCode() == null || request.getLanguage() == null || request.getInput() == null) {
```
alankbi, line 154: Looks good - a small nit I'd have would just be to add one more API test for this endpoint

alankbi:
Left a few more nits on this PR but otherwise looks good to me - lmk if you need any help for implementing some of this stuff too

src/main/java/com/rocketden/main/dto/game/TesterResponse.java
zpChris: Done - there were a decent amount of Mapper changes and an additional test for the `TesterResult` class, so you may want to take a look at the new changes just to make sure everything looks ok.

src/main/java/com/rocketden/main/service/GameManagementService.java
zpChris: Sounds good, I'll add this now. 👍

src/test/java/com/rocketden/main/service/SubmitServiceTests.java
zpChris: Yes, sounds good - done!

zpChris:
@alankbi, I added ~ 230 lines of code through tests and addressing the above notes, so you may want to re-review specifically the changes made - let me know when you think it's good to merge!

alankbi:
Looks good to merge for me :)

PR #107, opened 2021-02-27
Fix the frontend inaccurate timer.

### LINKED ISSUE #99: [Safari] Timer is Off by Two Minutes on Frontend
On the frontend of the Game page, the timer is off by two minutes (if you set it to 15 minutes, it counts down from 13 minutes; if you set it to 1 minute, it just shows 'Loading...'). The backend redirect isn't triggered until the **real, correct** timer finishes the countdown, but I don't know why this issue happens on the Safari frontend.

[Screencast of [Safari] Timer is Off by Two Minutes on Frontend](https://www.loom.com/share/1b2cd578939940da9e054e104b5ae085)

### LINKED ISSUE #110: [Safari] Multiple-contestant game page does not load properly
When multiple contestants are in a room, one of those contestants (I haven't solidified whether it has to be the host) is using Safari, the game page does not load - see [this comment](https://github.com/rocketden/main/pull/108#issuecomment-788710654) for a longer description.

zpChris:
### List of Changes:

- Fix #99, and other relevant timer bugs, by replacing `LocalDateTime` with `Instant` on the backend, and ending usage of the system's local timer (`Date.now()` and `new Date()`).
- Add `getInstant` endpoint to get the current backend `Instant`, which is the only reliable source of time information (other than similar endpoints on the web).
- Change fetch types to EAGER, which seems to also fix #110.
- Remove the debugging statements for the socket `StompClient` - see the [referenced solution on Stack Overflow](https://stackoverflow.com/questions/25683022/how-to-disable-debug-messages-on-sockjs-stomp).

### Notes:

- The motivation to change `LocalDateTime` to `Instant` is largely detailed [here](http://blog.schauderhaft.de/2018/03/14/dont-use-localdatetime/).
- The commands `Date.now()` and `new Date()` on the JavaScript frontend use the local computer's system timer; this means if the user updates their local clock, the frontend timer will change.
- When testing, I found that the frontend timer is occasionally off. I added a short time speed up of one second on the frontend, to compensate for the time it takes to make REST requests in `getInstant`, and so that the frontend timer should typically end slightly before the backend one.
- ~Another potential solution, instead of using a `setInterval` function that decrements a countdown every second, would be to calculate the offset between the real time and the local system time, then use that and the local system time to update the timer. This may cause issues if the user changes their local clock, but may be more reliable in longer games - thoughts, @alankbi?~ I changed this implementation to use the local clock to update the countdown; this does not control the backend however, so if the user messes with their local clock, it will only affect the display of the frontend time, not when the game actually ends or any game functionality.
  - The solution now seems to work, though sometimes the timer at the start is off by a second or two. Also, if the user changes the system clock, it will only change the frontend.

### Screencast and Images:

I am in the process of testing this PR, and will add these soon.


zpChris:
[Fix frontend inaccurate timer issue](https://trello.com/c/D1meogKp/272-fix-frontend-inaccurate-timer-issue)

alankbi (approved):
Looks good! Tests are failing right now but once those are fixed it should be good to merge I think

(I haven't looked too much into it but if it's not just a one-off error, then it might have something to do with the serialization of Instant in the DTO classes since we're getting rid of those `@LocalDateTimeSerializer` annotations)

alankbi:
Actually, I think that bug mentioned above is causing the page itself to not load sometimes - specifically when multiple people are in a room and you're redirecting to the game page, there's the following error: 

```
TypeError: Cannot read property 'map' of undefined
    at Kn (Game.tsx:278)
```

My guess is it has something to do with the socket or REST messages just being empty due to the failure of serialization?

zpChris:
Yeah sorry I saw the failing tests, I switched to coding #108 when working on this PR - I'll address the failing tests and `LocalDateTime` conversion issues. 👍

zpChris:
Thanks for finding a fix for this @alankbi! 

I wasn't able to test that this actually works remotely as the Heroku login seems to be broken for me right now, but hopefully I can test that soon. I'm also hoping to test that the time stays on track even for longer periods, like ten, thirty, and sixty minutes 👍 

Also, I had a note for the future: What happens if a user clicks "submit", and then before the result is displayed on the frontend, the timer ends? Would the result go through, would it be a mix (not immediately shown in the rankings, but in the database records), or something else?

zpChris:
I'm going to continue working on this, as it's slightly off for longer times. My current goal is to calculate the offset between the local Date and the backend Instant, and then use that to update the timer (and check for sudden large differences in the local Date).

zpChris:
Found something interesting - at least for individual rooms, both [binarysearch](https://binarysearch.com) and [LeetCode](https://leetcode.com/interview/1/) seem to use the local computer's timer as the single source of truth. This means that if I'm on LeetCode and I start an hour-long mock interview, then I move my local clock ahead an hour, the room immediately ends.

I think this can present some real (but niche, to be fair) issues with things like [Daylight Savings Time](https://stackoverflow.com/questions/2532729/daylight-saving-time-and-time-zone-best-practices), as I'm not sure how the local clock updates then. I'm aiming for our app to still function completely properly in that case, with the note that the _frontend_ timer will be off, as it's difficult to calculate when the user is editing their system clock directly and I don't think it'd be worth it to account for that edge case right now.

alankbi (approved):
👍

PR #111, opened 2021-03-05
Refresh and active status support on game/results page + README improvements

alankbi:
### List of Changes:

* Allow refreshing and navigating forward/back on the game and results pages
* Display players as active or inactive depending on socket status on the game/results pages
* Update results page to handle submissions that finish after game ends
* Write new test for game socket connect/disconnect behavior
* Save and load player's latest code/language after refreshes
* Set min-height on game page splitter container
* Update README with latest startup instructions and file structure
* Fix timer bug where notifications and time over socket updates are sent from previous game if playing again

### Notes:

* Only the latest submission's code is saved - so if a user coded in two languages, only one is saved, or if a user never submitted/ran code, nothing is saved
* A lot of infrequent socket bugs can occur (e.g. user shown as inactive at end of screencast) that disappear on refresh or are like a one-time occurrence - I don't know how to fix these (or if it's even possible to given network speeds, timing, etc. that can mess with the sending of socket updates). Potentially we could set up a set interval (maybe 3s or 5s) to repeatedly send out socket updates, that way any inaccuracies are always cleared up after a short period of time? 
* Some confusing bits on the language stuff - seems like there's three different states: 1. monaco editor string (python), 2. display string/enum key (Python), and 3. backend value/enum value (PYTHON). This took a lot of trial and error to get to work in unison with saving code across refreshes, and it's still a bit confusing, but this is my understanding: 
  * `'PYTHON'` to `Language.Python` -> use `value as Language` (converting backend strings into enums)
  * `'Python'` to `Language.Python` -> use `fromString` (converting select options to enums)
  * `Language.Python` to `python` -> use `languageToEditorLanguage` (converting enums to monaco editor language)

### Screencast and Images:

[Screencast showing new functionality and behavior](https://www.loom.com/share/acdb2b4d30714ca9816dcb3ec2e2e609)


zpChris:
[Refresh and active status support on game/results page + README updates](https://trello.com/c/fcfN3CPL/221-refresh-and-active-status-support-on-game-results-page-readme-updates)

zpChris:
Addressing a couple notes you put in your PR description:

> Only the latest submission's code is saved - so if a user coded in two languages, only one is saved, or if a user never submitted/ran code, nothing is saved

When the user submits or runs their code, should we also run a POST request to the `updateCode` function that's in the backend? That way, when the user refreshes, a GET request could be run to return the last-updated code to the user. This all being said, it's probably less ideal than storing it in localstorage, so maybe it's a poor idea to implement the above. Either way, I think this would go for a future PR - the current setup is certainly a huge improvement so thanks! 💯 

> Potentially we could set up a set interval (maybe 3s or 5s) to repeatedly send out socket updates, that way any inaccuracies are always cleared up after a short period of time?

I'm open to this - would that take up a lot of computing power? It's a good idea though, I'm not sure why our setup seems unreliable that way, maybe it has to do with keeping that connection alive and consistent pinging (like the ["heartbeat" or "ping pong" idea](https://stackoverflow.com/questions/46111656/must-websockets-have-heartbeats)).

> Some confusing bits on the language stuff - seems like there's three different states...

Wow good job on this 👍

zpChris:
Just to double-check - it seems like once the user presses "exit game" they cannot return. That's awesome, just wanted to check whether that's actually true / intentional or not.

zpChris:
Not sure if this is already an issue somewhere (I didn't see it let me know), but for a future PR, I think we can fix the problem where if a user exits the game and everybody else has finished, it goes to the results page.

alankbi:
> When the user submits or runs their code, should we also run a POST request to the `updateCode` function that's in the backend? That way, when the user refreshes, a GET request could be run to return the last-updated code to the user. This all being said, it's probably less ideal than storing it in localstorage, so maybe it's a poor idea to implement the above. Either way, I think this would go for a future PR - the current setup is certainly a huge improvement so thanks! 💯

Might have misstated this but yeah as of now this is what it does (gets the latest submitted or ran code - although not using updateCode since I don't think we've made that a public endpoint yet). However unless we periodically sync or save multiple languages at a time, it has the issues mentioned above - but I agree about the local storage stuff, I think that's prob the best path moving forward in future PRs

> Not sure if this is already an issue somewhere (I didn't see it let me know), but for a future PR, I think we can fix the problem where if a user exits the game and everybody else has finished, it goes to the results page.

Hm yeah that's a good point - I can see us maybe wanting to do some time-related feature where if a person leaves for more than X amount of time, the game just ends, but that way it doesn't just immediately end if someone refreshes the page

frontend/src/views/Results.tsx
```diff
@@ -28,36 +31,65 @@ function GameResultsPage() {
...
+          playAgainAction(updatedGame);
         }
       };
 
-      subscribe(routes(location.state.game.room.roomId).subscribe_game, subscribeCallback)
-        .catch((err) => setError(err.message));
+      /**
+       * Connect, subscribe, and then finally get the game details. Doing so in this order ensures
+       * that any late submissions are properly received (either through the socket update or
+       * through the get game request) and reflected on the leaderboard.
```
zpChris, line 74: So if a user submits their code right before the timer ends, the results page will initially display something, then will update through socket. I'm thinking that a good addition to make (doesn't have to be in this PR or sprint, just an idea) would be to have a field in the `Game` class that indicates a user's submission is currently being calculated, and if so, you could display something on the frontend that indicates as such, showing that the scores might change as the final submission comes in.

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -54,7 +54,7 @@ protected GameManagementService(RoomRepository repository, SocketService socketS
         currentGameMap = new HashMap<>();
     }
 
-    protected Game getGameFromRoomId(String roomId) {
+    public Game getGameFromRoomId(String roomId) {
```
zpChris, line 57: Why was this changed from `protected` to `public`?

src/main/java/com/rocketden/main/socket/WebSocketConnectionEvents.java
```diff
@@ -81,6 +88,14 @@ public void onSocketConnected(SessionConnectedEvent event) {
...
             RoomDto roomDto = RoomMapper.toDto(room);
             socketService.sendSocketUpdate(roomDto);
+
+            // If a game exists, update the room info for that game
+            try {
+                Game game = gameService.getGameFromRoomId(room.getRoomId());
+                game.setRoom(room);
+                game.getPlayers().get(user.getUserId()).setUser(user);
+                socketService.sendSocketUpdate(GameMapper.toDto(game));
+            } catch (ApiException ignored) {}
```
zpChris, line 98: This does have a Sonarlint warning (see below image), but I'm ok with that -- I'm wondering, would it be useful to add any logging statement there?

Sonarlint Warning on Empty Code Block

zpChris: Also, since you have this same block of code in two places (here and the method below), perhaps you make another method it for this code and invoke that?

frontend/src/components/card/PlayerResultsCard.tsx
```diff
@@ -48,7 +48,10 @@ function PlayerResultsCard(props: PlayerResultsCardProps) {
       <FlexHorizontalContainer>
         <FlexLeft>
           <CenteredContainer>
-            <PlayerIcon hexColor={color.hexColor} />
+            <PlayerIcon
+              hexColor={color.hexColor}
+              active={Boolean(player.user.sessionId || isCurrentPlayer)}
```
zpChris, line 53: Just wanted to note that here the current player is always shown to themselves as active, but on the lobby page the current user can be inactive; not sure which should change, both are fine by me, but should those be the same?

zpChris (changes requested):
This looks great, nice job! Requested changes just because I left some nits and potentially larger changes (up to you whether you want to implement any of them or leave them for future PRs, since this was likely a good amount of work), but depending on what you choose to update, the changes should be pretty small.

zpChris:
One other note - when we implement the game page so that hitting refresh truly means all changes will be saved (right now they may lose some code), we can remove the warning label. I'm ok leaving it in this PR though, just wanted to note that for the future.

frontend/src/views/Game.tsx
```diff
@@ -108,11 +112,22 @@ function GamePage() {
 
     // Get the result of promises and set the default code list.
     Promise.all(promises).then((result) => {
+      const codeMap = result[0];
+
+      // If previous code and language specified, save those as defaults
+      if (code) {
+        codeMap[language] = code;
+      }
```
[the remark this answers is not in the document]
alankbi: Yeah I saw the other comment, this is what I replied: 

> Might have misstated this but yeah as of now this is what it does (gets the latest submitted or ran code - although not using updateCode since I don't think we've made that a public endpoint yet). However unless we periodically sync or save multiple languages at a time, it has the issues mentioned above - but I agree about the local storage stuff, I think that's prob the best path moving forward in future PRs

So yeah in essence I agree that at some point we'll want to be able to save multiple languages - I think we'll probably want to use localstorage as you suggested in the other comment, that way everything's saved regardless of running/submitting (and also helps to keep the backend logic simpler)

In the meantime I'll add a card for this in the backlog

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: In order to do the socket stuff I had to call this method from the `WebSocketConnectionEvents.java` so I had to change this to public (I might actually be able to change this back to protected though reading one of your other comments/proposed changes, so we'll see haha)

frontend/src/views/Results.tsx
alankbi: I think this is a great idea - I'll add a card to the backlog for this (hopefully this type of thing doesn't show up too often but at some point it'll def be a good idea to make a fix)

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Just changed it back - made a separate method as suggested in another comment so this doesn't need to be public anymore

src/main/java/com/rocketden/main/socket/WebSocketConnectionEvents.java
alankbi: Sounds good - just made a new method for this and also added our first log for the main repo haha (which would prob be a good habit from here on out)

frontend/src/components/card/PlayerResultsCard.tsx
alankbi: I'd be down for that - I made this one show up as always active if it's the current player to help minimize potential bugs, so maybe we could do the same for the lobby page. However, maybe we could also wait till the UI PRs for that, since the way the lobby page is currently implemented, it splits it by users in the activeUsers and inactiveUsers list, so doing a conditional check for isCurrentPlayer might not fit into place under the current design.

alankbi:
> One other note - when we implement the game page so that hitting refresh truly means all changes will be saved (right now they may lose some code), we can remove the warning label. I'm ok leaving it in this PR though, just wanted to note that for the future.

One thing to note about this plus the end game when all active users are done story is that it might take extra work/complexity to figure out when someone actually leaves the page for good vs. is just temporarily refreshing the page. Might require us to do some timer stuff or like store like time of connection/disconnection to calculate when to end the game, which might be a little bit of extra difficulty to these two tasks

Another thing is that occasionally it seems after using the markdown editor dependency from #109, the build fails with this: 

Build failed due to npm stuff most likely

It only appears now and again (such as the previous two commits of mine as you can see), but might be worth looking into at some point. I'm pretty sure it has to do with peer dependency stuff (you can see the exact warning when running mvn start locally) that we've gotten in the past as well, but I tried fixing it and it just caused more peer dependency warnings

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -234,4 +236,16 @@ public void updateCode(String roomId, String userId, PlayerCode playerCode) {
...
     }
+
+    // Update people's socket active status
+    public void conditionallyUpdateSocketInfo(Room room, User user) {
+        Game game = currentGameMap.get(room.getRoomId());
+
+        if (game != null) {
+            log.info("Updating socket info for game {}", room.getRoomId());
+            game.setRoom(room);
+            game.getPlayers().get(user.getUserId()).setUser(user);
```
zpChris, line 247: This line actually causes a bug - see my screencast [here](https://www.loom.com/share/72644632b56843ae8b8c5566c381268c).

In short, when you create and end a game, then "Play Again", the game still exists with that room but the players aren't updated. Thus, when the game tried to get the player with that particular user ID, it returns NULL, and then the code tries to run `setUser` on NULL, causing a Null Pointer Exception. This makes the socket update not send (though oddly still shows the user as connected on the host's side) - do you know why this may be?

You _could_ add a test for this specific case, but I'll leave that up to you - I'll approve this after even if you don't add that.

In the future, when should we delete the "game" in the backend? When the "room" is deleted from the database, or right after the game ends each time (transforming it into some sort of history object in the database)?

zpChris (changes requested):
This looks great, but unfortunately requesting changes because this PR causes a bug in the "Play Again" flow - see my [screencast explanation](https://www.loom.com/share/72644632b56843ae8b8c5566c381268c). I also left a comment for this. Otherwise though, this looks great 👍

frontend/src/components/card/PlayerResultsCard.tsx
alankbi: That's fine - I'll get rid of the extra check

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -234,4 +236,16 @@ public void updateCode(String roomId, String userId, PlayerCode playerCode) {
     protected boolean isGameOver(Game game) {
         return game.getAllSolved() || (game.getGameTimer() != null && game.getGameTimer().isTimeUp());
     }
+
+    // Update people's socket active status
```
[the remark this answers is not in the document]
alankbi: I don't think so - this is the same logic as before but just refactored into a separate method to avoid duplicate code. Although that bug is very flaky so I don't actually know how often it occurs

src/main/java/com/rocketden/main/service/GameManagementService.java
alankbi: Good catch - fixed! Also added a test case for it

As for when games are deleted - I'm not sure, I think when the room is deleted makes sense, or maybe after play again is clicked. The current design needs the game object to exist in order to show the results page, but if in the future we had like history objects or something then yeah maybe we could also delete them immediately after the game ends (and history is saved).

alankbi:
Fixed the bug and added the Trello card you mentioned - I also pushed a commit/wrote tests to fix the timer bug we discovered yesterday on Heroku, so feel free to take a look at those changes and verify that it works as expected :)

src/main/java/com/rocketden/main/service/GameManagementService.java
```diff
@@ -231,7 +239,32 @@ public void updateCode(String roomId, String userId, PlayerCode playerCode) {
         liveGameService.updateCode(game.getPlayers().get(userId), playerCode);
     }
 
+    protected void endGame(Game game) {
```
zpChris, line 242: This is great design, thanks for adding this to the PR here! 😄

src/main/java/com/rocketden/main/socket/WebSocketConnectionEvents.java
```diff
@@ -81,6 +87,12 @@ public void onSocketConnected(SessionConnectedEvent event) {
             Room room = user.getRoom();
             RoomDto roomDto = RoomMapper.toDto(room);
             socketService.sendSocketUpdate(roomDto);
+
+            // If a game exists, update the room info for that game
+            gameService.conditionallyUpdateSocketInfo(room, user);
```
zpChris, line 92: Just want to make sure I understand - why is there a room socket update and a conditional game socket update back-to-back? Is this in the case that someone connects on the "Play Again" page or something, how does the frontend handle these two different types of objects?

zpChris (approved):
LGTM! ✅ 

Sorry this took a while to review, thanks for fixing the old bug in addition to some new ones 😄 Asked a clarifying question on the code to help my understanding, but these changes look great 👍

src/main/java/com/rocketden/main/socket/WebSocketConnectionEvents.java
alankbi: The previous update (room socket) only updates the room and sends the updated info to the room endpoint (used on the lobby page). So this new method makes sure to also update the game object (if a game has started) to store the updated connection info and send a socket update to the game endpoint (so people on the game page will be able to see if someone connects/disconnects as opposed to just the lobby page).

alankbi: Maybe in the future this could be simplified tho cause yeah the design is definitely not the most intuitive haha

PR #113, opened 2021-03-06
Redesign the landing page.

zpChris:
### List of Changes:

- Implement the design of the [frontpage on Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=23%3A0).
  - Create a `FloatingCircle.tsx` file that designs the frontpage floating colorful circles.
  - Track the user's mouse movement, and move the floating colorful circles in the same direction as the user's mouse.
- Change the minimum width of the site, so it can be viewed fully on mobile.
- Add the **Titillium Web** font through [an npm package](https://www.npmjs.com/package/@fontsource/titillium-web).
- Update the primary button styling.
- Set site title to "CodeJoust" and add a favicon - I used a screenshot of the purple circle, we can think of this as a placeholder as it's better than the react app one (imo).

### Notes:

- The updated primary button styling is shown on other pages - the lobby and results pages. These pages do not yet have a mockup in the Figma, and I'm happy to return the original styling for that button type if need be.
- The floating circles move around with the user's mouse; my original design idea was for them to move around randomly on their own, but this was a bit more complicated to implement. There should not be any lag, but let me know your thoughts on this design.
- The floating circles are positioned relative to the container that holds "CodeJoust ... Compete live against friends and classmates to solve coding challenges. ... etc," using `position: absolute` and the `bottom` and `left` attributes. Negative percentage styling is used here - if this is bad practice let me know.
- I removed the minimum width of the site, so it can be shown on mobile; this may screw up other pages, and I'm thinking that as we design each page, we fix the styling for mobile.
- The "Contact Us" link goes to a `/contact-us` page, which doesn't yet exist - should we have that page? Or should that just have a `mailto:` attribute; and if so, how do we avoid the annoying components an email unexpectedly popping up?
- These links [[1]](https://www.joshwcomeau.com/animation/css-transitions/) [[2]](https://blog.rocketinsights.com/getting-more-complicated-with-react-hooks/), especially the first one, were helpful when doing styling and calculating how the floating circles would move around.

### Screencast and Images:

[CodeJoust Landing Page Redesign](https://www.loom.com/share/fd603eafa4ef480685f9e6e2f29096da)

**Landing Fullscreen Frontpage**
Landing Fullscreen Frontpage

**Landing Smaller Screen Frontpage**
Landing Smaller Screen Frontpage


zpChris:
[Frontpage redesign](https://trello.com/c/KeAEDfgR/279-frontpage-redesign)

frontend/src/components/config/Theme.tsx
```diff
@@ -13,8 +15,14 @@ export const ThemeConfig: any = {
     gray: 'gray',
     lightGray: 'lightgray',
     white: 'white',
+    greenCircle: 'linear-gradient(207.68deg, #14D633 10.68%, #DAFFB5 91.96%)',
+    redCircle: 'linear-gradient(207.68deg, #DD145D 10.68%, #FFB7B7 91.96%)',
+    blueCircle: 'linear-gradient(207.68deg, #133ED7 10.68%, #B7D4FF 91.96%)',
+    pinkCircle: 'linear-gradient(207.68deg, #F25AFF 10.68%, #F5DAFF 91.96%)',
+    yellowCircle: 'linear-gradient(207.68deg, #FFC700 10.68%, #FFFDC1 91.96%)',
+    purpleCircle: 'linear-gradient(207.68deg, #9845EC 21.37%, #D9B4FF 102.65%)',
```
alankbi, line 23: Maybe rename these to just `color` without the circle part since we're using them not just for the circles? Or maybe we could move them into like a nested gradients object or something (i.e. `theme.colors.gradients.blue`)

frontend/src/components/core/Button.tsx
```diff
@@ -20,14 +21,14 @@ export const DefaultButton = styled.button`
...
+  font-size: ${({ theme }) => theme.fontSize.mediumLarge};
+  font-weight: 400;
+  background: ${({ theme }) => theme.colors.blueCircle};
   color: ${({ theme }) => theme.colors.white};
-  width: ${({ width }) => width || '32vw'};
-  height: ${({ height }) => height || '8vw'};
-  min-width: 280px;
-  min-height: 70px;
+  width: ${({ width }) => width || '16vw'};
+  height: ${({ height }) => height || '4vw'};
```
alankbi, line 29: Thanks for this - the button was definitely way too huge beforehand lol

frontend/src/views/Landing.tsx
```diff
@@ -1,21 +1,77 @@
...
-    <div>
+    <LandingPageContainer>
+      <FloatingCircle
+        color={ThemeConfig.colors.redCircle}
+        x={mousePosition.x}
+        y={mousePosition.y}
+        bottom={50}
+        left={-30}
+        size={4}
+      />
```
alankbi, line 28: Thoughts on moving this into a new say `layout/Circles.tsx` (maybe a better name though idk haha) file? That way we can reuse this background on different pages (if we choose to) by doing something like this in the `config/App.tsx` file: 

```jsx
<CustomRoute path="/" component={LandingPage} layout={CircleLayout} exact />
```

alankbi (approved):
Looks good! I have one larger suggestion which you can take a look at, but otherwise this looks beautiful :)

alankbi:
> The "Contact Us" link goes to a /contact-us page, which doesn't yet exist - should we have that page? Or should that just have a mailto: attribute; and if so, how do we avoid the annoying components an email unexpectedly popping up?
These links [1] [2], especially the first one, were helpful when doing styling and calculating how the floating circles would move around.

Yeah I think it might be good to add a contact page - doesn't have to be now, but I could imagine we could put an email subscription thing on that page (similar to on the results page) so people can subscribe for updates or leave feedback and stuff

frontend/src/components/core/Button.tsx
zpChris: Haha yeah no problem, I got the styling primarily from your [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=0%3A1), so thanks for the design 👍

frontend/src/views/Landing.tsx
zpChris: That's a good idea. I will say that this background is fairly customized to this page, but others like the "Join a Room" page and "Create a Room" page may be relatively similar to the landing page design. Done!

zpChris:
> > The "Contact Us" link goes to a /contact-us page, which doesn't yet exist - should we have that page? Or should that just have a mailto: attribute; and if so, how do we avoid the annoying components an email unexpectedly popping up?
> > These links [1] [2], especially the first one, were helpful when doing styling and calculating how the floating circles would move around.
> 
> Yeah I think it might be good to add a contact page - doesn't have to be now, but I could imagine we could put an email subscription thing on that page (similar to on the results page) so people can subscribe for updates or leave feedback and stuff

This sounds good! I added a [card in the sprint backlog](https://trello.com/c/w7gYSu4E/291-create-contact-us-page-figma-design) for this 👍

PR #114, opened 2021-03-12
Game page initial redesign 

alankbi:
### List of Changes:

* Change leaderboard colors to use gradient colors
* Randomize leaderboard colors for every game
* Style player icons and hover content/display
* Update CSS for overall game/splitter layout
* Fix CORS issue causing sockets to fail to connect on [codejoust.co](https://codejoust.co)
* Other styling changes all around:
  * Submit and run buttons
  * Console font and min/max text area sizes
  * Header bar font + spacing 
  * Backgound color has been updated

### Notes:

* I briefly tested this on Safari and Firefox - most major things should be good but there might be some stuff I missed in terms of displaying the CSS correctly
* I didn't initially begin this with the intention to finish the game page, so there are a lot of small things left to do (compared to the Figma):
  * - [ ] Console output for run/submit results
  * - [x] Monaco code editor styling + language dropdown
  * - [ ] Difficulty tag and timer location
  * - [ ] Markdown styles - looks like this is being done in the edit problem PR
  * - [ ] Drag handles between panes + varied dimensions support
  * - [ ] Gradient outlines
* Currently implementing some of these in a follow up PR

### Screencast and Images:

[Screencast of new UI (recommended 2x speed)](https://www.loom.com/share/c03b0cef222243c893db905a8ab4b57e)


frontend/src/components/card/LeaderboardCard.tsx
```diff
@@ -5,34 +5,60 @@ import { LowMarginText, SmallText } from '../core/Text';
...
+  z-index: 5;
   top: 0;
   left: 50%;
   transform: translate(-50%, 0);
   
-  width: 160px;
+  // width+padding, height+padding must match that of Content to perfect overlap
+  width: 70px;
+  min-height: 85px;
+  padding: 5px;
```
zpChris, line 56: Given this constraint, any thoughts on using a variable to match the Content and HoverBar components?

frontend/src/components/config/Theme.tsx
```diff
@@ -29,6 +30,7 @@ export const ThemeConfig: any = {
     xSmall: '0.4rem',
     small: '0.6rem',
     mediumSmall: '0.8rem',
+    xMediumSmall: '0.9rem',
```
zpChris, line 33: Perhaps just call this "medium"?

src/main/java/com/rocketden/main/config/WebSocketConfig.java
```diff
@@ -27,6 +27,6 @@ public void configureMessageBroker(MessageBrokerRegistry config) {
 
     @Override
     public void registerStompEndpoints(StompEndpointRegistry registry) {
-        registry.addEndpoint(BaseRestController.BASE_SOCKET_URL + "/join-room-endpoint").withSockJS();
+        registry.addEndpoint(BaseRestController.BASE_SOCKET_URL + "/join-room-endpoint").setAllowedOrigins("*").withSockJS();
```
zpChris, line 30: Is there any concern with allowing any origins? I'm ok with this but figured it'd be good to double-check.

frontend/src/components/core/Button.tsx
```diff
@@ -94,7 +95,15 @@ export const ProblemIOTypeButton = styled(DefaultButton)<ProblemIOTypeButtonProp
...
 export const SmallButton = styled(DefaultButton)`
   color: ${({ theme }) => theme.colors.white};
-  background-color: ${({ theme }) => theme.colors.blue};
-  padding: 0.5rem 1rem;
+  background: ${({ theme }) => theme.colors.gradients.blue};
+  font-size: ${({ theme }) => theme.fontSize.xMediumSmall};
+  padding: 0;
   margin: 0.4rem;
+  width: 5.8rem;
+  height: 2.1rem;
```
zpChris, line 103: Thoughts on rounding these numbers slightly? Not sure how you got these decimals, maybe from the Figma, but it might still look ok just rounded to the nearest 0.5 value.

zpChris (approved):
LGTM! ✅ I left some nits and general comments, and made a [screencast with some potential changes](https://www.loom.com/share/463917ffa7b34cc4b15f70a0d5e41224) - these are all small nits though, so all or none of them implemented is completely fine.

src/main/java/com/rocketden/main/config/WebSocketConfig.java
alankbi: [This link](https://stackoverflow.com/questions/12001269/what-are-the-security-risks-of-setting-access-control-allow-origin) seems to provide a good overview - thoughts? Seems like at some point if we added proper login/credential stuff it may be worth disabling (at least to be safe), but for now this comment seems reassuring: 

> Imagine what a user would see if they accessed the resource in an incognito window. If you're happy with everyone seeing this content (including the source code the browser received), it's safe to add Access-Control-Allow-Origin: *.

frontend/src/components/core/Button.tsx
alankbi: Haha yeah I might have overengineered this - as a compromise I can change it to 6 and 2.25 if that helps (6-2 and 6-2.5 I felt were just a tad bit too long/tall respectively)

frontend/src/components/config/Theme.tsx
alankbi: Sure sounds good - as a side note this will put medium as a smaller size than default, followed by mediumLarge

frontend/src/components/card/LeaderboardCard.tsx
alankbi: Hm interesting, do you know how this would work? Like would it be a constant CSS variable at the top that you could edit to modify both at the same time?

alankbi:
Implemented the nits and some of the UI suggestions (below for feedback, lmk if I missed any of them): 

1. For many players, I added a scroll effect (I think you suggested something different so lmk what you think about this)
Scrolling players

2. For the absolute style buttons, it works great for normal sizes, but this is what happens for small widths or heights (open to suggestions). Also, I increased the width of the input boxes

Small width buttons below each other

Buttons over text

3. I committed both of those changes above, but for the last one (padding for the editor container), I made the change but this is the result which I'm not too sure about haha - it could be just the shadow messing with it which I got rid of in my next branch, so maybe we could revisit it there and see if it looks better in that situation? 

Editor with padding

zpChris:
> 1. For many players, I added a scroll effect (I think you suggested something different so lmk what you think about this)
 
Scroll effect looks great! Nice 👍 

> 2. For the absolute style buttons, it works great for normal sizes, but this is what happens for small widths or heights (open to suggestions). Also, I increased the width of the input boxes

Yeah that's tough, but expected; thanks for increasing the width of those boxes, and I'm not entirely sure of a fix for that - honestly there's simply not enough room in the images shown for the buttons and the inputs, so either we set a restraint on how much you can drag the splitter layout, or we just accept that this would be brought about by the user.

> 3. I committed both of those changes above, but for the last one (padding for the editor container), I made the change but this is the result which I'm not too sure about haha - it could be just the shadow messing with it which I got rid of in my next branch, so maybe we could revisit it there and see if it looks better in that situation?

I think that's possible, so sounds good - I'd be ok with going back too tbh, especially because there's a padding mismatch between the top and bottom of the container, but I'll leave that up to you.

frontend/src/components/config/Theme.tsx
zpChris: Yeah I'm ok with that if you are, sounds good 👍

src/main/java/com/rocketden/main/config/WebSocketConfig.java
zpChris: That's a good source, thanks for including that - yeah I'm absolutely ok with keeping this for now, and potentially re-evaluating down the line 👍

frontend/src/components/core/Button.tsx
zpChris: That works by me haha, also this is just a nit so if you feel it works best with these numbers, no problem.

frontend/src/components/card/LeaderboardCard.tsx
zpChris: Yeah, you could add a constant `contentWidth` at the top, and then set the internal widths using `${contentWidth}px` (in addition to whatever additions or subtractions you want within the "{" and "}"). That way we can update this to change both components, so they should always be in sync.

zpChris (approved):
LGTM! Thanks for making those changes, looks great 👍 

My only nit, which I'm not sure has any easy solution, is that the "scrollable" aspect of the players at the top isn't particularly obvious when a new user first joins. The image below shows this - it's not super clear that there are other players in the game and that the top component can be scrolled horizontally.

Scrollable Player List

However, this is a small note, and overall the initial redesign of the game page looks great - we can make further changes / nits in future PRs 👍

alankbi:
Was able to get a shadow to show when overflow exists using [this link](https://stackoverflow.com/questions/57682185/horizontal-scrolling-shadows-on-top-of-colored-elements):

Both sides shadow

One side shadow

Hopefully this should help with that issue - the design for this isn't perfect (the shadow is too dark on Safari for example) but it at least seems to work properly on Chrome/Safari/Firefox. Since this PR is approved I'll merge, but if you have style concerns feel free to leave it on the next PR (which builds directly on this one so I can easily change it there).

alankbi:
[Player colors and general game page UI](https://trello.com/c/6sDKygXu/282-player-colors-and-general-game-page-ui)

frontend/src/components/card/LeaderboardCard.tsx
alankbi: Just realized I totally forgot about this comment lmao - maybe I can implement it in the current open PR if you want to resurface it there?

PR #115, opened 2021-03-12
Create the edit problem page design.

zpChris:
### List of Changes:

- Create the Edit Problem page design, as shown on the [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=0%3A1).
- Create new layout for the problem pages; this uses a "Minimal Header" so the content of the page is positioned higher.
- Change difficulty button colors, which are also reflected on the lobby page (shown in the images section).

### Notes:

- The design could be a lot better on smaller-sized screens; the main improvement that could be made is moving the right-side "Options" panel to shift below, as well as perhaps expanding the width to full-size at a certain width.
- I had trouble converting the Markdown description to use the Titillium-Web font, so I left that (and the title) as the default font - if you know how to change this @alankbi, please let me know. (It doesn't look awful as-is, but would be better with the font imo.)
- The "Delete" button on the Test Cases is slightly askew - I tried to fix this multiple times to no avail, so again @alankbi if you know how to fix this, let me know. This is a really small UI mismatch though, so I think leaving it would be fine too.
- The ["drag / reorder" functionality](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib?node-id=23:52#67601407) is implemented at #117.
- I tried to create the gradient border as shown on the [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=0%3A1), but was unable to make that look nice; I did however just find [a link](https://css-tricks.com/gradient-borders-in-css/) with a viable solution (ironically not in the article itself but the advertisement on the right, image pasted below), but given that a lot has been implemented here I think this change could be attempted in a new PR.
Gradient Border on Image

### Screencast and Images:

[Design the Edit Problem Page](https://www.loom.com/share/e352bebc772547b8b3c289bc97e620f6) (5 min - 2x speed recommended)

**Edit Problem Page Design**
Edit Problem Top of Page Design
Edit Problem Bottom of Page Design

**Small Addition - Updated Difficulty Buttons on Lobby Page**
Small Addition - Updated Difficulty Buttons on Lobby Page


zpChris:
[Design edit problem page](https://trello.com/c/C3zvkz4H/294-design-edit-problem-page)

frontend/src/api/Problem.ts
```diff
@@ -38,6 +38,33 @@ export enum ProblemIOType {
...
+    case ProblemIOType.ArrayString:
+      return 'ArrayString';
+    case ProblemIOType.ArrayInteger:
+      return 'ArrayInteger';
+    case ProblemIOType.ArrayDouble:
+      return 'ArrayDouble';
+    case ProblemIOType.ArrayCharacter:
+      return 'ArrayCharacter';
+    case ProblemIOType.ArrayBoolean:
+      return 'ArrayBoolean';
```
alankbi, line 62: Thoughts on if we have a custom map to display names if it would look better to do something like Double[]?

frontend/src/components/core/Button.tsx
```diff
@@ -22,8 +23,8 @@ export const DefaultButton = styled.button`
 
 export const PrimaryButton = styled(DefaultButton)<any>`
   font-size: ${({ theme }) => theme.fontSize.mediumLarge};
-  font-weight: 400;
-  background: ${({ theme }) => theme.colors.gradients.blue};
+  font-weight: ${({ fontWeight }) => (fontWeight || 400)};
+  background: ${({ color }) => color};
```
alankbi, line 27: This might be personal preference, but I feel like if we have multiple styling options it might be good to have multiple containers like so: 

```jsx
export const PrimaryButton = styled(DefaultButton)`
  ...
`;

export const RedButton /* or DangerButton like Bootstrap does it */ = styled(PrimaryButton)`
  color: red;
`;

...
```

Mainly modeling this idea off of how Bootstrap does it, and also potentially to save on lines in the other files (so we don't have to specify a bunch of options everywhere or manually import ThemeConfig outside of the withTheme thing)

frontend/src/components/core/Container.tsx
```diff
@@ -70,6 +74,12 @@ export const MainContainer = styled.div`
   width: 80%;
 `;
 
+export const TopContainer = styled.div`
```
alankbi, line 79: nit: rename this and the other TopX components to ProblemX (since for not it's only being used for those pages I think)

frontend/src/components/core/Link.tsx
```diff
@@ -1,22 +1,29 @@
...
 import styled from 'styled-components';
-import { PrimaryButton } from './Button';
+import { PrimaryButton, SmallButton } from './Button';
 
 // Wrap a button inside of a Link to get the styling of a button
 const createButtonLink = (Button:any, props:any) => {
-  const { to, children, ...rest } = props;
+  const {
+    to,
+    style,
```
alankbi, line 10: nit: If possible could we keep most styling to the styled-components (for one off styles another option could also be just defining an XContent component at the top of the file)? Mainly cause it might be easier to forget about inline styles or have duplicate ones that all need to be changed for e.g.

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -2,26 +2,56 @@ import React, { useState } from 'react';
...
+  padding: ${({ padding }) => padding || '1rem'};
+  border-radius: 10px;
   box-shadow: 0 -1px 4px rgba(0, 0, 0, 0.12);
   background: ${({ theme }) => theme.colors.white};
 `;
 
-const Content = styled.div`
-  padding: 10px;
+const MarginRightContainer = styled.div<any>`
+  margin-right: ${({ marginRight }) => marginRight}%;
```
alankbi, line 54: nit: Create simple types to replace any

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -142,141 +173,218 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
-      />
+    <>
+      <MainContent>
+        <FlexBareContainer>
+          <SmallHeaderText>Problem</SmallHeaderText>
+          <SmallButtonLink
+            color={ThemeConfig.colors.gradients.gray}
+            onClick={() => onClick(newProblem)}
+            to="/problems/all"
+            style={{ marginLeft: 'auto' }}
```
alankbi, line 184: Ties into my comment before but thoughts on moving some of the styling to the styled components? I know there's like a ton of small styling changes here and there which could make it difficult tho - not sure the best approach to something like that - maybe like a decent amount of styled components within this file at the top, idk

alankbi: For this case specificially though it might also help to put the two buttons in a new div together, and then make that div flex-right or something like that, that way you might not have to have an edge case style like marginLeft

alankbi: Maybe like a liberal use of styled-components specific to the type (e.g. TestCaseHeader) rather than style (PaddingTopContainer) might also help with some of the styling edge cases - might cause some extra components at the top though so I guess it's a tradeoff

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -142,141 +173,218 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
+            }
+          </PrimarySelect>
 
-      {loading ? <Loading /> : null}
-      {error ? <ErrorMessage message={error} /> : null}
-    </Content>
+          {editMode
+            ? (
+              <>
+                <LowMarginMediumText>Danger Zone</LowMarginMediumText>
```
alankbi, line 374: I like that danger zone haha

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -142,141 +173,218 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
+                <SettingsContainer>
+                  <FlexBareContainer>
+                    <MarginRightContainer marginRight="10">
+                      <Text style={{ marginTop: 0 }}>Input</Text>
+                      <FixedTextArea
+                        value={newProblem.testCases[index].input}
+                        onChange={(e) => {
+                          const current = newProblem.testCases[index];
+                          handleTestCaseChange(index, e.target.value,
+                            current.output, current.hidden, current.explanation);
```
alankbi, line 224: This might be a later on discussion but will we be using the explanation field anytime in the near future? I'm imagining when adding TeamsCode problems with multiple test cases per input that the explanations might become kind of redundant, like maze problem -> "this is the correct path" or find max -> "X is the max in the array" or palindrome -> "the given string is/is not a palindrome". Although it could be useful down the line with a teacher-oriented route for certain types of beginner problems - that might be a bit far in the future though from what I'm imagining

It's an optional field though so I'm def fine for keeping it as is for now tho 👍

alankbi (changes requested):
This looks great! And thanks for doing the dragging thing too on the other branch

Requesting changes mainly for some of the styled-components stuff - but I'm also fine revisiting those in a later PR since it looks good + we're on an MVP timeline + the problem display page is naturally complex with 10+ different fields and options and all

As for minor style suggestions, I left mine in the [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=0%3A1)

alankbi:
> The design could be a lot better on smaller-sized screens; the main improvement that could be made is moving the right-side "Options" panel to shift below, as well as perhaps expanding the width to full-size at a certain width.

Yeah that's a good idea - although the current mobile UI when I tested it didn't seem that bad (plus I wonder how much we're really supporting mobile for a coding site)

> I had trouble converting the Markdown description to use the Titillium-Web font, so I left that (and the title) as the default font - if you know how to change this @alankbi, please let me know. (It doesn't look awful as-is, but would be better with the font imo.)

Yeah this might be another case of the splitter layout issue, we'll probably have to find a workaround in order to get that working at some point

> The "Delete" button on the Test Cases is slightly askew - I tried to fix this multiple times to no avail, so again @alankbi if you know how to fix this, let me know. This is a really small UI mismatch though, so I think leaving it would be fine too. 👍 


> The "drag / reorder" functionality is implemented at #117. 🔥 


> I tried to create the gradient border as shown on the Figma, but was unable to make that look nice; I did however just find a link with a viable solution (ironically not in the article itself but the advertisement on the right, image pasted below), but given that a lot has been implemented here I think this change could be attempted in a new PR.

Thanks for finding that haha - I had the same issue for the game page but didn't look into it past a quick google search, so yeah that might be interesting to look into

frontend/src/api/Problem.ts
zpChris: I'd be slightly worried about these not exactly matching the style of other languages, since an identifier like `Double[]` seems specific to Java? I could be wrong on this, so I'll leave this open.

frontend/src/components/core/Container.tsx
zpChris: Done! ✅

frontend/src/components/core/Link.tsx
zpChris: Yeah I think that's a good idea, I went a little heavy on the inline styling on this PR - I'll update this to use styled components.

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Yes for sure, I agree with this - I converted all of the elements to styled components rather than inline CSS, let me know what you think 👍

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Yeah haha, it's on GitHub and I think other sites have adopted it too (like Gimkit actually lol) 😁

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Yeah my thinking is to keep it now, as I do think it could be helpful for teachers. There is another idea I actually wanted to bring up - thoughts on having a "name" for each test case? This would be optional, but essentially for any hidden test case, it would allow the user to see the name of the test case and perhaps troubleshoot from there - I think this is how Gradescope does it, for example.

Not a super important feature, but an easy one to add - though potentially tedious for the problem maker. Let me know your thoughts on this (these could be auto-generated too imo, if you have the same set of inputs every time).

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: I got rid of the styled components that necessitated this as I simply creates new components for this use 👍

frontend/src/components/core/Button.tsx
zpChris: Agreed, that makes sense - changed back 👍

frontend/src/api/Problem.ts
alankbi: I think most languages use brackets in some form but that does make sense - if we're going that route though, should we be concerned about languages like Python that use lists instead of arrays? e.g. Python lists would be analogous to Java Lists rather than arrays, which idk if that would affect our naming in any way

frontend/src/components/core/Button.tsx
zpChris: Agreed - done ✅

frontend/src/components/problem/ProblemDisplay.tsx
alankbi: Hm yeah that could be a good idea and also a good beginner story for new members. I agree though that it would be tedious for the problem maker - the way I'm thinking about it is that depending on the path we go we'd either go all in on name + explanation (e.g. Gradescrope route as you mentioned) or perhaps leave both out (e.g. LeetCode-style contests)

In the meantime though we could also do something with test case # kind of like how LeetCode does, so if you fail one it'll show like 45/90 passed, and then show the 46th test case or something

frontend/src/api/Problem.ts
zpChris: Yeah that's true - what's your preference on this?

(Also if you're updating this right now, feel free to just change this and other nits if you want / are free)

alankbi: Idk I don't feel strongly either way (especially since this page isn't public facing yet) - so I'm open to leaving as is or changing it, whatever you think makes sense

zpChris:
I forgot to update the button styling to inverted (white background, colored text) for some of the buttons as noted in the Figma, so I'll do that now 👍

frontend/src/api/Problem.ts
zpChris: Ok, I think I'll leave this but happy to come back to it later.

zpChris:
> I forgot to update the button styling to inverted (white background, colored text) for some of the buttons as noted in the Figma, so I'll do that now 👍

Here is the updated styling for those buttons:

Inverted Button Styling

Let me know if this looks good, or if I should change it back 👍

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: That all sounds good, could definitely be a good first story and we could definitely show #'s for test cases.

frontend/src/components/core/Button.tsx
```diff
@@ -53,43 +53,82 @@ export const TextButton = styled.button<ThemeType>`
   }
 `;
 
+export const RedTextButton = styled(TextButton)`
+  color: ${({ theme }) => theme.colors.red2};
+`;
+
+export const GrayTextButton = styled(TextButton)`
+  color: ${({ theme }) => theme.colors.regrayd2};
```
alankbi, line 61: Potential typo haha

```suggestion
  color: ${({ theme }) => theme.colors.gray};
```

frontend/src/components/core/Link.tsx
```diff
@@ -15,8 +23,29 @@ const createButtonLink = (Button:any, props:any) => {
...
+      {children}
+    </Button>
+  );
+
+  return <AutoLeftMarginLink to={to}>{button}</AutoLeftMarginLink>;
+};
+
 export const PrimaryButtonLink = (props:any) => createButtonLink(PrimaryButton, props);
 
+export const InvertedSmallButtonLinkAutoLeftMargin = (props: any) => createButtonLinkAutoLeftMargin(
```
alankbi, line 44: That's quite the mouthful lol

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -142,141 +233,204 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
-            return (
-              <ProblemIOTypeButton
-                onClick={() => handleInputChange(index,
-                  newProblem.problemInputs[index].name, inputType)}
-                active={inputType === newProblem.problemInputs[index].type}
+        </SettingsContainerHighPadding>
+
+        {editMode
+          ? (
+            <>
```
alankbi, line 269: Out of curiosity, what does this syntax do?

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -142,141 +233,204 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
+          <SmallHeaderText>Problem</SmallHeaderText>
+          <InvertedSmallButtonLinkAutoLeftMargin
+            onClick={() => onClick(newProblem)}
+            to="/problems/all"
+          >
+            Back
+          </InvertedSmallButtonLinkAutoLeftMargin>
+          <SmallButton
+            onClick={() => onClick(newProblem)}
+          >
```
alankbi, line 248: Since a bit of extra work was necessary to get the auto left margin to work (namely the createButtonLink function), maybe an alternative simpler way (to cut down on extra code) would be if we put these two buttons into a div and make that aligned/flex/margin right? Then the back button could just become an InvertedSmallButtonLink using the original creatButtonLink function

I haven't tested this though and sometimes flex stuff can be a hassle, so nbd if we keep this as-is for now though

alankbi (approved):
Looks good! Like 1-2 more small nits but this looks great I think

frontend/src/components/core/Button.tsx
zpChris: Nah it was just the second "regrayd" color 😁 Thanks for the fix haha

frontend/src/components/core/Link.tsx
zpChris: Yeah it's such a specific element haha - not sure what to do about that one, I would style it with a more specific name directly in `ProblemDisplay.tsx`, but I have to create this wrapper with the link. We can change this name if needed, np 👍

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: The first is a ternary operator which I'm sure you know, the `<>` is a [React Fragment](https://reactjs.org/docs/fragments.html#short-syntax) so you can return a list of children without adding extra nodes to the DOM.

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: That's definitely much smarter - thanks! Updated by creating a wrapper container with `margin-left: auto` 👍

PR #116, opened 2021-03-13
Enable the ordering of the test cases.

zpChris:
### List of Changes:

- Use the [react-beautiful-dnd (drag-and-drop)](https://github.com/atlassian/react-beautiful-dnd) framework to implement the test case ordering referred to on the [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib?node-id=23:52#67601407).
- Add a numbering to the test cases on the editing frontend.
- Add an id to the problem test cases on the frontend separate from the backend test case id, specifically used for identifying the keys of each of the Draggable elements - see further explanation [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md#keys-for-a-list-of-draggable-).

### Notes:

- The code is based on [this sandbox](https://codesandbox.io/s/k260nyxq9v?file=/index.js:0-2887), and in-depth explanation is available [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md).
- I initially attempted to implement this with [Shopify's Draggable framework](https://github.com/Shopify/draggable), using the type is beta-11 which is discussed [here](https://github.com/Shopify/draggable/issues/105), but could not find a working React TypeScript example from which to build.
- The code is getting very bloated in the `ProblemDisplay.tsx` file; it may be smart to split this up, but I had some trouble doing so - let me know if I should work on this more 👍 

### Screencast and Images:

[Problem Page Ordering Test Cases](https://www.loom.com/share/f10b2290c04f486798d202c94b6bee09)

**Ordering of Test Cases on Problem Page**
Ordering of Test Cases on Problem Page


zpChris:
This PR was closed because there are commits in here that accidentally install and uninstall an entire `node_modules` folder in the wrong place, without gitignore - to preserve accurate line counts, I closed this PR and opened #117.

PR #117, opened 2021-03-13
Enable the ordering of the test cases.

zpChris:
### List of Changes:

- Use the [react-beautiful-dnd (drag-and-drop)](https://github.com/atlassian/react-beautiful-dnd) framework to implement the test case ordering referred to on the [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib?node-id=23:52#67601407).
- Add a numbering to the test cases on the editing frontend.
- Add an id to the problem test cases on the frontend separate from the backend test case id, specifically used for identifying the keys of each of the Draggable elements - see further explanation [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md#keys-for-a-list-of-draggable-).

### Notes:

- The code is based on [this sandbox](https://codesandbox.io/s/k260nyxq9v?file=/index.js:0-2887), and in-depth explanation is available [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md).
- I initially attempted to implement this with [Shopify's Draggable framework](https://github.com/Shopify/draggable), using the type is beta-11 which is discussed [here](https://github.com/Shopify/draggable/issues/105), but could not find a working React TypeScript example from which to build.
- The code is getting very bloated in the `ProblemDisplay.tsx` file; it may be smart to split this up, but I had some trouble doing so - let me know if I should work on this more 👍 

### Screencast and Images:

[Problem Page Ordering Test Cases](https://www.loom.com/share/f10b2290c04f486798d202c94b6bee09)

**Ordering of Test Cases on Problem Page**
Ordering of Test Cases on Problem Page


zpChris:
[Implement draggable test cases](https://trello.com/c/ArhghhdW/295-implement-draggable-test-cases)

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -71,6 +87,24 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
+    const newTestCases: TestCase[] = reorder(
+      newProblem.testCases,
+      result.source.index,
+      result.destination.index,
+    );
+
+    const updatedProblem: Problem = newProblem;
+    updatedProblem.testCases = newTestCases;
+
+    setNewProblem(updatedProblem);
```
alankbi, line 105: Potential alternative based on [this post](https://stackoverflow.com/questions/37755997/why-cant-i-directly-modify-a-components-state-really) on directly modifying React state (btw I didn't test this suggestion so some of the syntax might be wrong):

```suggestion
    setNewProblem({ ...newProblem, testCases: newTestCases });
```

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -211,74 +246,103 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
+                      ref={providedDroppable.innerRef}
+                    >
+                      {newProblem.testCases.map((testCase, index) => (
+                        <Draggable key={testCase.id} draggableId={testCase.id} index={index}>
+                          {(providedDraggable) => (
+                            <div
+                              ref={providedDraggable.innerRef}
+                              {...providedDraggable.draggableProps}
+                              {...providedDraggable.dragHandleProps}
+                              style={providedDraggable.draggableProps.style}
```
alankbi, line 263: What do these providedDroppable/providedDraggable innerRef/props/style things do?

frontend/src/views/ProblemPage.tsx
```diff
@@ -26,6 +27,10 @@ function ProblemPage() {
   useEffect(() => {
     getSingleProblem(params.id)
       .then((res) => {
+        res.testCases.forEach((testCase) => {
+          // eslint-disable-next-line no-param-reassign
+          testCase.id = generateRandomId();
+        });
```
alankbi, line 33: If the testCases require an id, maybe we could add a quick `private String testCaseId =  UUID.randomUUID().toString();` to the backend and use that instead? That might require an extra line or two in the mapper and mapper test files, but IIRC those kinds of changes don't require an extra test so hopefully it wouldn't be too difficult to implement

alankbi (approved):
Looks great! Thanks for finding the airbnb draggable package, I'm surprised how elegantly it works

I left only a couple suggestions, and then for the UI, the main thing would be for the top-right numbered corner, I might suggest potentially unbolding it, maybe adding a `#` symbol in front, replacing with one of those dragging symbols, or lowering the size of it to see if any of those look good?

Here's an example of some of those (I also played around with some of the margin/border weights in this screenshot but I think that ties more into the EditProblemDesign PR than this one): 

Different test case top right styles

frontend/src/views/ProblemPage.tsx
zpChris: Ah, the reason I'm not doing that is because we need to generate an ID for the test cases that exist only on the frontend before they are saved as well - this is because a specific ID is needed to make up the Draggable keys, as described [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md#keys-for-a-list-of-draggable-). We could use a combination of frontend and backend solutions, but that seems superfluous if a frontend solution would work for both - thoughts?

zpChris:
> Looks great! Thanks for finding the airbnb draggable package, I'm surprised how elegantly it works
> 
> I left only a couple suggestions, and then for the UI, the main thing would be for the top-right numbered corner, I might suggest potentially unbolding it, maybe adding a `#` symbol in front, replacing with one of those dragging symbols, or lowering the size of it to see if any of those look good?
> 
> Here's an example of some of those (I also played around with some of the margin/border weights in this screenshot but I think that ties more into the EditProblemDesign PR than this one):
> 
> Different test case top right styles

Glad it works well 😁 🎉 

I updated the UI to match your suggestion and the size / formatting of the text in the box 👍 I'll plan to merge this PR after merging #115!

Update Test Case Numbering

frontend/src/views/ProblemPage.tsx
zpChris: Since this is approved I'll merge this now, but we can make changes on a future PR if a backend implementation would still be helpful @alankbi.

PR #118, opened 2021-03-16
Style create/join pages and improve game page UI

alankbi:
### List of Changes:

* Apply circle background and input styling to join and create pages
* Refactor circle background to use absolute positioning based on entire screen
  * This was to fix the issue on create/join pages where the popup text at the bottom changes the content height and causes the circles to jump down and up
* Update language select design to be compatible with Safari/Firefox
  * Unfortunately, there's no easy way to keep the arrow if we want it to look nice on those other browsers
* Apply custom styling to markdown and code editors to improve the UI (and add a small footer)
* Edit console to display run/submit output in a nice format

### Notes:

Couple things I didn't add (actually a lot of things):

* Difficulty tag as seen in the Figma
  * Since the colored tags are implemented in #115, I figured whichever of the two PRs merges last could just push an extra commit to add the tag below the problem title
* Drag icons between panels 
  * BinarySearch does this by styling a div inside the separator panel. However, idk how we'd be able to do this without doing something sketchy like using JS to like manually inject a custom-styled div inside the splitter library's separator class
* Timer absolutely positioned at top right of problem pane (as seen if Figma)
  * Personally I think it looks good where it is now (header bar), and moving it to the problem pane might cause the same absolutely positioned overlap errors (when the problem name is long) that we currently get with the run/submit code buttons
* Gradient outlines for the textareas
  * I don't think this is actually 100% feasible with textareas or input tags - the two things I got to work either didn't support border radii or didn't support padding (see below):

No padding inside textarea with gradient outline

Ultimately though if there are any things among that list that we think we should add for the MVP let me know and I can add that to this PR.

### Screencast and Images:

[Screencast (2x speed recommended)](https://www.loom.com/share/e27554e3130b4969aa5d67d39a6e489c)

PR #119, opened 2021-03-18
Create the Contact Us page.

zpChris:
### List of Changes:
- Add the Contact Us page.
- Add "copy email" functionality and animation.

### Notes:
- I had the email copy functionality rather than a `mailto:` link because I find that "open the email client" functionality annoying, and I think [others do too](https://ihatemailto.com/).
- The "copy email" setup could be used on #118 as described at [this discussion](https://github.com/rocketden/main/pull/118#discussion_r595580414), and for the lobby page "copy room link" feature.
- I updated the links to never turn purple and open in external tabs, but didn't style them further as I think that the link color actually fits.
- The [original implementation](https://stackoverflow.com/questions/39501289/in-reactjs-how-to-copy-text-to-clipboard) for copying to the clipboard, so I used the [`react-copy-to-clipboard` npm package instead](https://www.npmjs.com/package/react-copy-to-clipboard).
- I originally tried to add the "copy" icon through [npm](https://www.npmjs.com/package/material-design-icons), but couldn't get it to work, so I added [Google Fonts simply to the `index.html` file](https://developers.google.com/fonts/docs/material_icons#setup_method_1_using_via_google_fonts).
- The "Email copied!" button is never actually removed from the DOM, as I animate it in and out view instead of directly removing it. Thoughts on this @alankbi - is there an easier way to remove it once it's out of view?

### Screencast and Images:

[Screencast of the Contact Us Copy Page Functionality](https://www.loom.com/share/e0c79be40f3b431e81dfbc2752099d33)

**Contact Us Page**
Screen Shot 2021-03-17 at 6 04 44 PM

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,6 +19,7 @@
     "@types/stompjs": "^2.3.4",
     "@types/styled-components": "^5.1.2",
     "axios": "^0.21.1",
+    "copy-to-clipboard": "^3.3.1",
     "monaco-editor": "^0.21.2",
     "monaco-editor-webpack-plugin": "^2.0.0",
     "react": "^16.13.1",
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <meta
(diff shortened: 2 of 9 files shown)


zpChris:
[Contact Us Page](https://trello.com/c/jeAQwGxa/298-contact-us-page)

alankbi (approved):
Looks good to me!

I'm fine with the copy icon staying there - especially since it disappears once you leave the page. If you really wanted tho then maybe something with setTimeout idk

zpChris:
> Looks good to me!
> 
> I'm fine with the copy icon staying there - especially since it disappears once you leave the page. If you really wanted tho then maybe something with setTimeout idk

Yeah I don't think I can use `setTimeout` because some issues arise with clicking and closing the copy text multiple times in a row - this sounds good though, I'll leave that in the DOM.

PR #120, opened 2021-03-19
Design the new lobby page.

zpChris:
### List of Changes:
- Update the lobby page to match the [Figma design](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=0%3A1).
- Create a slider component for room duration.
- Add custom redirect from `/play?[querystring]` to `/game/join?[querystring]`.
- Add disabled button styling.

### Notes:
- Question: Should the "Room ID" cards have a white or transparent background? Not sure of my preference, so I wanted to check and see your thoughts @alankbi.
- The current user's name is bolded - there may be a better way to symbolize this. We could also allow hovers for non-hosts, which could hold additional information like the active status, the host, and which user is "me".
- I got the slider from [this w3schools tutorial](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_js_rangeslider_round) - I couldn't figure out how to style the background for only the first portion, so the entire background is blue. I was also going to have it in increments of 5, but since it starts at 1, this wouldn't work well without some tweaking.
- I tried to make the page not have any scroll, but this was difficult; I may be able to do this by putting the site into vertical flex components, but I think that should be in another PR instead of this one.
- I struggled with the overflow problem on the players container, as I want the player cards to be `overflow: auto` but the hover action item cards to go out of the box. I ended up placing these action cards _below_ the player cards so they would simply show up within the overflow container, as I couldn't solve it despite [research and attempts](https://css-tricks.com/popping-hidden-overflow/), but in a future PR we could update this.
- I didn't add several of the features that are in the [Figma](https://www.figma.com/file/RqMbCSJslbRDilJScKCVib/RocketDen?node-id=0%3A1) as the backend implementation for those are not yet present.
- Similar to the problem page, this doesn't work _great_ with smaller screen sizes (especially very small ones), but it doesn't completely break. There may be a PR in the future to fix the mobile styling for both of those pages.

### Screencast and Images:

[Screencast of the New Lobby Page Design](https://www.loom.com/share/b4adc77f85484aa581320cca357f7b4f)

**New Lobby Page Design**
New Lobby Page Design

### DIFF
--- a/frontend/src/components/card/HostActionCard.tsx
+++ b/frontend/src/components/card/HostActionCard.tsx
@@ -6,13 +6,13 @@ import { User } from '../../api/User';
 const Content = styled.div`
   // Center div above parent
   position: absolute;
-  top: 0;
+  top: -5px;
   left: 50%;
   transform: translate(-50%, 0);
   
   width: 150px;
   background-color: ${({ theme }) => theme.colors.white};
-  border-radius: 5px 5px 0 0;
+  border-radius: 0.5rem;
   box-shadow: 0 -1px 4px rgba(0, 0, 0, 0.12);
   
   height: 20px;
--- a/frontend/src/components/card/HostActionCard.tsx
+++ b/frontend/src/components/card/HostActionCard.tsx
@@ -1,25 +1,40 @@
 import React from 'react';
 import styled from 'styled-components';
-import { SmallActionText } from '../core/Text';
+import { SmallActionHeaderText, SmallActionText } from '../core/Text';
 import { User } from '../../api/User';
 
-const Content = styled.div`
+type PlayerIconType = {
+  isActive: boolean,
+};
+
+const Content = styled.div<PlayerIconType>`
   // Center div above parent
   position: absolute;
-  top: -5px;
+  top: ${({ isActive }) => (isActive ? '-115px' : '-90px')};
   left: 50%;
   transform: translate(-50%, 0);
   
-  width: 150px;
+  width: 120px;
   background-color: ${({ theme }) => theme.colors.white};
   border-radius: 0.5rem;
   box-shadow: 0 -1px 4px rgba(0, 0, 0, 0.12);
   
-  height: 20px;
+  height: ${({ isActive }) => (isActive ? '90px' : '65px')};
   padding: 8px;
-  
-  // -(height + 2 * padding - 3px)
-  margin-top: -36px;
+`;
+
+const ActionCardActiveIcon = styled.div<PlayerIconType>`
+  display: inline-block;
+  margin-right: 5px;
+  background: ${({ theme, isActive }) => (isActive ? theme.colors.gradients.green : theme.colors.gradients.red)};
+  border-radius: 0.5rem;
+  height: 0.5rem;
+  width: 0.5rem;
+`;
+
+const ActionCardSeparator = styled.hr`
+  border: none;
+  border-top: 1px solid ${({ theme }) => theme.colors.text};
 `;
 
 type HostActionCardProps = {
@@ -35,7 +50,10 @@ function HostActionCard(props: HostActionCardProps) {
   } = props;
 
   return (
-    <Content>
+    <Content isActive={userIsActive}>
+      <ActionCardActiveIcon isActive={userIsActive} />
+      <SmallActionHeaderText>{userIsActive ? 'active' : 'inactive'}</SmallActionHeaderText>
+      <ActionCardSeparator />
       {
         // Only show the make host button if user is active
         userIsActive
--- a/frontend/src/components/card/HostActionCard.tsx
+++ b/frontend/src/components/card/HostActionCard.tsx
@@ -10,7 +10,7 @@ type PlayerIconType = {
 const Content = styled.div<PlayerIconType>`
   // Center div above parent
   position: absolute;
-  top: ${({ isActive }) => (isActive ? '-115px' : '-90px')};
+  top: 55px;
   left: 50%;
   transform: translate(-50%, 0);
   
@@ -21,6 +21,9 @@ const Content = styled.div<PlayerIconType>`
   
   height: ${({ isActive }) => (isActive ? '90px' : '65px')};
   padding: 8px;
+
+  z-index: 2;
+  text-align: center;
 `;
 
 const ActionCardActiveIcon = styled.div<PlayerIconType>`
--- a/frontend/src/components/core/RangeSlider.tsx
+++ b/frontend/src/components/core/RangeSlider.tsx
@@ -45,4 +45,14 @@ export const Slider = styled.input.attrs((props: SliderRange) => ({
     background: linear-gradient(180deg, #133ED7 0%, #90BDFF 100%);
     cursor: pointer;
   }
+
+  &:disabled {
+    cursor: default;
+    opacity: 0.5;
+
+    &:hover {
+      cursor: default;
+      opacity: 0.5;
+    }
+  }
 `;
--- a/frontend/src/components/core/RangeSlider.tsx
+++ b/frontend/src/components/core/RangeSlider.tsx
@@ -15,6 +15,7 @@ export const Slider = styled.input.attrs((props: SliderRange) => ({
   value: props.value,
 }))<SliderRange>`
   -webkit-appearance: none;
+  max-width: 370px;
   width: 100%;
   height: 15px;
   border-radius: 5px;
--- a/frontend/src/components/core/RangeSlider.tsx
+++ b/frontend/src/components/core/RangeSlider.tsx
@@ -1,0 +1,48 @@
+import styled from 'styled-components';
+
+export const SliderContainer = styled.div`
+  width: 100%;
+  margin: 0.5rem 0;
+`;
+
+type SliderRange = {
+  value: number,
+};
+
+export const Slider = styled.input.attrs((props: SliderRange) => ({
+  type: 'range',
+  class: 'slider',
+  value: props.value,
+}))<SliderRange>`
+  -webkit-appearance: none;
+  width: 100%;
+  height: 15px;
+  border-radius: 5px;
+  background: #B7D4FF;
+  outline: none;
+  opacity: 0.7;
+  -webkit-transition: .2s;
+  transition: opacity .2s;
+
+  &:hover {
+    opacity: 1;
+  }
+
+  &::-webkit-slider-thumb {
+    -webkit-appearance: none;
+    appearance: none;
+    width: 25px;
+    height: 25px;
+    border-radius: 10px;
+    background: linear-gradient(180deg, #133ED7 0%, #90BDFF 100%);
+    cursor: pointer;
+  }
+
+  &::-moz-range-thumb {
+    width: 25px;
+    height: 25px;
+    border-radius: 10px;
+    background: linear-gradient(180deg, #133ED7 0%, #90BDFF 100%);
+    cursor: pointer;
+  }
+`;
--- a/frontend/src/components/core/RangeSlider.tsx
+++ b/frontend/src/components/core/RangeSlider.tsx
@@ -47,11 +47,17 @@ export const Slider = styled.input.attrs((props: SliderRange) => ({
   }
 
   &:disabled {
-    cursor: default;
     opacity: 0.5;
 
-    &:hover {
+    &::-webkit-slider-thumb {
       cursor: default;
+    }
+
+    &::-moz-range-thumb {
+      cursor: default;
+    }
+
+    &:hover {
       opacity: 0.5;
     }
   }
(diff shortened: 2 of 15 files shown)


zpChris:
[Lobby page redesign](https://trello.com/c/yTCS2ejv/299-lobby-page-redesign)

frontend/src/views/Lobby.tsx
```diff
@@ -330,88 +396,116 @@ function LobbyPage() {
...
+        </PrimaryButtonZeroLeftMargin>
+      </HeaderContainer>
+
+      <FlexBareContainerLeft>
+        <PlayersContainer>
+          <SmallHeaderTextZeroTopMargin>
+            Players
+            {
+              (activeUsers && inactiveUsers)
+                ? ` (${activeUsers.length + inactiveUsers.length})`
```
alankbi, line 442: nit: I think the roomdto still passes the original list of users, so you might be able to just return users.length instead

frontend/src/views/Lobby.tsx
```diff
@@ -330,88 +396,116 @@ function LobbyPage() {
...
+              {Object.keys(Difficulty).map((key) => {
+                const difficultyKey: Difficulty = Difficulty[key as keyof typeof Difficulty];
+                return (
+                  <SmallDifficultyButton
+                    difficulty={difficultyKey}
+                    onClick={() => updateDifficultySetting(key)}
+                    active={difficulty === difficultyKey}
+                    enabled={isHost(currentUser)}
+                    disabled={!isHost(currentUser)}
+                    title={!isHost(currentUser) ? 'Only the host can change these settings' : undefined}
```
alankbi, line 471: Is there any way to make this + the start game button more obvious that it's host-only? The title usually doesn't show up until you leave your cursor unmoving for a couple seconds (at first I was gonna suggest adding a title but then just now saw you already had it lol)

frontend/src/views/Lobby.tsx
```diff
@@ -1,31 +1,93 @@
...
+`;
+
+const RoomSettingsContainer = styled.div`
+  flex: 4;
+  padding-left: 5px;
+`;
+
+const SmallHeaderTextZeroTopMargin = styled(SmallHeaderText)`
+  margin: 0 0 10px 0;
+`;
```
alankbi, line 76: Slight personal preference but for some of these components, could we rename them to something more descriptive of its use case as opposed to the type of style? i.e. maybe something like SmallSettingsHeaderText for this, and maybe just StartGameButton instead of PrimaryButtonZeroLeftMargin for example (that way in the future if we ever change these styles we won't have to refactor the names of all the components to reflect the CSS)

frontend/src/components/card/HostActionCard.tsx
```diff
@@ -1,25 +1,43 @@
...
 import { User } from '../../api/User';
 
-const Content = styled.div`
+type PlayerIconType = {
+  isActive: boolean,
+};
+
+const Content = styled.div<PlayerIconType>`
   // Center div above parent
   position: absolute;
```
alankbi, line 12: if absolute doesn't work (in terms of getting it to overflow outside of the container), how about position fixed? If not yeah I think it'd be good to address this in some PR later since it can be slightly difficult to see the card when the player is near the bottom of the container (or maybe if we increased the height of the container?)

alankbi (changes requested):
Looks amazing! It's all coming together :)

I have a couple notes on UI and some minor code stuff but I don't think it's anything major: 

1. For the top bar, thoughts on making the text just a tad bit smaller? As well as or alternatively if there's a way we could reduce the number of words - there's two codejoust.co/play URLs which would be nice to somehow reduce to one. One possible suggestion (might not be the best) is if the large top text was like Join at (underline link maybe) codejoust.co/play with Room ID: ... and then below is smaller less emphasized text "or share the link with <the dark gray link>"
2. Maybe we could center the top part? It can look a little off on larger screen sizes aligned left (which actually would be the default for me). or maybe just making the width smaller idk
3. Could we increase the heights of the player and room settings container? When I first saw I thought it might have been cut off cause there was a lot of extra space at the bottom - personally I'm totally fine with the scroll (similar to on the problem edit page) but I can see how it'd be nice to minimize it too. Idk - maybe just having it reach the bottom of the screen or a larger height/width aspect ratio might help
4. maybe not for this PR but the players can jump around a bit with varied nickname lengths and when people refresh/leave the page (e.g. a top-placed player goes inactive and it shifts the whole list of players). not sure what the best solution for this is (maybe at some point adding more order/structure to player placements) but I think it still looks very good for now

Screenshots for reference


Large screen size empty space

zpChris:
> 1. For the top bar, thoughts on making the text just a tad bit smaller? As well as or alternatively if there's a way we could reduce the number of words - there's two codejoust.co/play URLs which would be nice to somehow reduce to one. One possible suggestion (might not be the best) is if the large top text was like Join at (underline link maybe) codejoust.co/play with Room ID: ... and then below is smaller less emphasized text "or share the link with "

For sure, I can make that text smaller - I do want to keep the copy text for the longer link, as that allows people to share the room link very quickly and is pretty convenient imo. Are you thinking that the top portion takes up too much space, or would simply shrinking the text here solve this issue?

> 2. Maybe we could center the top part? It can look a little off on larger screen sizes aligned left (which actually would be the default for me). or maybe just making the width smaller idk

I think I'll try making the width smaller, like a max-width, so it hopefully still works for those screens. Or the text could be even larger for this case - I do like the aligned left text imo. There is an example of the centered text below - let me know if you think this looks better, but I think that it makes the page seem a little less structured since the lines do not all start from the same left position.

Lobby Page Top Centered

> 3. Could we increase the heights of the player and room settings container? When I first saw I thought it might have been cut off cause there was a lot of extra space at the bottom - personally I'm totally fine with the scroll (similar to on the problem edit page) but I can see how it'd be nice to minimize it too. Idk - maybe just having it reach the bottom of the screen or a larger height/width aspect ratio might help

I'll definitely update this so that it takes up the full page (with a min-height of something), and hopefully that means it will be larger since the text above will be smaller now. I do want to restrict it to one page with no scroll if possible, as I think that's generally the design aesthetic for our site (with the exception of the problem edit page, but that page does have more information so I'm ok with that).

> 4. maybe not for this PR but the players can jump around a bit with varied nickname lengths and when people refresh/leave the page (e.g. a top-placed player goes inactive and it shifts the whole list of players). not sure what the best solution for this is (maybe at some point adding more order/structure to player placements) but I think it still looks very good for now

There is some level of order / structure - the inactive players go at the end of the list, and the active ones go at the start. I can see how that creates a problem - we could simply remove that distinction which means everybody would be in the same order, but I'll leave it as-is for now, just let me know if I should change anything specific on this PR.

zpChris:
@alankbi, do you know how to ensure that the final div (Players and Room Settings) simply takes up the remainder of the page? I struggled with implementing this - I tried referencing the Game Page and [this Stack Overflow post](https://stackoverflow.com/questions/90178/make-a-div-fill-the-height-of-the-remaining-screen-space), but to little avail.

frontend/src/components/core/RangeSlider.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const Slider = styled.input.attrs((props: SliderRange) => ({
+  type: 'range',
+  class: 'slider',
+  value: props.value,
+}))<SliderRange>`
+  -webkit-appearance: none;
+  width: 100%;
+  height: 15px;
+  border-radius: 5px;
+  background: #B7D4FF;
```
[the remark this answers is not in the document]
zpChris: Thanks for that note - I also changed the gradients below on the "thumb" to use the blue gradient from theme.

frontend/src/views/Lobby.tsx
zpChris: Yeah that's a good idea - I'll make some tooltip appear immediately on hover, as I agree that the "title" isn't a good solution here.

frontend/src/views/Lobby.tsx
zpChris: I think that makes sense, sounds good 👍 I had originally named it that because it was used in two different cases, but I simplified it to use the name `LobbyContainerTitle`.

frontend/src/components/card/HostActionCard.tsx
zpChris: Based on some quick testing that first solution may work - how do you get the fixed component to center right above the card? When I added that it put it relative to the page, but adding `position: relative` to the outer container didn't help.

frontend/src/views/Lobby.tsx
zpChris: Ok, I spent a very long time on a solution here, because I wanted the tooltip to move around with the mouse and it had to show only on disabled elements. Since disabled elements don't fire events, I then created a div parent wrapper around those elements, transferring the margin from the original elements to the container ones. Following that, I realized that this solution still didn't work, as this still prevented the divs from firing elements on the specific area where the disabled element is. The final solution has the wrapper div with the attribute `position: relative`, then I add a child element (sibling to the original disabled element) with absolute positioning that has the identical height and width of the disabled element, and a z-index of one greater.

This now works, though it's certainly a complicated solution. I shied away from simply removing the "disabled" component as I did want to keep that integrity, though I'd understand if that might be a better solution given this one's complexity. I think this should all work fine now, but just wanted to note down the process here 👍

frontend/src/components/card/HostActionCard.tsx
zpChris: I'm not sure how to do this - based on this [description of position fixed](https://stackoverflow.com/questions/5209814/can-i-position-an-element-fixed-relative-to-parent) and some initial attempts, I don't know how this would work.

I'm open to investigating more (maybe pair programming?), though I think that might be a best for a separate PR since there are additional last few updates that would be helpful (like setting up flex on this page so the bottom containers go to the end of the page, and no farther).

zpChris:
Re-requested review @alankbi - some updates like having the ending containers reach the bottom of the page and updating the host action hover cards are not yet implemented, but I think those would be better done on a new PR. I updated sizing and styling and set some max-widths, and added a decently complex system for providing a mouse tooltip on hover for non-hosts that works on disabled elements (trickier than it seems initially, as described [here](https://github.com/rocketden/main/pull/120#discussion_r600266607)). 👍

frontend/src/views/Lobby.tsx
alankbi: I think it looks great! thanks for doing that

frontend/src/components/card/HostActionCard.tsx
alankbi: Yeah that sounds good - I'll make a list at the bottom for this and the others mentioned so that it's consolidated in one place

alankbi:
> @alankbi, do you know how to ensure that the final div (Players and Room Settings) simply takes up the remainder of the page? I struggled with implementing this - I tried referencing the Game Page and [this Stack Overflow post](https://stackoverflow.com/questions/90178/make-a-div-fill-the-height-of-the-remaining-screen-space), but to little avail.

I do not haha - but yeah that's fine, we can look into this in the next PR as mentioned

alankbi:
1: I was mainly thinking the font size slightly smaller and/or rewording it to make the total text shorter, but this is more of a nit than a major blocking point

2. Thanks! I think it looks a lot better now with the new width setting

alankbi (approved):
Approved! As mentioned, I think we should be okay to move the following potential features either to #121 or a later PR (as well as since Jason is waiting on this to be merged):

* Figuring out the host action card hovering effect (if possible)
* Setting height of player and settings containers (if possible)
* Maybe looking into the ordering of player cards? More of a nit/slight design change tho

frontend/src/views/Lobby.tsx
```diff
@@ -1,31 +1,166 @@
...
+  border-radius: 0.25rem;
+  background: ${({ theme }) => theme.colors.text};
+  color: ${({ theme }) => theme.colors.white};
+  font-size: ${({ theme }) => theme.fontSize.medium};
+  position: absolute;
+  top: 0;
+  left: 0;
+`;
+
+const HoverContainer = styled.div`
```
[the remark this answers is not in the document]
zpChris: Agreed, I think that makes sense - and nice, glad it's being put to use elsewhere 😃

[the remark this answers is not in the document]
zpChris: Done ✅ I moved most of those components to a separate `HoverTooltip.tsx` file in components, but left some of the styled components specific to elements on this page in the `Lobby.tsx` file.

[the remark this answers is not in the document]
zpChris: Also want to note that there are some strange effects with positioning of the tooltip when the user is zoomed in or out (it still works, but is positioned a decent amount away from the mouse for some reason), but I don't think this is a huge styling issue.

zpChris:
When creating the new `HoverTooltip.tsx`, I added a styled-components import but accidentally used double quotes - this causes an eslint warning, which I fixed in #121, so merging that should fix the warning. Sorry that I didn't notice that before merging initially!

PR #121, opened 2021-03-20
Add refresh and leave / delete room to lobby page.

zpChris:
### List of Changes:
- Add a "refresh" button to the room page that fetches the latest room details through the `getRoom` REST request and attempts connecting to the socket again, if not already connected.
- Add a "Leave Room" button that anybody can use, and deletes the room if the last person leaves.
- Add a "Delete Room" endpoint and add the relevant tests.
- Slightly resize the landing page "Create a room" button.

### Notes:
- The "Delete Room" request also automatically deletes all of the users that were in that room, I believe due to cascading effects. This works well for now, as we don't want to keep anonymous users in the database, but will not be the desired behavior once / if we implement registered users (online, student, or teacher).
- I'm not certain how useful the refresh button is - it doesn't seem to provide any updates for the "active / inactive" state of users (there should be more troubleshooting on why exactly this is), but can update some other contents. It may be useful largely as a [Placebo Button](https://en.wikipedia.org/wiki/Placebo_button) (with the distinguishing factor that this button _can_ be helpful, though potentially only in select cases).
- Potentially in the future, the refresh button could also ping all of the users to confirm their "active / inactive" socket connection state. (Related to this, it may be important for reliability to set up "ping pong" socket confirmations.)
- When the user clicks "Leave Room" and accepts the window confirmation, they are immediately disconnected without waiting for the actual "kickUser" and "disconnect" callbacks to return; I did this because in my mind, regardless of whether those actions are successful, the user should leave the room. However, in the case that these functions do fail, this may create unexpected behaviors.

### Screencast and Images:

[Screencast of Lobby Page Refresh and Leave Room Buttons](https://www.loom.com/share/91075b8d52fa40c785f4387a0e352313)

_Note:_ I meant to demonstrate the "refresh" functionality on the above screencast too, but forgot - you can test this by directly deleting a user in the database (so no updates will be immediately rendered on the frontend), then clicking refresh to show that information is received and updated. I have confirmed it also attempts to connect to the socket if not already done so, though this feature in particular may be more flaky.

**Add Refresh and Leave Room Buttons to the Lobby Page**
Add Refresh and Leave Room Buttons to the Lobby Page

**Updated Lobby Page with Disabled Buttons**
Updated Lobby Page with Disabled Buttons

**Updated Button Size on Landing Page**
Updated Button on Landing Page

### DIFF
--- a/frontend/src/components/core/Button.tsx
+++ b/frontend/src/components/core/Button.tsx
@@ -158,3 +158,21 @@ export const InvertedSmallButton = styled(SmallButton)`
   color: ${({ theme }) => theme.colors.text};
   background: ${({ theme }) => theme.colors.background};
 `;
+
+export const InlineRefreshIcon = styled.i.attrs(() => ({
+  className: 'material-icons',
+}))`
+  display: inline-block;
+  margin: 0 0 0 10px;
+  padding: 0.25rem;
+  border-radius: 1rem;
+  font-size: ${({ theme }) => theme.fontSize.default};
+  background: ${({ theme }) => theme.colors.white};
+  box-shadow: 0 1px 8px rgba(0, 0, 0, 0.24);
+  color: ${({ theme }) => theme.colors.font};
+
+  &:hover {
+    cursor: pointer;
+    box-shadow: 0 2px 12px rgba(0, 0, 0, 0.24);
+  }
+`;
--- a/frontend/src/components/core/Button.tsx
+++ b/frontend/src/components/core/Button.tsx
@@ -43,7 +43,7 @@ export const PrimaryButton = styled(DefaultButton)<any>`
 
 export const SecondaryRedButton = styled(PrimaryButton)<any>`
   background: ${({ theme }) => theme.colors.white};
-  color: ${({ theme }) => theme.colors.red};
+  color: ${({ theme }) => theme.colors.red2};
   margin: 1.2rem 0;
 `;
 
--- a/frontend/src/components/core/Button.tsx
+++ b/frontend/src/components/core/Button.tsx
@@ -41,6 +41,12 @@ export const PrimaryButton = styled(DefaultButton)<any>`
   }
 `;
 
+export const SecondaryRedButton = styled(PrimaryButton)<any>`
+  background: ${({ theme }) => theme.colors.white};
+  color: ${({ theme }) => theme.colors.red};
+  margin: 1.2rem 0;
+`;
+
 export const TextButton = styled.button<ThemeType>`
   background: none;
   border: none;
--- a/frontend/src/views/Lobby.tsx
+++ b/frontend/src/views/Lobby.tsx
@@ -341,11 +341,13 @@ function LobbyPage() {
       conditionallyBootKickedUser(room, currentUser);
     };
 
+    setLoading(true);
     connect(roomId, userId).then(() => {
       // Body encrypt through JSON.
       subscribe(routes(roomId).subscribe_lobby, subscribeCallback).then((subscriptionParam) => {
         setSubscription(subscriptionParam);
         setSocketConnected(true);
+        setLoading(false);
       }).catch((err) => {
         setError(err.message);
       });
@@ -362,6 +364,11 @@ function LobbyPage() {
         .then((res) => {
           setStateFromRoom(res);
           updateCurrentUserDetails(res.users);
+
+          // Attempt to connect the user to the socket.
+          if (currentUser && currentUser.userId) {
+            connectUserToRoom(res.roomId, currentUser.userId);
+          }
           setLoading(false);
         })
         .catch((err) => setError(err));
--- a/frontend/src/views/Lobby.tsx
+++ b/frontend/src/views/Lobby.tsx
@@ -393,12 +393,12 @@ function LobbyPage() {
         });
         disconnect();
       }
-    }
 
-    // Redirect user regardless of POST request success.
-    history.replace('/game/join', {
-      error: errorHandler('You left the room.'),
-    });
+      // Redirect user regardless of POST request success.
+      history.replace('/game/join', {
+        error: errorHandler('You left the room.'),
+      });
+    }
   };
 
   // Grab the nickname variable and add the user to the lobby.
--- a/frontend/src/views/Lobby.tsx
+++ b/frontend/src/views/Lobby.tsx
@@ -389,19 +389,14 @@ function LobbyPage() {
         removeUser(currentRoomId, {
           initiator: currentUser,
           userToDelete: currentUser,
-        }).catch(() => {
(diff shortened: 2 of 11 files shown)


zpChris:
[Lobby page refresh, leave room, and delete room](https://trello.com/c/D6DAE97F/303-lobby-page-refresh-leave-room-and-delete-room)

frontend/src/components/core/Button.tsx
```diff
@@ -41,6 +41,12 @@ export const PrimaryButton = styled(DefaultButton)<any>`
   }
 `;
 
+export const SecondaryRedButton = styled(PrimaryButton)<any>`
```
alankbi, line 44: Is this any tag being used? If not I think we can get rid of it

```suggestion
export const SecondaryRedButton = styled(PrimaryButton)`
```

frontend/src/views/Lobby.tsx
```diff
@@ -337,19 +343,64 @@ function LobbyPage() {
...
+      getRoom(location.state.roomId)
+        .then((res) => {
+          setStateFromRoom(res);
+
+          // Boot the user from the room, if they are not present.
+          updateCurrentUserDetails(res.users);
+
+          // Attempt to connect the user to the socket.
+          if (currentUser && currentUser.userId) {
+            connectUserToRoom(res.roomId, currentUser.userId);
```
alankbi, line 377: Are there any checks for when the user is already connected not to connect them again?

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -137,7 +154,17 @@ public RoomDto removeUser(String roomId, RemoveUserRequest request) {
...
+        if (room.getHost().equals(userToDelete)) {
+
+            // If the host is the last user, delete the room and return info.
+            if (room.getUsers().size() == 1) {
+                DeleteRoomRequest deleteRoomRequest = new DeleteRoomRequest();
+                deleteRoomRequest.setHost(request.getUserToDelete());
+                return deleteRoom(roomId, deleteRoomRequest);
+            }
+
+            return conditionallyUpdateRoomHost(room, userToDelete, true);
```
alankbi, line 166: Suggestion: instead of returning here, could we just remove the user and then call the conditional room host update afterwards? That way we could avoid passing an extra boolean flag to handle an edge case

frontend/src/views/Lobby.tsx
```diff
@@ -337,19 +343,64 @@ function LobbyPage() {
...
+          userToDelete: currentUser,
+        });
+        disconnect();
+      }
+    }
+
+    // Redirect user regardless of POST request success.
+    history.replace('/game/join', {
+      error: errorHandler('You left the room.'),
+    });
```
alankbi, line 401: I think this should be moved into the if statement - there's a slight bug rn where even if you click cancel you'll still be redirect away from the room (although not kicked on the backend)

frontend/src/views/Lobby.tsx
```diff
@@ -337,19 +343,64 @@ function LobbyPage() {
...
+    // eslint-disable-next-line no-alert
+    if (window.confirm('Are you sure you want to leave the room?')) {
+      if (currentUser && currentUser.userId) {
+        setLoading(true);
+        setError('');
+        removeUser(currentRoomId, {
+          initiator: currentUser,
+          userToDelete: currentUser,
+        });
+        disconnect();
```
alankbi, line 394: I might prefer to have the redirect and disconnect moved into the then block of removeUser just for safety's sake (with async stuff if it redirects you from the page halfway through the removeUser axios call, idk if that would be prone to occasional misses) - but if you've tested this a lot and it works every time, I think that should be okay too

alankbi (changes requested):
Awesome, looks great! I requested changes because of one small bug, but other than that, no major gripes 👍

frontend/src/components/core/Button.tsx
zpChris: I don't believe so, yeah we can get rid of that.

frontend/src/views/Lobby.tsx
zpChris: I believe that check is done in the actual [`connect` api function](https://github.com/rocketden/main/blob/1bffb0d706f386b57ac51f001e8e62853386276c/frontend/src/api/Socket.ts#L36).

src/main/java/com/rocketden/main/service/RoomService.java
zpChris: So, that's what was originally done, but there was a bug where two separate actions would be made on the database - the host would be changed to a regular user and a new host is instated, and the old host is deleted - these two database updates seemed to conflict as the host was not actually being deleted from the room. Therefore, I changed the function to accept a parameter so `repository.save(room)` is only called once, and no database conflicts occur (this seemed to fix the problem, so I assumed this was the issue).

frontend/src/views/Lobby.tsx
zpChris: Ah good catch, thanks, I'll update this.

zpChris: Done ✅

frontend/src/views/Lobby.tsx
zpChris: I do think that we can leave this in - this doesn't have them leave the whole website, so this code still runs, and I haven't encountered any issues thus far. I'm happy to change this back if we do encounter that, but thus far even if it redirects and disconnects the user before completing the removeUser call, that call still comes back. Since I think it should redirect and disconnect the user either way, I think it makes sense to include it regardless of removeUser's success as well.

src/main/java/com/rocketden/main/service/RoomService.java
alankbi: Ah I see thanks - looks like another concurrency issue similar to what was discovered in #122. I do think it'd be good to not have to rely on passing booleans across methods, but I guess since there's potentially no easy fix until the concurrency stuff is implemented, this looks good for this PR 👍

zpChris:
Just saw that there are merge conflicts, so I'm resolving this now.

alankbi (approved):
Looks good, thanks for fixing the redirect bug!

If you're planning on implementing any of the changes from #120 in this PR, feel free to re-request, otherwise looks good to merge but could you add those items to a card in the backlog?

zpChris:
> Looks good, thanks for fixing the redirect bug!
> 
> If you're planning on implementing any of the changes from #120 in this PR, feel free to re-request, otherwise looks good to merge but could you add those items to a card in the backlog?

For sure - not planning on adding any of those changes to this PR, so I put them on [this Trello card](https://trello.com/c/lImY6Rzb/308-lobby-page-future-design-changes).

PR #122, opened 2021-03-23
adding backend and tests for a room limit for users

jasonz6688:
### List of Changes:

- A fifth user attempting to join a room will be unable to and given the message "Cannot join a room that is already full."

### Notes:

- isFull() depends on the users list being correct, with all the users
- Currently the room limit is hardcoded at four
- The exception for an already full room is HttpStatus.FORBIDDEN. I'm not sure that's the correct one.
- The JUnit tests are dependent on the room limit being set to 4.

### Screencast and Images:

Cap62
Cap61

### DIFF
--- a/src/main/java/com/rocketden/main/service/RoomService.java
+++ b/src/main/java/com/rocketden/main/service/RoomService.java
@@ -47,6 +47,10 @@ public class RoomService {
             throw new ApiException(RoomError.NOT_FOUND);
         }
 
+        if (room.isFull()) {
+            throw new ApiException(RoomError.ALREADY_FULL);
+        }
+
         // Get the user who initialized the request.
         User user = UserMapper.toEntity(request.getUser());
 
--- a/src/main/java/com/rocketden/main/exception/RoomError.java
+++ b/src/main/java/com/rocketden/main/exception/RoomError.java
@@ -14,7 +14,8 @@ public enum RoomError implements ApiError {
     NO_HOST(HttpStatus.BAD_REQUEST, "There is no host provided."),
     NOT_FOUND(HttpStatus.NOT_FOUND, "A room could not be found with the given id."),
     DUPLICATE_USERNAME(HttpStatus.CONFLICT, "A user with the nickname provided has already joined the room."),
-    ALREADY_ACTIVE(HttpStatus.FORBIDDEN, "Cannot join a room that is already active.");
+    ALREADY_ACTIVE(HttpStatus.FORBIDDEN, "Cannot join a room that is already active."),
+    ALREADY_FULL(HttpStatus.FORBIDDEN, "Cannot join a room that is already full.");
 
     private final HttpStatus status;
     private final ApiErrorResponse response;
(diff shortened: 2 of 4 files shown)


jasonz6688:
[Add limit on the number of people that can participate in a single room](https://trello.com/c/ihEUY1IF/292-add-limit-on-the-number-of-people-that-can-participate-in-a-single-room)

zpChris:
Awesome job - remember to add me (@zpChris) and / or Alan (@alankbi) as reviewers and add labels (likely `backend`) to this PR. 👍

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -47,6 +47,10 @@ public RoomDto joinRoom(String roomId, JoinRoomRequest request) {
             throw new ApiException(RoomError.NOT_FOUND);
         }
 
+        if (room.isFull()) {
+            throw new ApiException(RoomError.ALREADY_FULL);
+        }
```
zpChris, line 52: I wonder if two new users attempt to join simultaneously, if there is any chance that the `room.isFull()` function won't update for them correctly, meaning that the room has too many users. @alankbi, I'm not sure how our backend handles requests like these, will they be done in sequence such that this won't happen or is this bug possible?

If it's possible (that's a big if, I'm not sure), my suggestion would be to fix by checking on the frontend that if a room has more people than the current limit, then the last user to join (presumably the last person in the list, though I don't know if that list preserves order of entry) gets kicked out of the room. This could cause a slight delay where the user joins and then is kicked, but that's a better solution than an accidental violation of the room limit.

As for actions on this PR though, I don't think there are any as this would deal with more complexity and potentially a frontend issue - just noting this down as a potential issue for the future. 👍

zpChris (approved):
LGTM! Great first PR 😃 🎉 💯

src/main/java/com/rocketden/main/service/RoomService.java
zpChris: For reference, I just tested and my prediction was correct - if two users join at the same time, it still adds both of them. (See below image, sorry about the misspelled name lol.)

**Five Users in Four-Limit Room**
Five Users in Four-Limit Room

alankbi (approved):
No gripes on the first PR - awesome work :)

src/main/java/com/rocketden/main/service/RoomService.java
alankbi: Hm yeah that's definitely something that should be fixed in the future - right now it's not too bad cause we don't have any super critical stuff but as we add more things that could cause some bugs. 

In terms of a fix, I think that the best solution might be to implement some kind of [locking system](https://www.baeldung.com/java-jpa-transaction-locks) which addresses the root of the problem (and hopefully would make sure there are zero edge cases/bugs that come up)

zpChris: That's a good idea, I like it 👍 Yeah definitely a future implementation though, makes sense

PR #126, opened 2021-03-28
Adding front end for the room limit and connecting it to the back end

jasonz6688:
### List of Changes:

- Added a size slider in the lobby
- Added a size quantity to the Room.ts and the RoomDto
- The size slider sets the size in the Room.java, which constrains how many users can join the room. This slider cannot be set below the number of players already in the game.

### Notes:

- Anytime, the host attempts to set the room size to less than the number of players, the slider is set to the number players.
- I changed the original joinFullRoom test so that the room's size is set to 4 explicitly. I also changed the updateRoomSettingsSuccess test to include a room size.
- The code checks on both the front end and back end that the updateRoomSettings is not called on a value that is not allowed (less than 0, less than the number of users connected, or more than MAX_SIZE + 1). MAX_SIZE + 1 results in no room limit.
- The error message that displayed when the host attempts to set the room size to less than the number of players doesn't disappear until the slider is changed again. To do this, the updateSize method on the frontend doesn't clear the error message, which differs from the updateDuration method. Otherwise, releasing the slider would've automatically cleared the error message, making it fairly difficult to read.
- I didn't create a new exception on the backend. I just used BAD_SETTING.

 

### Screencast and Images:

Standard view of the slider:
Cap61

View when there's no size limit:
Cap62

Size limit in use:
Cap63
Cap64

Error message when the slider is set too low:
Cap65

EDIT: I fixed the capitalization to be "x people", and added a special case for "1 person".


frontend/src/views/Lobby.tsx
```diff
@@ -142,6 +142,7 @@ function LobbyPage() {
   const [active, setActive] = useState(false);
   const [difficulty, setDifficulty] = useState<Difficulty | null>(null);
   const [duration, setDuration] = useState<number | undefined>(15);
+  const [size, setSize] = useState<number | undefined>(4);
```
zpChris, line 145: nit: Thoughts on making the default larger, maybe even like `10`? This would require a change to the backend as well, though I'm open on this.

frontend/src/views/Lobby.tsx
```diff
@@ -639,6 +660,54 @@ function LobbyPage() {
...
+                      if (newSize >= (inactiveUsers?.length || 0) + (activeUsers?.length || 0) && newSize <= 31) {
+                        if (newSize !== size) {
+                          setError('');
+                        }
+
+                        setSize(newSize);
+                      } else if (newSize < (inactiveUsers?.length || 0) + (activeUsers?.length || 0)) {
+                        setSize((inactiveUsers?.length || 0) + (activeUsers?.length || 0));
+                        setError("The room limit cannot be set below the number of connected players. Kick a player to set the limit lower.");
+                      }
```
zpChris, line 704: Replace the addition of `inactiveUsers` and `activeUsers` with just `users`.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -210,6 +212,38 @@ public void joinFullRoomFailure() {
     }
 
 
+    @Test
+    public void manyUsersJoiningAnInfinitelySizedRoomSuccess() {
+        /**
+         * Verify join room request fails when the room is already full
+         * Define five users, and add to the room
+         */
```
zpChris, line 220: Change this comment to reflect what this test case is actually doing - I like the addition of the comments by the way, that helps a lot, thank you!

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -210,6 +212,38 @@ public void joinFullRoomFailure() {
...
+        for (int i = 0; i < 100; i++) {
+            User temp = new User();
+            temp.setNickname("Rocket" + i);
+            room.addUser(temp);
+        }
+
+        // Mock repository to return room when called
+        Mockito.doReturn(room).when(repository).findRoomByRoomId(eq(ROOM_ID));
+        assertDoesNotThrow(() -> roomService.joinRoom(ROOM_ID, request));
+        verify(repository).findRoomByRoomId(ROOM_ID);
```
zpChris, line 243: Perhaps also verify that the number of users is now equal to how many there are in this room - `101` or so. You could replace the `assertDoesNotThrow` with this, or simply keep both in (either works imo).

frontend/src/views/Lobby.tsx
```diff
@@ -639,6 +660,54 @@ function LobbyPage() {
...
+                  }
+                }}
+              />
+              <SliderContainer>
+                <Slider
+                  min={1}
+                  max={31}
+                  value={size}
+                  disabled={!isHost(currentUser)}
+                  onChange={(e) => {
```
zpChris, line 687: nit: This function is pretty large, and might be better suited to go outside as a separate function which this `onChange` action calls.

frontend/src/views/Lobby.tsx
```diff
@@ -331,6 +333,25 @@ function LobbyPage() {
       });
   };
 
+  const updateSize = () => {
```
zpChris, line 336: This comment isn't specifically related to this line, but install the [Visual Studio Code ESLint Extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and address the errors / warnings that come up! These include a max line length of 100, using single-quotes for strings, and using expected property shorthand. Let me know if you have any troubles with this - this is the same extension that I have, but if you run into issues or don't see these warnings just message me.

frontend/src/views/Lobby.tsx
```diff
@@ -639,6 +660,54 @@ function LobbyPage() {
                 />
               </SliderContainer>
             </HoverContainerSlider>
+            <NoMarginMediumText>Room Size</NoMarginMediumText>
+            <NoMarginSubtitleText>
+              {size === 31? 'No limit' : (size === 1? '1 person' : `${size} people`)}
```
zpChris, line 665: really minor nit:
```suggestion
              {size === 31 ? 'No limit' : (size === 1 ? '1 person' : `${size} people`)}
```

frontend/src/views/Lobby.tsx
```diff
@@ -331,6 +333,25 @@ function LobbyPage() {
...
+    setLoading(true);
+    const prevSize = size;
+    const settings = {
+      initiator: currentUser!,
+      size: size,
+    };
+
+
+    updateRoomSettings(currentRoomId, settings)
+      .then(() => setLoading(false))
```
zpChris, line 346: Also add `setError('')` here, as sometimes, the room size can show an error even when not necessary ([screencast showing this](https://www.loom.com/share/8ef1a6d5d499480ca59aa7ed3e4145bf)).

frontend/src/views/Lobby.tsx
```diff
@@ -331,6 +333,25 @@ function LobbyPage() {
...
+      .catch((err) => {
+        setLoading(false);
+        setError(err.message);
+        // Set duration back to original if REST call failed
```
zpChris, line 350:
```suggestion
        // Set size back to original if REST call failed
```

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -130,6 +132,10 @@ public boolean containsUserWithNickname(String nickname) {
      * than the maximum size of the room
      */
     public boolean isFull() {
-        return users.size() >= maxSize;
+        if (size == RoomService.MAX_SIZE + 1) {
+            return false;
+        }
```
zpChris, line 137: nit: Add comment explaining why this is in place, either in the method header or right above the if statement.

zpChris (changes requested):
This looks great, awesome job! Requesting changes as I have some comments on code style, ESLint warnings, and some slight error handling changes. 👍

alankbi:
[Add max users per room to frontend lobby page](https://trello.com/c/ryMdc0U6/315-add-max-users-per-room-to-frontend-lobby-page)

frontend/src/views/Lobby.tsx
alankbi: +1 on this, I remember when we played a game with the TeamsCode team we unexpectedly ran into this room limit haha (and from that demo it seems the service works fine with 7-8 people as long as we don't hit the max questions error, so I think my previous concern about that no longer exists)

frontend/src/views/Lobby.tsx
alankbi: Additionally, this gives me an `ESLint: Do not nest ternary expressions.(no-nested-ternary)` warning, so we might need a different way to do this to avoid the nested ternary (maybe putting the second ternary inside the template string?)

frontend/src/views/Lobby.tsx
```diff
@@ -639,6 +660,54 @@ function LobbyPage() {
...
+              <SliderContainer>
+                <Slider
+                  min={1}
+                  max={31}
+                  value={size}
+                  disabled={!isHost(currentUser)}
+                  onChange={(e) => {
+                    const { value } = e.target;
+
+                    // Set duration to undefined to allow users to clear field
```
alankbi, line 690: Suggested change - however, when is it necessary to setSize to be undefined? 

```suggestion
                    // Set size to undefined to allow users to clear field
```

frontend/src/views/Lobby.tsx
```diff
@@ -639,6 +660,54 @@ function LobbyPage() {
...
+                      const newSize = Number(value);
+                      if (newSize >= (inactiveUsers?.length || 0) + (activeUsers?.length || 0) && newSize <= 31) {
+                        if (newSize !== size) {
+                          setError('');
+                        }
+
+                        setSize(newSize);
+                      } else if (newSize < (inactiveUsers?.length || 0) + (activeUsers?.length || 0)) {
+                        setSize((inactiveUsers?.length || 0) + (activeUsers?.length || 0));
+                        setError("The room limit cannot be set below the number of connected players. Kick a player to set the limit lower.");
```
alankbi, line 703: To be honest, I might actually prefer if we didn't have an error message at all - I feel like it's pretty intuitive that the max number of players in a room is set at minimum to the number of people currently in the room. So like for example, if a group of 4 is playing and they decide they don't want to let anyone else join, they can simply drag the slider all the way to the left and it will stay at 4 as expected (without any long error message showing). 

However, if you both feel otherwise, I think that's fine to keep it as well, maybe just make it disappear immediately on change as mentioned in one of Chris's other comments

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -272,6 +273,16 @@ public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request)
             }
             room.setDuration(request.getDuration());
         }
+
+        // Set new size if not null
+        Integer size = request.getSize();
+        if (size != null) {
+            if (size <= 0 || size > MAX_SIZE + 1 || size < room.getUsers().size()) {
+                throw new ApiException(RoomError.BAD_SETTING);
```
alankbi, line 281: nit: maybe add a new custom error specifically for the size field, like "an invalid setting for room size was provided"? I know we currently reuse BAD_SETTING a lot, but I think it might actually be good for us to shift towards more descriptive/specific error messages, especially with all the recent confusing bugs we've come across lol

alankbi: Also, could you fix the typo for `ProblemErrors.INCORRECT_INPUT_COUNT`? One of the TeamsCode members found it while adding a problem lol

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -27,6 +27,7 @@
 
     public static final int ROOM_ID_LENGTH = 6;
     public static final long MAX_DURATION = 3600; // 1 hour
+    public static final long MAX_SIZE = 30;
```
alankbi, line 30: nit: change to to be an int (I think the reason duration is a long is because that's what the timer API expects, but size can just be an int)

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -345,13 +379,15 @@ public void updateRoomSettingsSuccess() {
...
+        request.setSize(5);
         request.setNumProblems(3);
 
         RoomDto response = roomService.updateRoomSettings(room.getRoomId(), request);
 
         verify(socketService).sendSocketUpdate(eq(response));
         assertEquals(request.getDifficulty(), response.getDifficulty());
         assertEquals(request.getDuration(), response.getDuration());
+        assertEquals(request.getSize(), response.getSize());
         assertEquals(request.getNumProblems(), response.getNumProblems());
```
alankbi, line 391: Let me know if you've already done this (such as in the previous PR), but could you add 2 API tests (in the RoomTests.java class), one a failure to join a room when max limit is reached and another for when trying to set an invalid room size (such as less than the number of current players for example)? As well, if there are existing tests such as updateRoomSettingsSuccess, then maybe just add two quick lines to those for setting the max size and checking with an assert in the returned object that the field is set properly

alankbi (commented):
Added a couple comments on top of the ones that @zpChris left, but overall this looks great! Moreso impressed that you did all this in two commits haha (in the future, feel free to be more liberal on this as it might help if something goes wrong and you need to go back a few commits)

frontend/src/views/Lobby.tsx
alankbi: I think in the call we discussed a couple solutions - if we decide to keep the message, I might lean towards having it disappear after the loading is finished as well. If we do that though, then maybe the message itself could be shortened? Cause otherwise it would disappear before people could finish reading it

src/test/java/com/rocketden/main/service/RoomServiceTests.java
jasonz6688: I added the tests to the RoomServiceTests.java file. Is that fine instead?

frontend/src/views/Lobby.tsx
jasonz6688: I installed the extension, and no warnings or issues came up. or at least I didn't see any. Where should I look for the warnings?

zpChris: Hmm, they should simply be red underlines on relevant parts of the code - maybe you have to close and reopen VSCode? Otherwise I'm not sure, we can investigate on the daily call tomorrow.

zpChris:
> Added a couple comments on top of the ones that @zpChris left, but overall this looks great! Moreso impressed that you did all this in two commits haha (in the future, feel free to be more liberal on this as it might help if something goes wrong and you need to go back a few commits)

Yeah I'd second @alankbi on the commits comment, from what I've heard it's actually good practice to commit often, so I'd definitely recommend doing that more - if you run into an issue and not sure what the cause is, commits can help you track when it first appeared.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
zpChris: We have tests in two places, both the unit tests and the integration tests. There is some redundancy here and we've talked about this before - if you think we should probably change habits and just use the service tests we can talk on the call. Thus far we've been doing both, but happy to chat more during the daily standup.

frontend/src/views/Lobby.tsx
jasonz6688: Yeah, I changed it so that it disappeared after loading and made the error shorter.

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -210,6 +253,40 @@ public void joinFullRoomFailure() {
     }
 
 
+    @Test
+    public void manyUsersJoiningAnInfinitelySizedRoomSuccess() {
+        /**
+         * Verify join room request works when the room is infinitely sized
+         * Define a hunderd users, add to the room, then request to add another user
```
zpChris, line 260:
```suggestion
         * Define a hundred users, add to the room, then request to add another user
```

frontend/src/views/Lobby.tsx
```diff
@@ -622,20 +665,56 @@ function LobbyPage() {
...
                   value={duration}
                   disabled={!isHost(currentUser)}
+                  onChange={onSizeSliderChange}
+                  onMouseUp={updateRoomDuration}
+                />
+              </SliderContainer>
+            </HoverContainerSlider>
+            <NoMarginMediumText>Room Size</NoMarginMediumText>
+            <NoMarginSubtitleText>
+              {size === 31 ? 'No limit' : `${size === 1 ? '1 person' : `${size} people`}`}
```
alankbi, line 675: is this a template string inside a ternary inside a template string inside a ternary haha

src/test/java/com/rocketden/main/api/RoomTests.java
```diff
@@ -322,8 +322,54 @@ public void createAndJoinRoomNoUser() throws Exception {
...
+        result = this.mockMvc.perform(put(String.format(PUT_ROOM_JOIN, roomId))
+                .contentType(MediaType.APPLICATION_JSON_VALUE)
+                .content(UtilityTestMethods.convertObjectToJsonString(joinRequest)))
+                .andDo(print()).andExpect(status().is(ERROR.getStatus().value()))
+                .andReturn();
+
+        jsonResponse = result.getResponse().getContentAsString();
+        ApiErrorResponse actual = UtilityTestMethods.toObject(jsonResponse, ApiErrorResponse.class);
+
+        assertEquals(ERROR.getResponse(), actual);
```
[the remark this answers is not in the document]
jasonz6688: That code was actually deleted and copied back, after I accidentally wrote my test in the wrong method. And yeah, I'll fix the comment.

src/test/java/com/rocketden/main/api/RoomTests.java
```diff
@@ -242,7 +242,8 @@ public void createAndJoinRoom() throws Exception {
 
     @Test
     public void createAndJoinRoomUserAlreadyExists() throws Exception {
-        // POST request to create room and PUT request to join room should succeed
+        // POST request to create room and PUT request to join room should fail
+        // because the room already has a user with the exact same information
```
zpChris, line 246: Multi-line comments usually have different styling than the single line `//` comments.
```suggestion
        /**
         * POST request to create room and PUT request to join room should fail
         * because the room already has a user with the exact same information
         */
```

zpChris: Do you mind also changing the comments for `createAndJoinRoomNoUser`? That's the last nit, just updating the ones in that method so they actually reflect what's going on (I know this wasn't your method in the first place, so thanks for fixing that lol).

alankbi (approved):
Awesome, looks great 👍

zpChris (approved):
LGTM! Left a nit and maybe updating some other comments, but that's a really minor documentation change, so this looks great! ✅

PR #131, opened 2021-04-03
problem approval functionality added in backend

sean-xcamp:
### List of Changes:

- Added problem approval boolean to model/Problem.java
- Added toggleApproval() setter method in model/Problem.java
- Changed getProblemByDifficulty method to check approval in addition to difficulty (only approved problems can be picked)
- Changed ProblemService, GameSocket, Game, and Problem tests by setting every problem to approved when it's created
- Added an exception that prevents people from approving a problem through editProblem if it has zero test cases (Note: not yet tested if this actually works. It's possible that an empty test case will still make the index of the test case arrayList 1 and bypass this error. Can test in production)

### Notes:

- Tests passed (phew)

### DIFF
--- a/src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
+++ b/src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
@@ -13,7 +13,6 @@ import lombok.Setter;
 public class CreateProblemRequest {
     private String name;
     private String description;
-    private Boolean approval;
     private ProblemDifficulty difficulty;
     private List<ProblemInputDto> problemInputs;
     private ProblemIOType outputType;
--- a/src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
+++ b/src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
@@ -13,6 +13,7 @@ import lombok.Setter;
 public class CreateProblemRequest {
     private String name;
     private String description;
+    private Boolean approval;
     private ProblemDifficulty difficulty;
     private List<ProblemInputDto> problemInputs;
     private ProblemIOType outputType;
--- a/src/main/java/com/rocketden/main/model/problem/Problem.java
+++ b/src/main/java/com/rocketden/main/model/problem/Problem.java
@@ -33,6 +33,7 @@ public class Problem {
     private String problemId = UUID.randomUUID().toString();
     private String name;
     private Boolean approval = false;
+
     @Column(columnDefinition = "TEXT")
     private String description;
 
--- a/src/main/java/com/rocketden/main/model/problem/Problem.java
+++ b/src/main/java/com/rocketden/main/model/problem/Problem.java
@@ -70,5 +70,4 @@ public class Problem {
         problemInputs.add(problemInput);
         problemInput.setProblem(this);
     }
-
 }
--- a/src/main/java/com/rocketden/main/model/problem/Problem.java
+++ b/src/main/java/com/rocketden/main/model/problem/Problem.java
@@ -71,7 +71,4 @@ public class Problem {
         problemInput.setProblem(this);
     }
 
-    public void toggleApprovedStatus() {
-        approval ^= true;
-    }
 }
--- a/src/main/java/com/rocketden/main/model/problem/Problem.java
+++ b/src/main/java/com/rocketden/main/model/problem/Problem.java
@@ -31,8 +31,8 @@ public class Problem {
 
     // Auto-generate default business ID for each problem
     private String problemId = UUID.randomUUID().toString();
-
     private String name;
+    private Boolean approval = false;
     @Column(columnDefinition = "TEXT")
     private String description;
 
@@ -69,5 +69,9 @@ public class Problem {
     public void addProblemInput(ProblemInput problemInput) {
         problemInputs.add(problemInput);
         problemInput.setProblem(this);
+    }
+
+    public void toggleApprovedStatus() {
+        approval ^= true;
     }
 }
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -62,7 +62,6 @@ public class ProblemService {
         problem.setDescription(request.getDescription());
         problem.setDifficulty(request.getDifficulty());
         problem.setOutputType(request.getOutputType());
-        problem.setApproval(request.getApproval());
 
         // Add all problem inputs in list.
         for (ProblemInputDto problemInput : request.getProblemInputs()) {
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -179,10 +179,8 @@ public class ProblemService {
 
         List<Problem> problems;
         if (difficulty == ProblemDifficulty.RANDOM) {
-            //Change to findAllByApproval once it works
             problems = repository.findAllByApproval(true);
         } else {
-            //Change to findAllByDifficultyAndApproval once it works
             problems = repository.findAllByDifficultyAndApproval(difficulty, true);
         }
 
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -62,6 +62,7 @@ public class ProblemService {
         problem.setDescription(request.getDescription());
         problem.setDifficulty(request.getDifficulty());
         problem.setOutputType(request.getOutputType());
+        problem.setApproval(request.getApproval());
 
         // Add all problem inputs in list.
         for (ProblemInputDto problemInput : request.getProblemInputs()) {
@@ -104,7 +105,9 @@ public class ProblemService {
                 || updatedProblem.getOutputType() == null) {
             throw new ApiException(ProblemError.EMPTY_FIELD);
         }
-
+        if (updatedProblem.getApproval() && updatedProblem.getTestCases().size() == 0) {
+            throw new ApiException(ProblemError.BAD_APPROVAL);
+        }
         if (updatedProblem.getDifficulty() == ProblemDifficulty.RANDOM) {
             throw new ApiException(ProblemError.BAD_DIFFICULTY);
         }
@@ -113,6 +116,7 @@ public class ProblemService {
         problem.setDescription(updatedProblem.getDescription());
         problem.setDifficulty(updatedProblem.getDifficulty());
         problem.setOutputType(updatedProblem.getOutputType());
+        problem.setApproval(updatedProblem.getApproval());
 
         problem.getProblemInputs().clear();
         for (ProblemInputDto problemInput : updatedProblem.getProblemInputs()) {
@@ -169,19 +173,21 @@ public class ProblemService {
             throw new ApiException(ProblemError.EMPTY_FIELD);
         }
 
+        if (numProblems <= 0) {
+            throw new ApiException(ProblemError.INVALID_NUMBER_REQUEST);
+        }
+
         List<Problem> problems;
         if (difficulty == ProblemDifficulty.RANDOM) {
-            problems = repository.findAll();
+            //Change to findAllByApproval once it works
+            problems = repository.findAllByApproval(true);
         } else {
-            problems = repository.findAllByDifficulty(difficulty);
+            //Change to findAllByDifficultyAndApproval once it works
+            problems = repository.findAllByDifficultyAndApproval(difficulty, true);
         }
 
         if (problems == null || problems.isEmpty()) {
             throw new ApiException(ProblemError.NOT_FOUND);
-        }
-
-        if (numProblems <= 0) {
-            throw new ApiException(ProblemError.INVALID_NUMBER_REQUEST);
         }
 
         // If the user wants more problems than exists, just return all of them
--- a/src/main/java/com/rocketden/main/service/ProblemService.java
+++ b/src/main/java/com/rocketden/main/service/ProblemService.java
@@ -102,12 +102,15 @@ public class ProblemService {
                 || updatedProblem.getDifficulty() == null
                 || updatedProblem.getProblemInputs() == null
                 || updatedProblem.getTestCases() == null
-                || updatedProblem.getOutputType() == null) {
+                || updatedProblem.getOutputType() == null
+                || updatedProblem.getApproval() == null) {
             throw new ApiException(ProblemError.EMPTY_FIELD);
         }
+
         if (updatedProblem.getApproval() && updatedProblem.getTestCases().size() == 0) {
             throw new ApiException(ProblemError.BAD_APPROVAL);
         }
+
         if (updatedProblem.getDifficulty() == ProblemDifficulty.RANDOM) {
             throw new ApiException(ProblemError.BAD_DIFFICULTY);
         }
--- a/src/test/java/com/rocketden/main/api/GameTests.java
+++ b/src/test/java/com/rocketden/main/api/GameTests.java
@@ -31,12 +31,8 @@ import com.rocketden.main.model.problem.ProblemDifficulty;
 import com.rocketden.main.model.problem.ProblemIOType;
 import com.rocketden.main.service.SubmitService;
 import com.rocketden.main.util.UtilityTestMethods;
-import com.rocketden.main.service.ProblemService;
 
-import org.junit.Assert;
 import org.junit.jupiter.api.Test;
-import org.mockito.InjectMocks;
-import org.mockito.Spy;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
@@ -98,11 +94,6 @@ public class GameTests {
 
     // Predefine notification content.
     private static final String CONTENT = "[1, 2, 3]";
-
-    //ProblemService for editing the problem
-    @Spy
-    @InjectMocks
-    private ProblemService problemService;
 
     // Helper method to start the game for a given room
     private void startGameHelper(RoomDto room, UserDto host) throws Exception {
--- a/src/test/java/com/rocketden/main/api/GameTests.java
+++ b/src/test/java/com/rocketden/main/api/GameTests.java
@@ -124,7 +124,7 @@ public class GameTests {
         createProblemRequest.setName(NAME);
         createProblemRequest.setDescription(DESCRIPTION);
         createProblemRequest.setDifficulty(ProblemDifficulty.EASY);
-
+        createProblemRequest.setApproval(true);
         List<ProblemInputDto> problemInputs = new ArrayList<>();
         ProblemInputDto problemInput = new ProblemInputDto(INPUT_NAME, IO_TYPE);
         problemInputs.add(problemInput);
--- a/src/test/java/com/rocketden/main/api/GameTests.java
+++ b/src/test/java/com/rocketden/main/api/GameTests.java
@@ -159,6 +159,7 @@ public class GameTests {
 
         String testCaseJsonResponse = testCaseResult.getResponse().getContentAsString();
         ProblemTestCaseDto testCaseActual = UtilityTestMethods.toObject(testCaseJsonResponse, ProblemTestCaseDto.class);
+        problemActual.setTestCases(Collections.singletonList(testCaseActual));
 
         assertEquals(INPUT, testCaseActual.getInput());
         assertEquals(OUTPUT, testCaseActual.getOutput());
@@ -177,11 +178,6 @@ public class GameTests {
         problemDto.setOutputType(ProblemIOType.CHARACTER);
         problemDto.setName(NAME);
         problemDto.setApproval(true);
-
-        ProblemTestCaseDto testCaseDto = new ProblemTestCaseDto();
-        testCaseDto.setInput(INPUT);
-        testCaseDto.setOutput("a");
-        problemDto.setTestCases(Collections.singletonList(testCaseDto));
 
         // Edit problem with new values
         String endpoint = String.format(PUT_PROBLEM_EDIT, problemDto.getProblemId());
--- a/src/test/java/com/rocketden/main/api/GameTests.java
+++ b/src/test/java/com/rocketden/main/api/GameTests.java
@@ -124,7 +124,6 @@ public class GameTests {
         createProblemRequest.setName(NAME);
         createProblemRequest.setDescription(DESCRIPTION);
         createProblemRequest.setDifficulty(ProblemDifficulty.EASY);
-        createProblemRequest.setApproval(true);
         List<ProblemInputDto> problemInputs = new ArrayList<>();
         ProblemInputDto problemInput = new ProblemInputDto(INPUT_NAME, IO_TYPE);
         problemInputs.add(problemInput);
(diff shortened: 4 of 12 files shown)


src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -169,21 +173,23 @@ public ProblemDto deleteProblem(String problemId) {
...
+        }
+
         List<Problem> problems;
         if (difficulty == ProblemDifficulty.RANDOM) {
-            problems = repository.findAll();
+            //Change to findAllByApproval once it works
+            problems = repository.findAllByApproval(true);
         } else {
-            problems = repository.findAllByDifficulty(difficulty);
+            //Change to findAllByDifficultyAndApproval once it works
```
alankbi, line 185: nit: remove these comments now that the feature has been implemented

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -124,7 +124,7 @@ private ProblemDto createSingleProblemAndTestCases() throws Exception {
         createProblemRequest.setName(NAME);
         createProblemRequest.setDescription(DESCRIPTION);
         createProblemRequest.setDifficulty(ProblemDifficulty.EASY);
-
+        createProblemRequest.setApproval(true);
```
alankbi, line 127: nit: here and for the other 1-3 createSingleProblem methods, could you add at the bottom something like `assertTrue(response.getApproval)`? Just to verify that this was properly set in the database

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -173,10 +173,10 @@ public void getProblemsSuccess() {
         problem.setName(NAME);
         problem.setDescription(DESCRIPTION);
         problem.setDifficulty(ProblemDifficulty.EASY);
-
+        problem.setApproval(true);
         List<Problem> expected = new ArrayList<>();
         expected.add(problem);
-
+        //do findAllByApproval once it works
```
alankbi, line 179: nit: remove comment (and the other new comments/commented out lines in the file)

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -452,6 +452,30 @@ public void editProblemBadTestCaseOutput() {
...
+        problem.setOutputType(ProblemIOType.STRING);
+        ProblemInput problemInput = new ProblemInput(INPUT_NAME, IO_TYPE);
+        problem.addProblemInput(problemInput);
+
+        Mockito.doReturn(problem).when(repository).findProblemByProblemId(problem.getProblemId());
+
+        ProblemDto updatedProblem = ProblemMapper.toDto(problem);
+        updatedProblem.setApproval(true);
+
+        ArrayList<ProblemTestCaseDto> emptyCases = new ArrayList<ProblemTestCaseDto>();
```
alankbi, line 470: small nit but a suggestion would be to use List as the main type instead of ArrayList

```suggestion
        List<ProblemTestCaseDto> emptyCases = new ArrayList<ProblemTestCaseDto>();
```

src/main/java/com/rocketden/main/model/problem/Problem.java
```diff
@@ -70,4 +70,8 @@ public void addProblemInput(ProblemInput problemInput) {
         problemInputs.add(problemInput);
         problemInput.setProblem(this);
     }
+
+    public void toggleApprovedStatus() {
+        approval ^= true;
+    }
```
alankbi, line 76: nit: I don't think this is used anywhere, so I think we can get rid of it

alankbi (changes requested):
Awesome work on the first PR! I requested changes mostly because of small nits here and there (plus the nit below), but otherwise the PR as a whole looks great 👍 

nit: for 1-2 tests in ProblemTests and ProblemService tests (maybe `createEditDeleteProblemSuccess` and `editProblemSuccess` respectively), could you add a line setting approval to true/false, and an assertTrue/assertFalse at the end ensuring the value was set properly?

I also wanted to suggest adding a test in GameTests for `startGameApprovedProblemNotFound`, but in my PR I'm changing some of that behavior so it might be easier if I handle that test case after this is merged into main


Also, two side notes: 

1. I think you might need to configure your GitHub/local git username or email cause your commits right now are showing up as anonymous
2.  
>Added an exception that prevents people from approving a problem through editProblem if it has zero test cases (Note: not yet tested if this actually works. It's possible that an empty test case will still make the index of the test case arrayList 1 and bypass this error. Can test in production)

One recommendation I have would be to download Postman, which is what we use to test the backend routes (I think a while ago we sent an email invite for our workspace which has all the example requests saved). This helps us test things before production and verify that it all works as expected

src/main/java/com/rocketden/main/service/ProblemService.java
sean-xcamp: k

src/test/java/com/rocketden/main/api/GameTests.java
sean-xcamp: k

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
sean-xcamp: k

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
sean-xcamp: k

src/main/java/com/rocketden/main/model/problem/Problem.java
sean-xcamp: k

zpChris:
> ### List of Changes:
> * Added problem approval boolean to model/Problem.java
> * Added toggleApproval() setter method in model/Problem.java
> * Changed getProblemByDifficulty method to check approval in addition to difficulty (only approved problems can be picked)
> * Changed ProblemService, GameSocket, Game, and Problem tests by setting every problem to approved when it's created
> * Added an exception that prevents people from approving a problem through editProblem if it has zero test cases (Note: not yet tested if this actually works. It's possible that an empty test case will still make the index of the test case arrayList 1 and bypass this error. Can test in production)
> 
> ### Notes:
> * Tests passed (phew)

Wanted to note that for where you reference file or method names you can place `` around the lines like follows, \`model/Problem.java\`, which makes them code-formatted: `model/Problem.java`.

src/main/java/com/rocketden/main/service/ProblemService.java
zpChris: My first though on all the responses lol: [https://www.youtube.com/watch?v=07TinrpKG6o&t=70s](https://www.youtube.com/watch?v=07TinrpKG6o&t=70s).

src/main/java/com/rocketden/main/model/problem/Problem.java
```diff
@@ -70,4 +70,5 @@ public void addProblemInput(ProblemInput problemInput) {
         problemInputs.add(problemInput);
         problemInput.setProblem(this);
     }
+
```
zpChris, line 73: nit: Remove line.

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -104,7 +105,9 @@ public ProblemDto editProblem(String problemId, ProblemDto updatedProblem) {
                 || updatedProblem.getOutputType() == null) {
             throw new ApiException(ProblemError.EMPTY_FIELD);
         }
-
+        if (updatedProblem.getApproval() && updatedProblem.getTestCases().size() == 0) {
+            throw new ApiException(ProblemError.BAD_APPROVAL);
+        }
```
zpChris, line 110: very small nit: Add lines around this if statement to separate them from the above for and below if statements.

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -104,7 +105,9 @@ public ProblemDto editProblem(String problemId, ProblemDto updatedProblem) {
                 || updatedProblem.getOutputType() == null) {
```
zpChris, line 105: Include a check to see if `updatedProblem.getApproval() == null`.

src/main/java/com/rocketden/main/service/ProblemService.java
```diff
@@ -169,21 +173,21 @@ public ProblemDto deleteProblem(String problemId) {
             throw new ApiException(ProblemError.EMPTY_FIELD);
         }
 
+        if (numProblems <= 0) {
```
zpChris, line 176: Thanks for adding this! 👍

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -276,8 +276,8 @@ public void getRandomProblemMediumDifficulty() {
         Problem problem1 = new Problem();
         problem1.setDifficulty(ProblemDifficulty.MEDIUM);
         List<Problem> problems = Collections.singletonList(problem1);
-
-        Mockito.doReturn(problems).when(repository).findAllByDifficulty(ProblemDifficulty.MEDIUM);
+        // Should be findAllByDifficultyAndApproval()
```
zpChris, line 279: Is there a reason you left this comment in? Can you remove it?

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -332,8 +330,7 @@ public void getRandomProblemZeroNumProblems() {
         Problem problem1 = new Problem();
         problem1.setDifficulty(ProblemDifficulty.MEDIUM);
         List<Problem> problems = Collections.singletonList(problem1);
-
-        Mockito.doReturn(problems).when(repository).findAll();
+        //Mockito.doReturn(problems).when(repository).findAllByApproval(true);
```
zpChris, line 333: I'm not sure why you removed this, so maybe re-examine that - either way, either delete the line or un-comment this out in the code.

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -347,7 +344,7 @@ public void getRandomProblemNegativeNumProblems() {
         problem1.setDifficulty(ProblemDifficulty.MEDIUM);
         List<Problem> problems = Collections.singletonList(problem1);
 
-        Mockito.doReturn(problems).when(repository).findAll();
+        //Mockito.doReturn(problems).when(repository).findAll();
```
zpChris, line 347: Same as above (I'm not sure why you removed this, so maybe re-examine that - either way, either delete the line or un-comment this out in the code).

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
```diff
@@ -452,6 +449,30 @@ public void editProblemBadTestCaseOutput() {
...
+        ProblemInput problemInput = new ProblemInput(INPUT_NAME, IO_TYPE);
+        problem.addProblemInput(problemInput);
+
+        Mockito.doReturn(problem).when(repository).findProblemByProblemId(problem.getProblemId());
+
+        ProblemDto updatedProblem = ProblemMapper.toDto(problem);
+        updatedProblem.setApproval(true);
+
+        List<ProblemTestCaseDto> emptyCases = new ArrayList<ProblemTestCaseDto>();
+        updatedProblem.setTestCases(emptyCases);
```
zpChris, line 468: Just want to note that this shouldn't be necessary, as the test cases are, by default, set to an empty `ArrayList`.

src/test/java/com/rocketden/main/util/SocketTestMethods.java
```diff
@@ -70,7 +70,7 @@ public static void createSingleProblemAndTestCases(TestRestTemplate template, in
         createProblemRequest.setName(NAME);
         createProblemRequest.setDescription(DESCRIPTION);
         createProblemRequest.setDifficulty(ProblemDifficulty.EASY);
-
+        createProblemRequest.setApproval(true);
```
zpChris, line 73: nit: When you add lines, thoughts on keeping these line spacers between different segments of the code? There are a good number of other places with this as well, if you're open to it you could add those blank lines between the different code sections.

src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
```diff
@@ -13,6 +13,7 @@
 public class CreateProblemRequest {
     private String name;
     private String description;
+    private Boolean approval;
```
zpChris, line 16: I just realized that if we include `approval` in the `CreateProblemRequest`, we break one of the invariants that was set in this PR - you added a check in the `editProblem` method that the user cannot set the problem to be approved if there are no test cases. However, through creating the problem, we currently have it set up such that you cannot add any test cases until it's actually created, after which you can. You can see that reflected on our UI in the image below:

Create Problem Page

Since it'd be a long fix, I propose that we just include the `approval` boolean for the `editProblem` endpoint. Sorry @alankbi and I didn't identify this earlier! We could enable the ability to add test cases when creating the problem, but that would take a lot more work and we had reasons for preventing this in the first place (we wanted people to first verify that they had the problem inputs set).

zpChris (changes requested):
Awesome job on the first PR! I left some nits and requested changes, but that's all a part of the process, and won't be the case for every PR (just cause this is the first) 👍 

I identified a potential invariant violation that I outlined in [this comment](https://github.com/rocketden/main/pull/131#discussion_r606728850) - sorry that neither @alankbi nor I saw that initially! This does require updating the PR, but hopefully that's actually a fairly easy task. Let me know if you have any questions!

zpChris:
[Add approved boolean to problems to allow for review process](https://trello.com/c/Ylo486wp/318-add-approved-boolean-to-problems-to-allow-for-review-process)

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
sean-xcamp: I wasn't sure myself, but I reran this test and got some "unncessary stubbing" error at those two lines with the `Mockito.doReturn(...)` function; I vaguely recall working through this with Alan and he said the lines weren't necessary but I'm not 100% certain about that. Either way the test passes without this line so I'll delete it for now
Screen Shot 2021-04-05 at 10 45 24 PM

src/test/java/com/rocketden/main/util/SocketTestMethods.java
sean-xcamp: Will definitely keep this in mind for future PRs! I looked through some of the files I modified and couldn't find many instances of this, though I'm 100% sure they exist; sorry about that

src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
sean-xcamp: Got it, I just delete that `Boolean approval` line I added in `CreateProblemRequest right, or is there more to this fix?

alankbi: It might be slightly more challenging than that because in the `createSingleProblem`-type methods (the ones in the test files), it's currently just setting approval to true on creation of the problem, so in order to still have that feature (which allows the tests to pass), you'd need to add an additional edit problem request to set the approval to true.

zpChris:
I went to review this (sorry, after quite the delay!) but I don't see any additional commits - is it possible you forgot to push them to this branch @sean-xcamp?

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
zpChris: I think I still see the commented-out Mockito lines - can you remove those?

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
zpChris: Same here - just delete that comment?

src/test/java/com/rocketden/main/util/SocketTestMethods.java
zpChris: No problem, all good 👍

src/main/java/com/rocketden/main/dto/problem/CreateProblemRequest.java
zpChris: I don't think you've made this change yet @sean-xcamp - the list I think is this:
1. Remove the `approval` field in `CreateProblemRequest.java `.
2. Remove all the relevant approval code for `createProblem` in `ProblemService.java`. (You can look at exactly this link - [https://github.com/rocketden/main/pull/131/files](https://github.com/rocketden/main/pull/131/files) - to simply remove the changes one-by-one.)
3. Remove the approval setting in `SocketTestMethods.java`.

There may be others but that's all I could find, I suggest looking at the same "Files Changed" link [here](https://github.com/rocketden/main/pull/131/files) to revert the relevant changes.

zpChris (changes requested):
Just requesting review so that you can finish the changes from last time - I would usually approve but want to make sure those updates get made. Should be really quick, great job thus far 👍

src/test/java/com/rocketden/main/service/ProblemServiceTests.java
sean-xcamp: > I think I still see the commented-out Mockito lines - can you remove those @sean-xcamp? (There are two lines I see 👍 )

Yep, thanks for pointing that out; not sure how I missed it! I included it in the next commit

zpChris (approved):
LGTM! ✅ 

Full disclosure I didn't re-run all the tests because this looks good, but maybe @alankbi when you end up reviewing it you can run that command? Otherwise np - thanks @sean-xcamp! 👍 

Also, for the PRs following this one, I think likely just @alankbi OR me (@zpChris) could act as reviewers, and maybe @sean-xcamp and @jasonz6688 you can start to review PRs as well, as that will be helpful training for when we may add more people in the future.

alankbi:
Why the tests are most likely failing: when you removed the approval setting from the create problem request, I think it most likely re-introduced the original bug where the helper methods `createSingleProblem` creates an unapproved problem, and none of the `getProblemByDifficultyAndApproval` tests are able to find a problem in the tests. 

So to solve this, I think you'll probably have to edit each of the `createSingleProblem`-type methods to include a second `UpdateSettingsRequest` call that sets the newly created problem to be approved.

zpChris:
> Why the tests are most likely failing: when you removed the approval setting from the create problem request, I think it most likely re-introduced the original bug where the helper methods `createSingleProblem` creates an unapproved problem, and none of the `getProblemByDifficultyAndApproval` tests are able to find a problem in the tests.
> 
> So to solve this, I think you'll probably have to edit each of the `createSingleProblem`-type methods to include a second `UpdateSettingsRequest` call that sets the newly created problem to be approved.

It would be smart to make a helper method that is `createSingleApprovedProblem` so this change is only made once.

Sorry I didn't notice that when approving the PR initially! Thanks for catching that @alankbi

sean-xcamp:
> > Why the tests are most likely failing: when you removed the approval setting from the create problem request, I think it most likely re-introduced the original bug where the helper methods `createSingleProblem` creates an unapproved problem, and none of the `getProblemByDifficultyAndApproval` tests are able to find a problem in the tests.
> > So to solve this, I think you'll probably have to edit each of the `createSingleProblem`-type methods to include a second `UpdateSettingsRequest` call that sets the newly created problem to be approved.
> 
> It would be smart to make a helper method that is `createSingleApprovedProblem` so this change is only made once.

This would result in a lot of duplicate code right @zpChris would that be fine?

zpChris:
> This would result in a lot of duplicate code right @zpChris would that be fine? 

Sorry not sure I follow - wouldn't my solution result in **less** duplicate code, since you're placing the duplicate code about editing the problem to change its approval in a helper method?

sean-xcamp:
> > This would result in a lot of duplicate code right @zpChris would that be fine?
> 
> Sorry not sure I follow - wouldn't my solution result in **less** duplicate code, since you're placing the duplicate code about editing the problem to change its approval in a helper method?

My understanding was that there would be a `createSingleProblem()` method and **another** `createSingleApprovedProblem()` method, with almost all lines of code shared between those two. I'm not exactly sure if I follow if otherwise.

zpChris:
> My understanding was that there would be a `createSingleProblem()` method and **another** `createSingleApprovedProblem()` method, with almost all lines of code shared between those two. I'm not exactly sure if I follow if otherwise.

Ah - what you could do is in the `createSingleApprovedProblem()` method, you could call the  `createSingleProblem()` method to avoid that duplicate code.

sean-xcamp:
> > My understanding was that there would be a `createSingleProblem()` method and **another** `createSingleApprovedProblem()` method, with almost all lines of code shared between those two. I'm not exactly sure if I follow if otherwise.
> 
> Ah - what you could do is in the `createSingleApprovedProblem()` method, you could call the `createSingleProblem()` method to avoid that duplicate code.

Got it, that clarification makes a lot more sense. Will do.

zpChris:
Ok - here's the list of fixes (I found these all through debugging, so that's a _super_ important tool):
- The `RoomSocketTests` were failing because the `createSingleProblemAndTestCases` never actually attached the test cases to the **returned** `problemActual`. Thus, when you tried to update the problem and change the setting to approve, it also registered it as deleting the test cases and setting it to approved. Since you cannot have an approved problem without test cases, this threw an error.
  - You initially tried to solve this issue by adding test cases to the `createSingleApprovedProblemAndTestCases` method; however, the output you set was `"a"`, which was an invalid type since it was supposed to be an Array of Integers, thus it threw an error.
- The `socketReceivesMessageOnStartGame` in `RoomSocketTests` had to change `createSingleProblemAndTestCases` to `createSingleApprovedProblemAndTestCases`.

src/main/java/com/rocketden/main/model/problem/Problem.java
```diff
@@ -31,8 +31,8 @@
 
     // Auto-generate default business ID for each problem
     private String problemId = UUID.randomUUID().toString();
-
     private String name;
+    private Boolean approval = false;
     @Column(columnDefinition = "TEXT")
```
alankbi, line 36: nit: add a line between these two

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -93,9 +99,14 @@
     // Predefine notification content.
     private static final String CONTENT = "[1, 2, 3]";
 
+    //ProblemService for editing the problem
+    @Spy
+    @InjectMocks
+    private ProblemService problemService;
+
```
alankbi, line 106: nit: delete this since it's not being used

src/test/java/com/rocketden/main/api/GameTests.java
```diff
@@ -163,10 +172,39 @@ private ProblemDto createSingleProblemAndTestCases() throws Exception {
...
+    private ProblemDto createSingleApprovedProblemAndTestCases() throws Exception {
+        ProblemDto problemDto = createSingleProblemAndTestCases();
+        problemDto.setOutputType(ProblemIOType.CHARACTER);
+        problemDto.setName(NAME);
+        problemDto.setApproval(true);
+
+        ProblemTestCaseDto testCaseDto = new ProblemTestCaseDto();
+        testCaseDto.setInput(INPUT);
+        testCaseDto.setOutput("a");
+        problemDto.setTestCases(Collections.singletonList(testCaseDto));
```
alankbi, line 193: Just wanted to double-check - do we need to be adding a test case in this method? I thought the `createSingleProblemAndTestCases` method returns a ProblemDto that already has a test case

Same thing for the other `createSingleApprovedProblemAndTestCases` method

alankbi (approved):
Approved! However I left a few small nits, so be sure to take a look at those before merging :)

src/test/java/com/rocketden/main/api/GameTests.java
sean-xcamp: previously we tried running these tests without this and it failed, so I'm pretty sure it's necessary. It's a little bit odd.

zpChris: See [this](https://github.com/rocketden/main/pull/131#issuecomment-817464717) as to why that was @sean-xcamp - I don't think it's necessary anymore.

PR #132, opened 2021-04-03
Check valid parameter names for problems on frontend and backend.

zpChris:
### List of Changes:
- Add frontend check using the [valid-identifier](https://www.npmjs.com/package/valid-identifier) npm module, and add `HoverTooltip` to the create and edit problem pages for invalid parameter names.
- Add backend verification for Java through [`isIdentifier`, `isKeyword`, and `isName`](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/SourceVersion.html), as well as checking against the list of case-sensitive [Python keywords](https://www.programiz.com/python-programming/keyword-list).
- Add parameterized integration and unit tests.

### Notes:
- I'm not sure why, but two of the commits added and then removed 17,517 lines - I think @jasonz6688 may have encountered this, or maybe this is the same add / removal that @alankbi had in his PR? Either way, I think I'll likely make a new PR by copying over the changed files one-by-one.
- There is a bug right now (only in this PR) causing the frontend "Difficulty" buttons to toggle rapidly on hover - I'm not sure why this is, but I'll try to fix the bug when creating the new PR.
- I made an additional restriction that all parameter names must start with a letter (using [this method](https://stackoverflow.com/a/32567789/7517518)), as I thought this would improve readability.
- As we add more languages, we may need to check against additional keywords. I thought of checking valid variable name using the [Python method](https://stackoverflow.com/questions/36330860/pythonically-check-if-a-variable-name-is-valid) as well, but I thought that may be too complicated to incorporate.

### Screencast and Images:

[Invalid Parameter Names on Problem Pages](https://www.loom.com/share/c958a1a6a74b4839b6a8ec027b8463a9)

**Create Page Invalid Parameter with Hover**
Create Page Invalid Parameter

**Edit Page Invalid Parameter with Error**
Edit Page Invalid Parameter with Error


zpChris:
Not requesting review yet, as I have to fix a small bug and then undo the 17,517 lines added / removed.

zpChris:
[Prevent users from entering invalid keywords when creating problems (e.g. int, class)](https://trello.com/c/3TaRVpie/250-prevent-users-from-entering-invalid-keywords-when-creating-problems-eg-int-class)

zpChris:
Opened instead #133 by copying files 1-by-1, to avoid the +/- 17,517 lines in this PR.

PR #133, opened 2021-04-04
Check valid parameter names for problems on frontend and backend.

zpChris:
### List of Changes:
- Add frontend check using the [valid-identifier](https://www.npmjs.com/package/valid-identifier) npm module, and add `HoverTooltip` to the create and edit problem pages for invalid parameter names.
- Add backend verification for Java through [`isIdentifier`, `isKeyword`, and `isName`](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/SourceVersion.html), as well as checking against the list of case-sensitive [Python keywords](https://www.programiz.com/python-programming/keyword-list).
- Add parameterized integration and unit tests.

### Notes:
- I closed a duplicate #132 PR because it added and removed 17,517 lines; I noticed this was because when I ran `npm i valid-identifier` in the `frontend` folder, it removed those lines, but when I ran `./mvnw spring-boot:run`, it added them back in. I'm not sure why this might be, but to fix that problem I did both actions in the same commit, so there shouldn't be any accidentally very large commits in this PR.
- I changed how `HoverTooltip` displays as the `InlineErrorIcon` is never `disabled`, so I can use `onMouseEnter`, `onMouseMove`, and `onMouseLeave` directly on that object. This fixed the bug which caused the frontend "Difficulty" buttons to toggle rapidly on hover, which I think may have been because the state was rapidly refreshing.
  - When using the hover tooltip on other pages, it's important to note that if something is not disabled, then the outer `div` and absolutely-positioned hover element are not necessary.
- On the `mouseMoveHandler` function, I use a type `any` because I could not get the types for that function to work - it works with `any` though, so as long as that's ok this should work out.
- I made an additional restriction that all parameter names must start with a letter (using [this method](https://stackoverflow.com/a/32567789/7517518)), as I thought this would improve readability.
- As we add more languages, we may need to check against additional keywords. I thought of checking valid variable name using the [Python method](https://stackoverflow.com/questions/36330860/pythonically-check-if-a-variable-name-is-valid) as well, but I thought that may be too complicated to incorporate.

### Screencast and Images:

[Invalid Parameter Names on Problem Pages](https://www.loom.com/share/c958a1a6a74b4839b6a8ec027b8463a9)

**Create Page Invalid Parameter with Hover**
Create Page Invalid Parameter

**Edit Page Invalid Parameter with Error**
Edit Page Invalid Parameter with Error

### DIFF
--- a/frontend/src/util/Utility.ts
+++ b/frontend/src/util/Utility.ts
@@ -3,6 +3,9 @@ import { removeUser } from '../api/Room';
 import { disconnect } from '../api/Socket';
 import { errorHandler } from '../api/Error';
 
+// Require validator for identifiers as no types are provided.
+const validateIdentifier = require('valid-identifier');
+
 /**
  * Check whether all keys in param exist in location.state
  */
@@ -50,3 +53,17 @@ export const leaveRoom = (history: any, roomId: string, user: User | null) => {
     });
   }
 };
+
+// Return true iff the character is a letter.
+const isLetter = (character: string) => {
+  if (character.length !== 1) {
+    return false;
+  }
+
+  // Character is a letter iff the upper case does not equal lower case.
+  return character.toUpperCase() !== character.toLowerCase();
+};
+
+// Return true iff the variable is a valid identifier, and starts with a letter.
+export const validIdentifier = (identifier: string) => validateIdentifier(identifier)
+  && isLetter(identifier.charAt(0));
--- a/src/test/java/com/rocketden/main/api/ProblemTests.java
+++ b/src/test/java/com/rocketden/main/api/ProblemTests.java
@@ -286,7 +286,7 @@ class ProblemTests {
     }
 
     @ParameterizedTest
-    @ValueSource(strings = {"", "rocket rocket", "12", "$Hello", "jimmy=neutron"})
+    @ValueSource(strings = {"finally", "void", "throw", "EP<>", "new"})
     public void editProblemInvalidIdentifier(String inputName) throws Exception {
         ProblemDto problemDto = createSingleProblem();
 
@@ -340,7 +340,7 @@ class ProblemTests {
     }
 
     @ParameterizedTest
-    @ValueSource(strings = {"", "rocket rocket", "12", "$Hello", "jimmy=neutron"})
+    @ValueSource(strings = {"False", "try", "jeremy ", "-minus", "@annotation"})
     public void createProblemInvalidIdentifier(String inputName) throws Exception {
         CreateProblemRequest request = new CreateProblemRequest();
         request.setName(NAME);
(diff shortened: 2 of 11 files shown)


zpChris:
[Prevent users from entering invalid keywords when creating problems (e.g. int, class)](https://trello.com/c/3TaRVpie/250-prevent-users-from-entering-invalid-keywords-when-creating-problems-eg-int-class)

frontend/src/util/Utility.ts
```diff
@@ -3,6 +3,9 @@ import { removeUser } from '../api/Room';
 import { disconnect } from '../api/Socket';
 import { errorHandler } from '../api/Error';
 
+// Require validator for identifiers as no types are provided.
+const validateIdentifier = require('valid-identifier');
```
alankbi, line 7: When I accepted IntelliJ's suggestion to change it into a default import, I think I got the error you're referring to, but it also provided this second suggestion: "Try `npm install @types/valid-identifier` if it exists". Do you know if this works or not?

frontend/src/util/Utility.ts
```diff
@@ -50,3 +53,17 @@ export const leaveRoom = (history: any, roomId: string, user: User | null) => {
...
 };
+
+// Return true iff the character is a letter.
+const isLetter = (character: string) => {
+  if (character.length !== 1) {
+    return false;
+  }
+
+  // Character is a letter iff the upper case does not equal lower case.
+  return character.toUpperCase() !== character.toLowerCase();
```
alankbi, line 64: clever lol

src/test/java/com/rocketden/main/api/ProblemTests.java
```diff
@@ -283,6 +285,34 @@ public void createEditDeleteProblemSuccess() throws Exception {
         assertEquals(ERROR.getResponse(), response);
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {"", "rocket rocket", "12", "$Hello", "jimmy=neutron"})
```
alankbi, line 289: For each of these different places you use the parameterized test, could you use different strings? For example instead of the same 5 strings you could cover like 20 or so different examples to minimize the chances of any edge cases we didn't think about (also testing like the reserved keywords and all)

alankbi (approved):
Approved! I left like 1-2 small nits, but they should be pretty quick to implement so this looks good 👍

frontend/src/util/Utility.ts
zpChris: Yeah unfortunately, that is not actually in the npm registry - I tried that initially but apparently no types are implemented, so I used this new syntax to circumvent that issue.

src/test/java/com/rocketden/main/api/ProblemTests.java
zpChris: Sure, that's a good idea! Done ✅

PR #134, opened 2021-04-04
Fix tester bug, update database, new faster build option

alankbi:
### List of Changes:

* Fix failure to connect to tester service bug caused by special characters in some problem descriptions
  * This is done by setting all the non-required fields to null before initiating the POST request
* Update relevant tests to check for null fields
* Potentially fix max questions error by switching from ClearDB to Google Cloud SQL (@zpChris and I)
* Add ability to skip building the frontend by running `./mvnw spring-boot:run -Dfrontend.skip`
  * This command takes only a few seconds to finish as opposed to several minutes

### Notes:

* This previously occurred on the LCS problem, although it's possible other problems had the same issue too

_Note: This PR is extended off of #128 but set to be directly merged into main instead (and with additional updates)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,22 +20,29 @@ with an explanation if you would like access to this for some reason. Otherwise,
 replace the `spring.datasource` variables in `application.properties` with your
 own database to proceed. 
 
-Run `./mvnw clean install` to build.
+Run `./mvnw clean install` if it's your first time running the app (or for whatever
+reason you need to re-download Maven dependencies). 
 
-Run `./mvnw test` to run tests.
+Run `./mvnw test` to run the project's unit tests.
 
-Run `./mvnw spring-boot:run` to start the server.
+Run `./mvnw spring-boot:run` to build and start the server (both frontend and backend).
 
 Once the server is started, the site should be running at 
 [localhost:8080](http://localhost:8080), and you can send REST requests to the routes
 at `localhost:8080/api/v1/`. 
 
-Note: the server startup process can take some time, so if you are looking to test
+**Tip 1**: the full server startup process can take some time, so if you are looking to test
 some quick frontend changes, you can start `./mvnw spring-boot:run` in one bash window
 and run `cd ./frontend && npm start` in another. You can then use 
 [localhost:3000](http://localhost:3000) instead, and any changes to frontend files will
 immediately be reflected (so you don't need to re-run Maven every time). Be aware that 
 occasional socket connection errors can occur when using this method. 
+
+**Tip 2**: On the other hand, if you're only testing backend changes, you can avoid 
+re-building the entire frontend by running `./mvnw spring-boot:run -Dfrontend.skip`.
+This will reuse the frontend code built from the latest call to `./mvnw spring-boot:run`
+(so if you're previous build failed, you may need to rerun the full command before this
+tip will work). 
 
 ## File Structure
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ own database to proceed.
 * Run `./mvnw clean install` if it's your first time running the app (or for whatever
 reason you need to re-download Maven dependencies). 
 
-* Run `./mvnw test` to run the project's unit tests.
+* Run `./mvnw test` to run the project's unit and API tests.
 
 * Run `./mvnw spring-boot:run` to build and start the server (both frontend and backend).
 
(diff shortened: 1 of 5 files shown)


alankbi:
[An internal error occurred connecting to the tester service](https://trello.com/c/EzfKSMkv/317-an-internal-error-occurred-connecting-to-the-tester-service)

alankbi:
Discussion from #128:

> * How did you know that this was the problem causing the tester connection to fail? (comment by @zpChris)

It took a decent amount of trial and error, but basically after manually setting those fields to null before sending the request, the bug went away

> * When you return `"dksoe"` on the LCS problem, it still doesn't label that solution as correct. I'm not sure why that would be, since I thought we fixed that error with the latest tester PR.

I just checked and I think it's cause the problem had special `“ ” ` quotes instead of `""` - after updating the problem it works as expected

> * I might recommend that you merge it directly into main as you mentioned - this shouldn't take too long to move into a separate PR (or perhaps to remove the changes from `ChooseProblem` here, not sure which), and would let us merge it in faster. I'm ok waiting if you'd rather just merge it through after `ChooseProblem` though.

Sure that works

alankbi:
[max_questions and poolSize errors](https://trello.com/c/5nUsP8r4/312-maxquestions-and-poolsize-errors)

alankbi:
For some reason when running the tests we now get a `JDBCConnectionException: Unable to release JDBC Connection` error - the tests generally speaking still pass though. It looks like the same issue is on main so idk if it's always been like this and we haven't noticed or something came up at some point that we missed.

README.md
```diff
@@ -20,23 +20,35 @@ with an explanation if you would like access to this for some reason. Otherwise,
...
 [localhost:8080](http://localhost:8080), and you can send REST requests to the routes
 at `localhost:8080/api/v1/`. 
 
-Note: the server startup process can take some time, so if you are looking to test
+### Speed Up Build Times
+
+The following two tips can dramatically speed up build times and save you a lot of
+time when developing (a matter of minutes vs. seconds on each code change):
+
+**Tip 1**: the full server startup process can take some time, so if you are looking to test
```
zpChris, line 39: nit: Capitalize "the" to "The", or make the first word on "Tip 2" ("On") lowercase for consistency.

README.md
```diff
@@ -20,23 +20,35 @@ with an explanation if you would like access to this for some reason. Otherwise,
...
 [localhost:3000](http://localhost:3000) instead, and any changes to frontend files will
 immediately be reflected (so you don't need to re-run Maven every time). Be aware that 
 occasional socket connection errors can occur when using this method. 
 
+**Tip 2**: On the other hand, if you're only testing backend changes, you can avoid 
+re-building the entire frontend by running `./mvnw spring-boot:run -Dfrontend.skip`.
+This will reuse the frontend code built from the latest call to `./mvnw spring-boot:run`
+(so if you're previous build failed, you may need to rerun the full command before this
+tip will work). 
+
```
zpChris, line 51: Could you combine Tip 1 with Tip 2 to always very quickly develop on `localhost:3000` (albeit with occasional socket connection errors)?

README.md
```diff
@@ -20,23 +20,35 @@ with an explanation if you would like access to this for some reason. Otherwise,
 replace the `spring.datasource` variables in `application.properties` with your
 own database to proceed. 
 
-Run `./mvnw clean install` to build.
+* Run `./mvnw clean install` if it's your first time running the app (or for whatever
+reason you need to re-download Maven dependencies). 
 
-Run `./mvnw test` to run tests.
+* Run `./mvnw test` to run the project's unit tests.
```
zpChris, line 26: It's not just the unit tests, right? Doesn't this run the API / integration tests as well?

pom.xml
```diff
@@ -165,6 +165,8 @@
         <configuration>
           <workingDirectory>frontend</workingDirectory>
           <installDirectory>target</installDirectory>
+          <!-- Skip this plugin step if skip property is set to true -->
+          <skip>${frontend.skip}</skip>
```
zpChris, line 169: Thanks so much for adding this! Fantastic idea, and I think it'll really pay off 💯

zpChris (approved):
LGTM! ✅ 

Very excited to see if migrating to GCP fixes the `max_questions` error, and these faster build options are very very helpful. Looks amazing! 💯 ✨

README.md
alankbi: Most likely yeah - I can add a note about that (maybe a tip 3)

README.md
alankbi: Yeah - I'll update this to say "and API tests"

PR #136, opened 2021-04-06
Spectator boolean

sean-xcamp:
### List of Changes:

- Added `setSpectator()` POST request in `RoomController`/`RoomService`
- Add `isSpectator` Boolean in `UserDto` and `User`; default set to `null`
- Create `SetSpectatorRequest` with two UserDto objects: initiator and receiver
- Error handling: if the initiator is different from the receiver and the initiator is not the host, throw `RoomError.INVALID_PERMISSIONS` error
- Wrote `setSpectatorSuccess()` and `setSpectatorFailure()` test in `RoomServiceTests`. Success: initiator is host or initiator is the same user as receiver; Failure: initiator is non-host and the receiver is not the initiator

### Notes:

Two newly-written tests passed


src/main/java/com/rocketden/main/controller/v1/RoomController.java
```diff
@@ -67,4 +69,9 @@ public RoomController(RoomService service) {
                                                       @RequestBody UpdateSettingsRequest request) {
         return new ResponseEntity<>(service.updateRoomSettings(roomId, request), HttpStatus.OK);
     }
+
+    @PostMapping("/rooms/{roomId}/spectator/{isSpectator}")
+    public ResponseEntity<RoomDto> setSpectator(@PathVariable String roomId, @PathVariable Boolean isSpectator, @RequestBody SetSpectatorRequest request) {
+        return new ResponseEntity<>(service.setSpectator(roomId, isSpectator, request), HttpStatus.OK);
+    }
```
zpChris, line 76: Thinking more, I'm somewhat split on how this current PR is designed, with regards to taking in room details. It seems that the request isn't actually using any of the room details, which calls into question the relevance of `roomId` and why this request is in the `RoomController.java` class and whatnot.

Also, this is unrelated to this PR, but I just realized that I'm not sure that we handle the case where a user is already in a room and tries to join another. Details like this will become especially relevant when we add accounts, and users could log in and thus realistically attempt to join two different rooms at the same time.

@alankbi, what are your thoughts on both of the above?

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -298,4 +299,25 @@ public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request)
...
+
+        // Return error if the initiator is not the host or the same as the receiver
+        User initiator = UserMapper.toEntity(request.getInitiator());
+        User receiver = UserMapper.toEntity(request.getReceiver());
+        if (!room.getHost().equals(initiator) && !initiator.equals(receiver)) {
+            throw new ApiException(RoomError.INVALID_PERMISSIONS);
+        }
+
+        User modifiedUser = room.getUserByUserId(receiver.getUserId());
+        modifiedUser.setIsSpectator(isSpectator);
```
zpChris, line 319: You need to handle null values here:
1. The values `initiator` and `receiver` have to be verified as non-null.
2. The `modifiedUser` below can be null if the `receiver.getUserId()` is invalid, meaning that you also have to confirm that the Users provided actually exist in the database by checking against `findUserById` and verifying that you get a non-null result.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -298,4 +299,25 @@ public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request)
...
+
+        // Return error if room could not be found
+        if (room == null) {
+            throw new ApiException(RoomError.NOT_FOUND);
+        }
+
+        // Return error if the initiator is not the host or the same as the receiver
+        User initiator = UserMapper.toEntity(request.getInitiator());
+        User receiver = UserMapper.toEntity(request.getReceiver());
+        if (!room.getHost().equals(initiator) && !initiator.equals(receiver)) {
```
zpChris, line 314: The `initiator` and `receiver` could potentially both be bogus objects and still pass this, as there is a simple equality check; you have to check that they are real users in the database (or, if we stick with this room incorporation in this PR, check that both users exist in the room).

zpChris (changes requested):
Great first work! I have some questions about how the PR is structured, primarily with respect to its incorporation of rooms. I know we had talked about this before and I had a-ok'd the broad design, so sorry about that mistake - I'll talk to @alankbi to see how we can best structure this.

src/main/java/com/rocketden/main/controller/v1/RoomController.java
zpChris: Ah, I realized that you **are** using the room - you're allowing the host to set the user as a spectator. I think that actually makes sense, I'm not certain what this use case would be but perhaps that extra room management could be useful.

You can ignore the concerns about the room details! I think this structure makes sense 👍

zpChris:
> Great first work! I have some questions about how the PR is structured, primarily with respect to its incorporation of rooms. I know we had talked about this before and I had a-ok'd the broad design, so sorry about that mistake - I'll talk to @alankbi to see how we can best structure this.

Rethinking this, I don't think this PR needs to be re-structured, since you allow the host to set another user's spectator value. Sorry about this false alarm 👍

src/main/java/com/rocketden/main/controller/v1/RoomController.java
```diff
@@ -67,4 +69,9 @@ public RoomController(RoomService service) {
                                                       @RequestBody UpdateSettingsRequest request) {
         return new ResponseEntity<>(service.updateRoomSettings(roomId, request), HttpStatus.OK);
     }
+
+    @PostMapping("/rooms/{roomId}/spectator/{isSpectator}")
```
alankbi, line 73: nit: maybe change this to `/rooms/{roomId}/spectator` and include the spectator variable in the request body (SetSpectatorRequest), just to be more consistent with [REST naming conventions](https://restfulapi.net/resource-naming/)

src/main/java/com/rocketden/main/dto/room/SetSpectatorRequest.java
```diff
@@ -0,0 +1,12 @@
+package com.rocketden.main.dto.room;
+
+import com.rocketden.main.dto.user.UserDto;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class SetSpectatorRequest {
```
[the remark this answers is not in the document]
alankbi: Agreed - actually just left another comment mentioning the same thing before I saw this haha

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -298,4 +299,25 @@ public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request)
...
+            throw new ApiException(RoomError.INVALID_PERMISSIONS);
+        }
+
+        User modifiedUser = room.getUserByUserId(receiver.getUserId());
+        modifiedUser.setIsSpectator(isSpectator);
+
+        return RoomMapper.toDto(room);
```
alankbi, line 321: I think you need a call to `repository.save(room)` of sorts, otherwise this might not actually change the object in the database

src/test/java/com/rocketden/main/service/RoomServiceTests.java
```diff
@@ -814,4 +816,73 @@ public void deleteRoomBadHost() {
...
+
+        // Mock repository to return room when called
+        Mockito.doReturn(room).when(repository).findRoomByRoomId(eq(ROOM_ID));
+
+        SetSpectatorRequest request1 = new SetSpectatorRequest();
+        request1.setInitiator(UserMapper.toDto(firstUser));
+        request1.setReceiver(UserMapper.toDto(secondUser));
+        RoomDto response1 = roomService.setSpectator(ROOM_ID, true, request1);
+        assertTrue(response1.getUsers().get(1).getIsSpectator());
+
```
alankbi, line 849: nit: maybe here add an additional `verify(repository).save(room)` check

alankbi (commented):
Just left a couple comments in addition to what @zpChris wrote! One more thing is I'd suggest adding one more setSpectatorSuccess test in the `RoomTests.java` file - these API tests are slightly more robust than the unit tests and it'd be good to have at least one

alankbi (approved):
Approved! A couple nits but overall nothing major

zpChris:
Sorry I didn't realize this was waiting for my review! I'll check it out tonight.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -298,4 +299,36 @@ public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request)
...
+        if (room == null) {
+            throw new ApiException(RoomError.NOT_FOUND);
+        }
+
+        //Return error if the initiator or receiver are null
+        if (request.getInitiator() == null || request.getReceiver() == null) {
+            throw new ApiException(UserError.NOT_FOUND);
+        }
+
+        //Return error if the initiator or receiver are not in the room
```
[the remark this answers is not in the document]
zpChris: Seconded - the other one is right above on line 311.

src/main/java/com/rocketden/main/service/RoomService.java
```diff
@@ -298,4 +299,36 @@ public RoomDto updateRoomSettings(String roomId, UpdateSettingsRequest request)
...
+        // Return error if the initiator is not the host or the same as the receiver
+        User initiator = UserMapper.toEntity(request.getInitiator());
+        User receiver = UserMapper.toEntity(request.getReceiver());
+        if (!room.getHost().equals(initiator) && !initiator.equals(receiver)) {
+            throw new ApiException(RoomError.INVALID_PERMISSIONS);
+        }
+
+        User modifiedUser = room.getUserByUserId(receiver.getUserId());
+        modifiedUser.setSpectator(request.getSpectator());
+        repository.save(room);
```
zpChris, line 330: Check that `request.getSpectator() == null` above this code (and throw `BAD_SETTING` if so); otherwise, the spectator in the backend can equal `null` on a bad request, which we want to avoid.

zpChris (approved):
LGTM! ✅ 

I left a note that is quick to resolve but important, so be sure to address that before merging.

zpChris:
[Add spectator boolean to backend](https://trello.com/c/3ANkP4Sj/328-add-spectator-boolean-to-backend)

PR #138, opened 2021-04-06
Add numProblems slider

jasonz6688:
### List of Changes:

- Added slider to the lobby page
- Added numProblems value to the room, lobby, and updateSettingsRequest
- Added two tests, one that tests the correct number of problems is returned, one that tests that an error is thrown when numProblems is too large or too small

### Notes:

The slider is set between 1 and 10, as that's what the pre-written code in the backend accepted. I kept all the backend code.

### Screencast and Images:

Cap63
Cap64


frontend/src/api/Game.ts
```diff
@@ -115,4 +115,4 @@ export const playAgain = (roomId: string, params: PlayAgainParams):
   .then((res) => res.data)
   .catch((err) => {
     throw axiosErrorHandler(err);
-  });
+  });
```
alankbi, line 118: nit: re-add new line

frontend/src/views/Lobby.tsx
```diff
@@ -143,6 +143,7 @@ function LobbyPage() {
   const [difficulty, setDifficulty] = useState<Difficulty | null>(null);
   const [duration, setDuration] = useState<number | undefined>(15);
   const [size, setSize] = useState<number | undefined>(10);
+  const [numProblems, setNumProblems] = useState<number | undefined>(1);
```
alankbi, line 146: nit: this can be simply of type `<number>` (or no type at all, since it can infer that it's a number from the default value)

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
```diff
@@ -174,6 +174,39 @@ public void startGameSuccess() {
...
+        assertTrue(response.isActive());
+
+        // Game object is created when the room chooses to start
+        Game game = gameService.getGameFromRoomId(ROOM_ID);
+        assertNotNull(game);
+
+        assertNotNull(game.getGameTimer());
+        assertEquals(room.getDuration(), game.getGameTimer().getDuration());
+        assertEquals(room.getNumProblems(), response.getNumProblems());
+    }
```
alankbi, line 208: Could you check if there are any existing tests that cover this? My intuition would be that we've already written tests to cover setting numProblems and invalid numProblem errors - however it's been a while so I could be wrong on this, in which case thanks for adding these two tests :)

alankbi (approved):
Looks good, approved! However, I left a few nits which you should take a look at before merging (also would recommend waiting till @zpChris reviews as well in case he has any other comments)

zpChris (approved):
LGTM! ✅ Just left two small notes, definitely respond to @alankbi's comments before merging 👍

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
jasonz6688: So the most similar tests I found were a couple of get problem tests in ProblemServiceTests, which tested whether multiple problems are sent from the backend. Also I could theoretically just modify the startGameSuccess test to include multiple questions. What do you think I should do about this?

frontend/src/views/Lobby.tsx
```diff
@@ -711,6 +732,47 @@ function LobbyPage() {
...
+                  min={1}
+                  max={10}
+                  value={numProblems}
+                  disabled={!isHost(currentUser)}
+                  onChange={(e) => {
+                    const { value } = e.target;
+
+                    // Set numProblems to undefined to allow users to clear field
+                    if (!value) {
+                      setNumProblems(undefined);
```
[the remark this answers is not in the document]
jasonz6688: Good point!

frontend/src/views/Lobby.tsx
```diff
@@ -711,6 +732,47 @@ function LobbyPage() {
...
+                    } else {
+                      const newNumProblems = Number(value);
+                      if (newNumProblems >= 1 && newNumProblems <= 10) {
+                        setNumProblems(newNumProblems);
+                      }
```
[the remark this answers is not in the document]
jasonz6688: I added this because all our other sliders seem to have another check, despite the number still being restricted by the slider and unallowable answers caught by the backend. I could delete this if statement, but I don't think there is much harm in leaving it in here.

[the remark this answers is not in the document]
zpChris: Got it - agreed, that makes sense, leave it in! 👍

src/test/java/com/rocketden/main/service/GameManagementServiceTests.java
alankbi: Nah that's fine, your explanation makes sense!

PR #140, opened 2021-04-15
Change the front end to receive and use multiple problems

jasonz6688:
### List of Changes:

- CurrentCode and CurrentLanguage made into arrays
- Added int currentProblem which designates which problem is current worked on
- Replaced all problem[0] with problem[currentProblem], currentCode with currentCode[currentProblem], currentLanguage with currentLanguage[currentProblem]
- Deleted a dependency on callback getDefaultCode

### Notes:

This PR is to test whether GitHub will accept the callback without the function as a dependency.


frontend/src/views/Game.tsx
```diff
@@ -158,16 +167,16 @@ function GamePage() {
...
       if (code) {
-        codeMap[language] = code;
+        codeMap[language[currentProblem]] = code;
       }
 
       // Set this user's current code and language
-      setCurrentCode(codeMap[language]);
-      setCurrentLanguage(language);
+      setCurrentCode(codeMap[language[currentProblem]]);
+      setCurrentLanguage(language[currentProblem]);
```
alankbi, line 179: nit: add currentProblem to the dependency list for this callback

frontend/src/views/Game.tsx
```diff
@@ -269,7 +278,7 @@ function GamePage() {
         error: errorHandler('No valid room details were provided, so you could not view the game page.'),
       });
     }
-  }, [location, history, setDefaultCodeFromProblems]);
+  }, [location, history]);
```
alankbi, line 281: yeah i do think this will probably need to be added back

frontend/src/views/Game.tsx
```diff
@@ -147,8 +148,16 @@ function GamePage() {
     setTimeUp(newGame.gameTimer.timeUp);
   };
 
+  const setCurrentLanguage = (newLanguage: Language) => {
+    currentLanguage[currentProblem] = newLanguage;
+  };
+
+  const setCurrentCode = (newCode: string) => {
+    currentCode[currentProblem] = newCode;
+  };
```
alankbi, line 157: With react state you're [required](https://stackoverflow.com/questions/37755997/why-cant-i-directly-modify-a-components-state-really) to use the provided `setX` methods that they provide, any state changes have to come through those methods (so you'll have to revert back to `const [currentLanguage, setCurrentLanguages] = useState(...)`. So in this situation, I'd recommend taking a look at `ProblemDisplay.tsx` starting at about line 220 to get an idea for how to modify lists and try to mimic that here

Idk if this is related to the issue you're currently having but hopefully changing this could potentially help?

PR #141, opened 2021-04-15
Allow users to select specific problems on the frontend

alankbi:
### List of Changes:

* Create problem selector component for choosing problems on the lobby page
* Display selected problems using a new SelectedProblemsDisplay component
* Add ability to perform a simple search/filter over the list of problems
* Update get problems endpoint to provide option for finding only approved problems
* Make minor code refactorings in various files

### Notes:

* Right now I'm not sure how intuitive it is that if you don't choose a problem, it'll be filled by the difficulty setting or vice vera. Is there any design or explanation field that we should add to clear this up? 

### Screencast and Images:

[Screencast of new behavior](https://www.loom.com/share/a559a39d87aa4cdfb12f35b451d14165) (forgot to show search functionality)

Lobby page showing ability to select problems


alankbi:
[Create UI for selecting problems on the frontend](https://trello.com/c/5oMACHhW/326-create-ui-for-selecting-problems-on-the-frontend)

zpChris:
Reviewing now!

zpChris:
> Right now I'm not sure how intuitive it is that if you don't choose a problem, it'll be filled by the difficulty setting or vice vera. Is there any design or explanation field that we should add to clear this up?

I think this PR could pass without adding something to address this concern, but one potential solution would be to add a little (?) icon next to "Selected Problems" which uses the HoverElement attribute already on the page, but with custom text. If you hovered over the (?), you could say something like: "The problems not chosen through 'Selected Problems' will be fetched randomly from the database using the difficulty requirement set above." That's just one possibility though, I'm sure there are other solutions / explanations as well.

zpChris:
The test `GameTests.startGameWithProblemIdGetsCorrectProblem` is failing right now - you likely want to investigate that @alankbi, I'm guessing it may be the cause of some unnoticed merge conflict.

frontend/src/components/problem/ProblemSelector.tsx
```diff
@@ -0,0 +1,144 @@
...
+          if (selectedProblems.some((p) => p.problemId === problem.problemId)) {
+            return null;
+          }
+
+          if (searchText && !problem.name.toLowerCase().includes(searchText.toLowerCase())) {
+            return null;
+          }
+
+          return (
+            <InlineProblem
```
zpChris, line 120: In the future, we could add a little "See More" icon that opens up the problem in a modal, so the host could decide if they want to include it! (No changes to this PR, just an idea for the future.)

frontend/src/components/special/IdContainer.tsx
```diff
@@ -24,9 +24,11 @@ function IdContainer(props: IdProps) {
       {
         id.split('').map((c, index) => {
           if (id.length - 1 === index) {
-            return <NumberContainer style={{ margin: 0 }}>{c}</NumberContainer>;
+            // eslint-disable-next-line react/no-array-index-key
+            return <NumberContainer key={index} style={{ margin: 0 }}>{c}</NumberContainer>;
           }
-          return <NumberContainer>{c}</NumberContainer>;
+          // eslint-disable-next-line react/no-array-index-key
+          return <NumberContainer key={index}>{c}</NumberContainer>;
```
zpChris, line 31: I think you added the key to remove the warning, but you left the disable warning comment here - did you mean to remove that?

frontend/src/views/Lobby.tsx
```diff
@@ -525,6 +565,20 @@ function LobbyPage() {
     }
   }, [history, active, currentUser, currentRoomId, difficulty, subscription]);
 
+  const hoverProps = {
```
zpChris, line 568: Thanks for adding this! 👍

frontend/src/views/Lobby.tsx
```diff
@@ -333,6 +338,40 @@ function LobbyPage() {
...
+
+    const prevProblems = selectedProblems;
+    setSelectedProblems(newProblems);
+
+    const settings = {
+      initiator: currentUser!,
+      problems: newProblems,
+    };
+
+    updateRoomSettings(currentRoomId, settings)
```
zpChris, line 356: I think I'm missing something here; the object you pass in has `problems` which is a `SelectableProblem[]`, but the `settings` object for `updateRoomSettings` has no such field - how does this work?

src/main/java/com/rocketden/main/model/Room.java
```diff
@@ -66,6 +68,7 @@
 
     @ManyToMany(fetch = FetchType.EAGER)
     @JoinColumn(name = "selectable_problem_id")
+    @Fetch(FetchMode.SUBSELECT)
```
zpChris, line 71: What does this do?

zpChris (approved):
LGTM! ✅ 

I did leave a good number of comments, but they are all nits or potential new features, so I think whatever you decide to do for all of them, this looks good to merge! This is awesome, good job on this PR.

alankbi:
> The test `GameTests.startGameWithProblemIdGetsCorrectProblem` is failing right now - you likely want to investigate that @alankbi, I'm guessing it may be the cause of some unnoticed merge conflict.

Good catch - I had fixed a bug on my latest merged PR but hadn't merged from main yet

frontend/src/api/Problem.ts
```diff
@@ -77,7 +84,7 @@ export type DefaultCodeType = {
 
 const basePath = '/api/v1/problems';
 const routes = {
-  getProblems: `${basePath}/`,
+  getProblems: `${basePath}`,
   createProblem: `${basePath}/`,
```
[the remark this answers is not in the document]
alankbi: good catch - this is a leftover change from when I was trying to debug an error, so I'll change this back

frontend/src/api/Room.ts
```diff
@@ -27,6 +30,7 @@ export type UpdateSettingsParams = {
   initiator: User,
   difficulty?: Difficulty,
   duration?: number,
+  problem?: SelectableProblem,
```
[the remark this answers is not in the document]
alankbi: I originally implemented this with one problem in mind so I guess I just forgot to update this haha ¯\_(ツ)_/¯

frontend/src/components/problem/ProblemSelector.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const [error, setError] = useState('');
+  const [problems, setProblems] = useState<SelectableProblem[]>([]);
+  const [showProblems, setShowProblems] = useState(false);
+  const [searchText, setSearchText] = useState('');
+
+  const ref = useRef<HTMLDivElement>(null);
+
+  // Close list of problems if clicked outside of div
+  useEffect(() => {
```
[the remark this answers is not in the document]
alankbi: yeah that's a good idea! we can also do that for the modal too probably

frontend/src/components/problem/ProblemSelector.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const handleClickOutside = (e: MouseEvent) => {
+      if (ref.current && !ref.current!.contains(e.target as Node)) {
+        setShowProblems(false);
+      }
+    };
+
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => document.removeEventListener('mousedown', handleClickOutside);
+  }, [ref]);
```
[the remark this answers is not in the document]
alankbi: that's actually such a good point lol - essentially useRef here gives you access to the specified DOM node (at the `ref={ref}`via `ref.current`. then, it adds an event listener for mousedown that checks if it clicked inside of the chosen node (if not, then it will close the window)

frontend/src/components/problem/ProblemSelector.tsx
```diff
@@ -0,0 +1,144 @@
...
+      />
+
+      <InnerContent show={showProblems} ref={ref}>
+        {problems.map((problem, index) => {
+          // Only show problems that haven't been selected yet
+          if (selectedProblems.some((p) => p.problemId === problem.problemId)) {
+            return null;
+          }
+
+          if (searchText && !problem.name.toLowerCase().includes(searchText.toLowerCase())) {
```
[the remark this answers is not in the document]
alankbi: hopefully it's still efficient enough when we have 100+ problems :)

frontend/src/components/problem/ProblemSelector.tsx
```diff
@@ -0,0 +1,144 @@
...
+            return null;
+          }
+
+          return (
+            <InlineProblem
+              key={problem.name}
```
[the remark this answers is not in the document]
alankbi: Changed this to use `problem.problemId` instead which should be pretty much guaranteed to be unique (UUID)

frontend/src/components/special/IdContainer.tsx
alankbi: The key removed the errors in the console, but ESLint still gives a warning for using the index as the key (it doesn't want you to since if the items in a list reorder, then using indices can be bad for performance). However I couldn't find any other option (the character `c` can be duplicate) and also this is splitting up a static string so it will never change order, so I figured it'd be ok - but lmk your thoughts

frontend/src/views/Lobby.tsx
alankbi: haha yeah i forgot to update the typescript type (mentioned in another comment). as for why it works, i think typescript allows you to have additional properties (which is why it allows `getAllProblems => Problem[]` to convert to `SelectableProblem[]`), so it just ignores the optional `problems` property while this correct one is sent anyways

src/main/java/com/rocketden/main/model/Room.java
alankbi: ManyToMany eagerly fetched apparently by default does a [cartesian product](https://stackoverflow.com/questions/22224931/getting-duplicate-records-when-using-manytomany-annotation) causing duplicates, so this was necessary to fix that

frontend/src/components/problem/ProblemSelector.tsx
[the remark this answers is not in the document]
zpChris: Do you think we should revise the current "Hover Element" setup we have with **ref** then when we revise the frontend?

frontend/src/components/special/IdContainer.tsx
zpChris: Ok gotcha - yeah this is totally fine by me, thanks for the explanation 👍

frontend/src/views/Lobby.tsx
zpChris: Got it, thanks for updating this 👍

src/main/java/com/rocketden/main/model/Room.java
zpChris: Got it, sounds good - I will note that the [SubSelect Mode](https://stackoverflow.com/questions/32984799/fetchmode-join-vs-subselect) could mean performance issues later in development, so we end up restructuring this in the future, but I think this works well for now.

PR #146, opened 2021-04-21
Change frontend to take multiple problems

jasonz6688:
### List of Changes:

- CurrentCode and CurrentLanguage made into arrays
- Added int currentProblem which designates which problem is current worked on
- Replaced all problem[0] with problem[currentProblem], currentCode with currentCode[currentProblem], currentLanguage with currentLanguage[currentProblem]

### Notes:

Ignore the buttons that switch current problem and any other code associated with this functionality.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -2,9 +2,7 @@ name: Java CI/CD
 
 on:
   push:
-    branches:
-      [ main ]
-      [ changeFrontEndtoTakeMultipleProblems ]
+    branches: [ main, changeFrontEndtoTakeMultipleProblems ]
   pull_request:
     branches: [ main ]
 
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -2,7 +2,9 @@ name: Java CI/CD
 
 on:
   push:
-    branches: [ main ], [ changeFrontEndtoTakeMultipleProblems ]
+    branches:
+      [ main ]
+      [ changeFrontEndtoTakeMultipleProblems ]
   pull_request:
     branches: [ main ]
 
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -2,7 +2,7 @@ name: Java CI/CD
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main ], [ changeFrontEndtoTakeMultipleProblems ]
   pull_request:
     branches: [ main ]
 
--- a/frontend/src/api/Problem.ts
+++ b/frontend/src/api/Problem.ts
@@ -170,5 +170,6 @@ export const getDefaultCodeMap = (problemId: string): Promise<DefaultCodeType> =
   .get<DefaultCodeType>(routes.defaultCodeMap(problemId))
   .then((res) => res.data)
   .catch((err) => {
+    console.log("this is where it's all coming from");
     throw axiosErrorHandler(err);
   });
--- a/frontend/src/components/game/Editor.tsx
+++ b/frontend/src/components/game/Editor.tsx
@@ -8,9 +8,10 @@ import { DefaultCodeType } from '../../api/Problem';
 type EditorProps = {
   onLanguageChange: ((language: Language) => void) | null,
   onCodeChange: ((code: string) => void) | null,
-  codeMap: DefaultCodeType | null,
+  codeMap: DefaultCodeType[] | null,
   defaultLanguage: Language,
   defaultCode: string | null,
+  currentProblem: number
 };
 
 const Content = styled.div`
@@ -88,7 +89,7 @@ const monacoEditorOptions: EditorConstructionOptions = {
 // This function refreshes the width of Monaco editor upon change in container size
 function ResizableMonacoEditor(props: EditorProps) {
   const {
-    onLanguageChange, onCodeChange, codeMap, defaultLanguage, defaultCode,
+    onLanguageChange, onCodeChange, codeMap, defaultLanguage, defaultCode, currentProblem,
   } = props;
 
   const theme = useContext(ThemeContext);
@@ -124,8 +125,8 @@ function ResizableMonacoEditor(props: EditorProps) {
   const handleLanguageChange = (language: Language) => {
     // Save the code for this language
     if (codeMap != null && codeEditor != null) {
-      codeMap[currentLanguage] = codeEditor.getValue();
-      codeEditor.setValue(codeMap[language]);
+      codeMap[currentProblem][currentLanguage] = codeEditor.getValue();
+      codeEditor.setValue(codeMap[currentProblem][language]);
     }
 
     // Change the language and initial code for the editor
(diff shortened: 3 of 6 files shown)


jasonz6688:
[Multiple problems for a single room](https://trello.com/c/Ik0cqGYE/327-multiple-problems-for-a-single-room)

jasonz6688:
[Multiple problems for a single room](https://trello.com/c/Ik0cqGYE/327-multiple-problems-for-a-single-room)

frontend/src/views/Game.tsx
```diff
@@ -147,8 +148,28 @@ function GamePage() {
...
+    while (currentLanguage.length < problems.length) {
+      currentLanguage.push(Language.Python);
+    }
+
+    while (currentCode.length < problems.length) {
+      currentCode.push('');
+    }
+  };
+
+  createCodeLanguageArray();
```
zpChris, line 161: If I understood correctly, I'm ignoring this - I do have comments so let me know when / if you want me to add those (could be here or on a new PR).

frontend/src/api/Problem.ts
```diff
@@ -170,5 +170,6 @@ export const getDefaultCodeMap = (problemId: string): Promise<DefaultCodeType> =
   .get<DefaultCodeType>(routes.defaultCodeMap(problemId))
   .then((res) => res.data)
   .catch((err) => {
+    console.log("this is where it's all coming from");
```
zpChris, line 173: Are you going to keep this in on the merge? Do you want me to comment on this?

frontend/src/views/Game.tsx
```diff
@@ -355,6 +376,20 @@ function GamePage() {
...
+
+  const previousProblem = () => {
+    let temp = currentProblem.valueOf() - 1;
+
+    if (temp < 0) {
+      temp += problems?.length;
+    }
+
+    setCurrentProblem(temp);
+  };
```
zpChris, line 391: I know you didn't want me to comment on this, but I'll say one thing - I would recommend that you don't have it wrap around to problem 1 after clicking "Next" on problem 4 (out of 4 problems). This could make sense if you have the annotation 1/4, 2/4, etc., but if we don't then this functionality shouldn't exist.

zpChris (commented):
I'm not entirely certain what you want me to comment on - besides the "Previous / Next" buttons and the default code map, everything else seems fine. Do you want me to comment on perhaps how to fix the default code map?

.github/workflows/workflow.yml
```diff
@@ -2,7 +2,7 @@ name: Java CI/CD
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, changeFrontEndtoTakeMultipleProblems ]
```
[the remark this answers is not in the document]
jasonz6688: Yeah, sorry, you're totally right about that. It was mighty small brain of me to use the specific branch instead of the staging one.

frontend/src/views/Game.tsx
jasonz6688: Umm, yeah, this code is necessary to switch between problems, so we can talk about this in pair programming later today.

frontend/src/api/Problem.ts
jasonz6688: No, sorry, this is one of the console.log statements I didn't see and delete before pushing this branch.

.github/workflows/workflow.yml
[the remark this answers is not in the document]
zpChris: No problem, you're all good 👍

frontend/src/components/game/Editor.tsx
```diff
@@ -136,10 +137,16 @@ function ResizableMonacoEditor(props: EditorProps) {
   };
 
   useEffect(() => {
+    console.log(codeMap);
+    console.log(currentProblem);
+    console.log(currentLanguage);
```
zpChris, line 142: You'll likely want to remove these before merging?

zpChris (approved):
LGTM! ✅ 

I'll leave two notes here:
- As I mentioned on the call, there may be things to improve on that is separate from what you indicated for me to review - I will likely note those down on future PRs.
- I'm not sure if you're planning on doing this, but as far as I can tell right now, this PR does not update the submission grading system to handle multiple problems - this will need to be done at some point on the `stagingNumProblems` branch, if not already planned.

frontend/src/views/Game.tsx
alankbi, line 161: I'm not sure if this actually works - if I change the language to Java for any problem, then the rest will also be Java. I think instead you could put this into a useEffect hook (with a comment explaining that this initializes these values) for whenever problems changes (that way this also isn't repeatedly called on every re-render

frontend/src/views/Game.tsx
```diff
@@ -147,8 +148,28 @@ function GamePage() {
...
+
+    while (currentCode.length < problems.length) {
+      currentCode.push('');
+    }
+  };
+
+  createCodeLanguageArray();
+
+  const setOneCurrentLanguage = (newLanguage: Language) => {
+    currentLanguage[currentProblem.valueOf()] = newLanguage;
```
alankbi, line 164: is there a need to use valueOf here and below? i would have thought it would be automatic (maybe lowercase `number` instead of `Number` could help? idk)

frontend/src/views/Game.tsx
```diff
@@ -158,22 +179,26 @@ function GamePage() {
...
+      const codeList = [];
+
+      for (let i = 0; i < result.length; i += 1) {
+        codeList.push(result[i][language[i]]);
+      }
 
       // If previous code and language specified, save those as defaults
       if (code) {
-        codeMap[language] = code;
+        codeList[currentProblem.valueOf()] = code;
```
alankbi, line 190: so this only works with one set of code for now right? totally fine if so, just wanted to double check though

frontend/src/views/Game.tsx
```diff
@@ -158,22 +179,26 @@ function GamePage() {
 
     // Get the result of promises and set the default code list.
     Promise.all(promises).then((result) => {
-      const codeMap = result[0];
+      const codeList = [];
+
+      for (let i = 0; i < result.length; i += 1) {
+        codeList.push(result[i][language[i]]);
```
alankbi, line 185: how does this work? it indexes language[i] but it seems like language is always a list of length 1 with either the default Python or whatever language the last submission was in?

alankbi (approved):
Looks good! I left a couple of nits in addition to Chris's and I think there def other changes, but those will likely come in the later PRs so that should be fine.

frontend/src/views/Game.tsx
jasonz6688: So the codeLangeArray is just the initial array. Afterwards, the language select will only change one of the elements in the array. This can be checked in 147. Although, I'm not sure what to do about the re-render problem. I'm afraid that if I put it in a useEffect hook that it'll overwrite itself accidentally, as the hook would be called every time the languageArray or codeArray is changed.

frontend/src/views/Game.tsx
jasonz6688: Yeah ok, I'll change that in my next PR.

frontend/src/views/Game.tsx
jasonz6688: Yep, it's a much bigger change to keep track of all code that was previously submitted because we'd half to keep track of a lot more on the back end per user.

frontend/src/views/Game.tsx
zpChris: I think this was updated on the latest PR #152, so I'll resolve this comment.

jasonz6688:
[Changing frontend to take multiple problems](https://trello.com/c/JIb9zWWP/389-changing-frontend-to-take-multiple-problems)

PR #147, opened 2021-04-29
Adding the buttons to Switch Problem, Fixed to 

jasonz6688:
### List of Changes:

- The Game now has the necessary buttons and functions to change numProblems
- The editor takes a few more and different props, the defaultCodeMap is now a 2D array, and there is a function prop that tells the editor which language is mapped to the currentProblem
- The console, test cases, and description all change based on the numProblems
- There is another field in submission request which tells the backend which problem is being worked on. The backend uses this field as the index of problems to find the problem. The default of this field is 0, so that all previous tests still work.

### Notes:

The style of the buttons is a little bit messed up. They're not quite lined up and one is on top of the other, so someone will need to redo these in a later PR. Also, the 10 minute left notification covers the buttons.

Additionally, the game seems to end as long as everyone answers one question completely correct. I'll fix this bug in the next PR. Also, the submission is erased after numProblems change because editor takes only one submission and will only display that in the box if set. If desired, we can store a list of submissions.

My way of initializing the code and language arrays in Game.tsx is inelegant, and I'm open to suggestions about that.

I have not written a test for the added field in the submissionRequest. I could add one, but the change seemed so insignificant, and the existing tests seem quite complicated to change to test whether the added field works. That being said, it is abundantly clear from mock games that submissions are working for the 2nd, 3rd, 4th, etc. problem.

### Screencast and Images:

Cap55
Cap56


alankbi:
Does this build on top of #146? If so, I'd change this to merge into that branch instead, that way it's easier to tell which changes come from that PR and which from this PR

frontend/src/views/Results.tsx
```diff
@@ -273,7 +273,9 @@ function GameResultsPage() {
           <ResizableMonacoEditor
             onLanguageChange={null}
             onCodeChange={null}
-            codeMap={null}
+            getCurrentLanguage={null}
+            defaultCodeMap={null}
+            currentProblem={0}
```
alankbi, line 278: so for now the results page only shows the code for the first problem? if you changed this to be 1 would that cause it to show the code for the second problem (or n/a if it doesn't exist)?

src/main/java/com/codejoust/main/dto/game/SubmissionRequest.java
```diff
@@ -13,4 +13,5 @@
     private String code;
     private String input;
     private UserDto initiator;
+    private int problem = 0;
```
alankbi, line 16: nit: rename this to index or problemIndex or something (since problem is usually used with a Problem or ProblemDto class

frontend/src/views/Game.tsx
```diff
@@ -251,14 +276,14 @@ function GamePage() {
           // If this user refreshed and has already submitted code, load and save their latest code
           res.players.forEach((player) => {
             if (player.user.userId === location.state.currentUser.userId && player.code) {
-              setDefaultCodeFromProblems(res.problems, player.code, player.language as Language);
+              setDefaultCodeFromProblems(res.problems, player.code, [player.language as Language]);
```
alankbi, line 279: just a heads up, i don't think the player code is loaded after a refresh (this might be a later issue tho, especially since we'll have to figure out saving multiple submissions/code for different problems and everything)

alankbi:
@jasonz6688 in your screenshots I noticed a couple of things: 1. there are those scrollbars everywhere showing by default, 2. the button text isn't perfectly centered, and 3. the font for the editor is wrong (see below for how it looks on Mac)

Game page on Mac

do you know if this occurs on codejoust.co as well? If so then we might need to do some styling changes for PC lol

frontend/src/views/Results.tsx
jasonz6688: Actually, the defaultCode will be displayed regardless of what currentProblem is set to, as the editor defaults to the defaultCode and only uses the codeMap and currentProblem after one of the callbacks/effects are triggered, which shouldn't happen on the Results page. I just put a 0 there because ints can't be null. Also to answer your question in the main comment, I just merged 146 into stagingNumProblems.

frontend/src/views/Game.tsx
jasonz6688: Yeah, I was quite certain that these lines were going to cause a problem. The alternative is saving another code array and language array in each player object, but I think there would be a problem with generating an incomplete array from the backend. Design-wise maybe we should just not remember submitted code after a refresh.

jasonz6688:
I don't really have an answer to the style questions. I don't believe I messed with the style of anything, especially not font. The new scroll bar might be because some of the components have a fixed height, and the two new buttons messed some stuff up. Additionally, I think these style problems can be resolved in a final PR on this story.

src/main/java/com/codejoust/main/dto/game/SubmissionRequest.java
alankbi: be sure to push the change if you haven't already

frontend/src/views/Game.tsx
alankbi: yeah just thinking about it it seems like it's going to get too messy. I would prefer to remember submitted code, but at the same time there's a chance in my Redux PR that after refreshing you're booted from the room so who knows lol

However I would say that code should be remembered for each problem and each language when not refreshing (so if you just switch randomly between different problems and languages it won't lose your code)

alankbi (approved):
Approved! With the caveat that many of the current issues will be addressed in the next PR. 

Also, before merging, @zpChris and I left a few comments on #146 that I don't think were addressed, so be sure to fix those here before merging if you can. 

I will note I am getting slightly worried that this multi-problem feature is getting a bit complicated and potentially introducing cascading issues that require a lot of nested or multi-dimensioned lists on both frontend and backend; just something to note cause I can see this feature becoming more difficult than imagined.

jasonz6688:
[Switching between Problems PR](https://trello.com/c/uXiHLWDE/388-switching-between-problems-pr)

PR #148, opened 2021-04-30
Update default code setup.

zpChris:
### List of Changes:
- Change the default code to `Java`.
- Update Python code to use an instance method and include type hinting with `typeInstantiationToString`.
- Update relevant tests for Python default code.

### Notes:
- This PR requires the merge of [tester #22](https://github.com/CodeJoustHQ/tester/pull/22) to function properly, as described in [this discussion](https://github.com/CodeJoustHQ/main/pull/148#discussion_r624234885).
  - If you want to run this locally, you have to point the run code call to a local instance of tester that has the changes in the PR referenced above.
- An extra line was included for the Python default code to prevent the language selector on the frontend from overlapping with an extra-long method line.

### Screencast and Images:

[Screencast of New Python Default Code](https://www.loom.com/share/ab67abedd46547d4a7fd581f72a59b40)

**Python Default Code Type Hinting**
Python Default Code Type Hinting

### DIFF
--- a/src/main/java/com/codejoust/main/service/generators/PythonDefaultCodeGeneratorService.java
+++ b/src/main/java/com/codejoust/main/service/generators/PythonDefaultCodeGeneratorService.java
@@ -61,15 +61,15 @@ public class PythonDefaultCodeGeneratorService implements DefaultCodeGeneratorSe
             case BOOLEAN:
                 return "bool";
             case ARRAY_STRING:
-                return "List[str]";
+                return "list[str]";
             case ARRAY_INTEGER:
-                return "List[int]";
+                return "list[int]";
             case ARRAY_DOUBLE:
-                return "List[float]";
+                return "list[float]";
             case ARRAY_CHARACTER:
-                return "List[str]";
+                return "list[str]";
             case ARRAY_BOOLEAN:
-                return "List[bool]";
+                return "list[bool]";
             default:
                 throw new ApiException(ProblemError.BAD_IOTYPE);
         }
--- a/src/main/java/com/codejoust/main/service/generators/PythonDefaultCodeGeneratorService.java
+++ b/src/main/java/com/codejoust/main/service/generators/PythonDefaultCodeGeneratorService.java
@@ -25,11 +25,13 @@ public class PythonDefaultCodeGeneratorService implements DefaultCodeGeneratorSe
             methodLineBuilder.append(
                 String.format(", %s: %s", 
                     problemInput.getName(),
-                    typeInstantiationToString(outputType)
+                    typeInstantiationToString(problemInput.getType())
                 )
             );
         }
-        methodLineBuilder.append("):");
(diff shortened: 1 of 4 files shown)


zpChris:
[Update the default code setup](https://trello.com/c/Pq5GSwaj/355-update-the-default-code-setup)

src/main/java/com/codejoust/main/service/generators/PythonDefaultCodeGeneratorService.java
```diff
@@ -27,8 +27,12 @@ public String getDefaultCode(List<ProblemInput> problemInputs, ProblemIOType out
         }
         methodLineBuilder.append("):");
 
+        // Add a method header comment to describe the return type.
+        String returnComment = String.format("\t# The solve method should return the type %s", outputType.getClassType().getSimpleName());
+
```
alankbi, line 32: Instead a comment actually perhaps we could use Python's built-in [type hinting](https://docs.python.org/3/library/typing.html)? Didn't realize this was a thing till now but also saw that this is what LeetCode does if you select Python 3 as the language

Could also be a good use of `typeInstantiationToString` since I don't think it's being used currently

Also as a side note I realized we're defining the Python method wrong lol, I think we need a `self` as the first parameter otherwise it's a static method, so sometime in the future we could probably update that (since we'll need to fix tester too to call it as an instance method rather than currently as a static method)

alankbi (approved):
Looks good, thanks for doing this! I did leave one suggestion which might make it a bit cleaner though if you have time before merging

src/main/java/com/codejoust/main/service/generators/PythonDefaultCodeGeneratorService.java
zpChris: > Instead a comment actually perhaps we could use Python's built-in type hinting? Didn't realize this was a thing till now but also saw that this is what LeetCode does if you select Python 3 as the language
>
> Could also be a good use of typeInstantiationToString since I don't think it's being used currently
That's a great idea, I'll add that now!

> Also as a side note I realized we're defining the Python method wrong lol, I think we need a self as the first parameter otherwise it's a static method, so sometime in the future we could probably update that (since we'll need to fix tester too to call it as an instance method rather than currently as a static method)
I think we could fix that here as well @alankbi - maybe I could add that to [tester #22](https://github.com/CodeJoustHQ/tester/pull/22) since that PR is very small and could use another addition?

zpChris: Done 👍

zpChris:
Re-requesting review because I added a good deal to this PR stemming from [this discussion](https://github.com/CodeJoustHQ/main/pull/148#discussion_r624234885).

alankbi:
Looks good to me!

PR #150, opened 2021-05-06
Add ability to create authenticated accounts using Firebase

alankbi:
### List of Changes:

* Add authenticated accounts to CodeJoust using Firebase
  * Seemed like the best option with completely free authentication if not using phone numbers
  * The firebase project is connected to our existing Google Cloud rocketden-tester project
  * Configuration, user accounts, etc. can be found at [console.firebase.google.com](console.firebase.google.com)
* Create a login and register page with email or Google authentication options
* Change navbar header to dynamically display links based on auth status
* Persist auth status across different pages using Redux
  * Firebase also handles persisting auth status across browser windows and tabs
* Add basic dashboard page and create private routes for problem pages
  * If not authenticated, you'll be redirected to the login page and redirected back after logging in

### Notes:

* Ignore the lines changed - 1000 comes from the firebase packages that needed to be downloaded so the actual amount is less
* Two issues with Google authentication currently:
  * After logging in, you won't be redirected back to the referring page because logging in through Google redirects to their page before redirecting back to `/login`, which loses the referrer URL in location state
  * On the Google OAuth page, it shows `rocketden-tester.firebaseapp.com` as the URL. I briefly tried [changing this](https://firebase.google.com/docs/auth/web/google-signin#customizing-the-redirect-domain-for-google-sign-in) to auth.codejoust.co, but seems like there are some issues with [Cloudflare and SSL](https://stackoverflow.com/questions/35143193/cloudflare-and-nginx-too-many-redirects)
    * I didn't spend too much time on this but it seems like it might require some extra effort with setting up better SSL - if so, maybe we can pair program next week @zpChris?
* The initial page load time slightly increased due to the need to wait for firebase to load in the auth status saved in the browser (otherwise it will cause a lot of bugs). Right now this is simply the current `<Loading />` component, but I'm open to changing that if we have any ideas for design.
* The dashboard is empty right now but I figure that is where users will be eventually able to view their own problems/rooms/etc. 

### Screencast and Images:

[Screencast of new behavior](https://www.loom.com/share/68b0128bfc0443a290161fe66007cd37)

Login page:
Login page

Dashboard with different navbar:
Dashboard with different navbar

Firebase console with list of user accounts:
Firebase console with list of user accounts

### DIFF
--- a/frontend/src/api/Firebase.ts
+++ b/frontend/src/api/Firebase.ts
@@ -1,0 +1,13 @@
+import firebase from 'firebase/app';
+import 'firebase/auth';
+
+const app = firebase.initializeApp({
+  apiKey: process.env.REACT_APP_FIREBASE_API_KEY,
+  authDomain: process.env.REACT_APP_FIREBASE_DOMAIN,
+  databaseUrl: process.env.REACT_APP_FIREBASE_DATABASE,
+  projectId: process.env.REACT_APP_FIREBASE_PROJECT_ID,
+  storageBucket: process.env.REACT_APP_FIREBASE_STORAGE_BUCKET,
+  messagingSenderId: process.env.REACT_APP_FIREBASE_SENDER_ID,
+});
+
+export default app;
--- a/frontend/src/components/config/GoogleLogin.tsx
+++ b/frontend/src/components/config/GoogleLogin.tsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import styled from 'styled-components';
+import firebase from 'firebase';
+import app from '../../api/Firebase';
+import { PrimaryButton } from '../core/Button';
+
+const GoogleButton = styled(PrimaryButton)`
+  position: relative;
+  display: block;
+  margin: 10px auto;
+  background: ${({ theme }) => theme.colors.white};
+  text-align: center;
+`;
+
+const GoogleImage = styled.img`
+  display: inline;
+  position: absolute;
+  left: 15px;
+  width: 30px;
+  vertical-align: middle;
+`;
+
+const GoogleText = styled.span`
+  color: ${({ theme }) => theme.colors.text};
+`;
+
+type GoogleLoginProps = {
+  successAction: () => void,
+  errorAction: (err: string) => void,
+};
+
+function GoogleLogin(props: GoogleLoginProps) {
+  const { successAction, errorAction } = props;
+
+  const onGoogleSubmit = () => {
+    const provider = new firebase.auth.GoogleAuthProvider();
+    app.auth()
+      .signInWithRedirect(provider)
+      .then(() => successAction())
+      .catch((err) => errorAction(err.message));
+  };
+
+  return (
+    <GoogleButton type="button" className="login-provider-button" onClick={onGoogleSubmit}>
+      <GoogleImage src="https://img.icons8.com/fluent/48/000000/google-logo.png" />
+      <GoogleText> Google</GoogleText>
+    </GoogleButton>
+  );
+}
+
+export default GoogleLogin;
--- a/frontend/src/components/config/GoogleLogin.tsx
+++ b/frontend/src/components/config/GoogleLogin.tsx
@@ -13,11 +13,9 @@ const GoogleButton = styled(PrimaryButton)`
 `;
 
 const GoogleImage = styled.img`
-  display: inline;
   position: absolute;
   left: 15px;
   width: 30px;
-  vertical-align: middle;
 `;
 
 const GoogleText = styled.span`
--- a/frontend/src/components/core/Input.tsx
+++ b/frontend/src/components/core/Input.tsx
@@ -148,8 +148,8 @@ export const NumberInput = styled(Input).attrs(() => ({
   }
 `;
 
-export const TextInput = styled(Input).attrs(() => ({
-  type: 'text',
+export const TextInput = styled(Input).attrs((props) => ({
+  type: (props?.type ? props.type : 'text'),
 }))`
   display: inline-block;
   margin: 0 0.5rem 0 0;
--- a/frontend/src/components/core/Input.tsx
+++ b/frontend/src/components/core/Input.tsx
@@ -148,8 +148,8 @@ export const NumberInput = styled(Input).attrs(() => ({
   }
 `;
 
-export const TextInput = styled(Input).attrs((props) => ({
-  type: (props?.type ? props.type : 'text'),
+export const TextInput = styled(Input).attrs(() => ({
+  type: 'text',
 }))`
   display: inline-block;
   margin: 0 0.5rem 0 0;
@@ -164,6 +164,16 @@ export const TextInput = styled(Input).attrs((props) => ({
     border: 2px solid ${({ theme }) => theme.colors.darkBlue};
   }
 `;
+
+export const AuthInput = styled(TextInput)`
+  display: block;
+  margin: 15px auto;
+  width: 20rem;
+`;
+
+export const AuthPasswordInput = styled(AuthInput).attrs(() => ({
+  type: 'password',
+}))``;
 
 export const CheckboxInput = styled(Input).attrs(() => ({
   type: 'checkbox',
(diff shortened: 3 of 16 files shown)


zpChris:
In my opinion, both of the issues you mentioned with Google authentication are small and don't need to be addressed now, but I'll leave the below comments as potential solutions.

> After logging in, you won't be redirected back to the referring page because logging in through Google redirects to their page before redirecting back to /login, which loses the referrer URL in location state

Maybe we could use local storage for this? Especially because we may need that eventually to remove the history props, if that's a change we want to make.

> On the Google OAuth page, it shows rocketden-tester.firebaseapp.com as the URL. I briefly tried changing this to auth.codejoust.co, but seems like there are some issues with Cloudflare and SSL
> I didn't spend too much time on this but it seems like it might require some extra effort with setting up better SSL - if so, maybe we can pair program next week @zpChris?

I am down to pair program on this, but maybe farther into the future, since I don't think it'd be high priority for teachers or students? (Unless it was a personal preference of yours to have this implemented earlier.)

frontend/src/redux/Account.ts
```diff
@@ -0,0 +1,31 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+/**
+ * Due to serialization issues, we can't directly store the
+ * firebase.User object in Redux (instead we store a toJson
+ * object version). This means we'll have to manually specify
+ * the type and copy any relevant fields from the docs below:
+ * https://firebase.google.com/docs/reference/js/firebase.User
+ */
+export type AccountType = {
```
zpChris, line 10: Would additional account attributes like access level go here as well, and be linked to the SQL database?

frontend/src/redux/Account.ts
```diff
@@ -0,0 +1,31 @@
...
+  displayName: string,
+  email: string,
+  emailVerified: boolean,
+  refreshToken: string,
+  uid: string,
```
zpChris, line 15: Why are these attributes needed on the frontend? I could potentially see users for `uid` as the `userId` as we've briefly discussed, but not sure about these attributes' uses in this PR?

frontend/src/components/config/GoogleLogin.tsx
```diff
@@ -0,0 +1,51 @@
...
+  background: ${({ theme }) => theme.colors.white};
+  text-align: center;
+`;
+
+const GoogleImage = styled.img`
+  display: inline;
+  position: absolute;
+  left: 15px;
+  width: 30px;
+  vertical-align: middle;
```
zpChris, line 20:
```suggestion
  position: absolute;
  left: 15px;
  width: 30px;
```
Remove those attributes because I don't think they actually do anything here.

frontend/src/components/core/Input.tsx
```diff
@@ -148,8 +148,8 @@ export const NumberInput = styled(Input).attrs(() => ({
   }
 `;
 
-export const TextInput = styled(Input).attrs(() => ({
-  type: 'text',
+export const TextInput = styled(Input).attrs((props) => ({
+  type: (props?.type ? props.type : 'text'),
 }))`
```
zpChris, line 153: Given that this component is named `TextInput`, not sure about being able to change its type - maybe rename the component or make a new one specific for that other use?

zpChris (approved):
LGTM! ✅ 

I left a few comments, but this PR looks great - kudos on finding Firebase, such a quick implementation. 👍 Definitely gonna use that in the future for anything that needs login and whatnot 💯

frontend/src/redux/Account.ts
alankbi: I think this is only for the attributes provided in Firebase's user object - in the next PR though I expanded/refactored this redux state to include both the firebase user and also a custom account type, which would be linked to our backend database.

frontend/src/components/core/Input.tsx
alankbi: Good point, this was mainly a workaround to get passwords to work, but I created a separate component for that as suggested. I also actually refactored the LoginInput and RegisterInput into a single AuthInput and AuthPasswordInput in the `Input.tsx` file, cause otherwise there'd be four almost duplicate components across the login and register files (LoginInput, LoginPasswordInput, RegisterInput, RegisterPasswordInput).

frontend/src/api/Firebase.ts
```diff
@@ -0,0 +1,13 @@
+import firebase from 'firebase/app';
+import 'firebase/auth';
+
+const app = firebase.initializeApp({
+  apiKey: process.env.REACT_APP_FIREBASE_API_KEY,
+  authDomain: process.env.REACT_APP_FIREBASE_DOMAIN,
+  databaseUrl: process.env.REACT_APP_FIREBASE_DATABASE,
+  projectId: process.env.REACT_APP_FIREBASE_PROJECT_ID,
+  storageBucket: process.env.REACT_APP_FIREBASE_STORAGE_BUCKET,
+  messagingSenderId: process.env.REACT_APP_FIREBASE_SENDER_ID,
```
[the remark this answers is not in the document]
alankbi: Sure I can do that - just wanted to bring up though if in the future we wanted to have different environments (similar to the Heroku Review Apps), this could be useful as we could [customize settings](https://create-react-app.dev/docs/deployment/#customizing-environment-variables-for-arbitrary-build-environments) based on the environment such as staging, production, etc. (in like `.env.staging` and `.env.production` files). Thoughts? That could be further in the future tho so I can revert it in the meantime if you prefer

frontend/src/redux/Account.ts
alankbi: These come from Firebase's [User type](https://firebase.google.com/docs/reference/js/firebase.User), which includes these properties plus a lot of other miscellaneous fields. Let me know if you want me to remove all but the uid and email fields - the other fields I think will still exist in the actual objects, but if we remove them from the TypeScript definition then we'd basically be able to ignore them/act like they don't exist

frontend/src/api/Firebase.ts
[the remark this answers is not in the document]
zpChris: Got it - in that case I'm ok with keeping this file, if you think that's unlikely for the Firebase variables then move them directly into `Firebase.ts`, but if you think that's possible sure we can keep them separate.

frontend/src/redux/Account.ts
zpChris: If we don't use them here, I think it would make sense to remove them, similar to how we only include relevant fields in DTOs. If for some reason you want to keep all of them though, that's OK.

frontend/src/components/core/Input.tsx
zpChris: Looks good, thanks! 👍

frontend/src/api/Firebase.ts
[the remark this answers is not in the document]
alankbi: Sounds good, I'll keep them separate for now

frontend/src/redux/Account.ts
alankbi: Sounds good - I removed everything but uid and email, which are two fields that we do use.

alankbi:
Just realized I haven't replied to your comments yet:

> Right now, I think that's ok, maybe in the future we can have a logo animation, like the little rounded rectangles in the logo all moving up and down or something - that would be ambitious though, we can just have a spinner or keep this current setup too haha.

Haha I like that animated logo idea, it's one of those unimportant but cool features that we can add if we ever have extra free time haha

> In my opinion, the two issues you mentioned with Google authentication are small and don't need to be addressed now, but I'll leave the below comments as potential solutions.

👍 

> Maybe we could use local storage for this? Especially because we may need that eventually to remove the history props, if that's a change we want to make.

Yeah I think that could work - there's a lot of stuff that we've talked about switching to local storage so this might be another good use case for that. I might leave that for a later PR though just to get this series merged in though

> I am down to pair program on this, but maybe farther into the future, since I don't think it'd be high priority for teachers or students? (Unless it was a personal preference of yours to have this implemented earlier.)

Yeah not too big of a deal especially since we're in the beta stage - maybe once we move past that stage is when it'd be good to pick that up

PR #152, opened 2021-05-09
use Submissions while reloading changes, and other minor fixes

jasonz6688:
### List of Changes:

- Removed code and language field in PlayerDTO and frontend player object
- Add problem index to submission's object and request's object
- Setting submissions and code on reload now determined by submission object passed from backend to frontend
- renamed a lot of objects into more correct words

### Notes:

I did not remove the code and language fields from the backend player object because there is a LiveUpdateService that uses it. LiveUpdateService is not actually implemented, but could be in the future. Run solutions are now added onto submissions, so that the frontend can display the submissions; to do this, I had to modify a test about submissions; this does not affect the scores. I also had to delete some assertion statements concerning playerCode and playerLanguage in the DTO.

### Screencast and Images:

Loom video of me explaining some stuff:
https://www.loom.com/share/65084292ce65404ba923738bf5a2aad4

### DIFF
--- a/frontend/src/views/Game.tsx
+++ b/frontend/src/views/Game.tsx
@@ -116,10 +116,10 @@ function GamePage() {
   const [players, setPlayers] = useState<Player[]>([]);
   const [gameTimer, setGameTimer] = useState<GameTimer | null>(null);
   const [problems, setProblems] = useState<Problem[]>([]);
-  const [languageList, setLanguageList] = useState<Language[]>([Language.Python]);
+  const [languageList, setLanguageList] = useState<Language[]>([Language.Java]);
   const [codeList, setCodeList] = useState<string[]>(['']);
   const [currentSubmission, setCurrentSubmission] = useState<Submission | null>(null);
-  const [currentProblem, setCurrentProblem] = useState<number>(0);
+  const [currentProblemIndex, setCurrentProblem] = useState<number>(0);
   const [timeUp, setTimeUp] = useState(false);
   const [allSolved, setAllSolved] = useState(false);
   const [defaultCodeList, setDefaultCodeList] = useState<DefaultCodeType[]>([]);
@@ -152,7 +152,7 @@ function GamePage() {
 
   const createCodeLanguageArray = () => {
     while (languageList.length < problems.length) {
-      languageList.push(Language.Python);
+      languageList.push(Language.Java);
     }
 
     while (codeList.length < problems.length) {
@@ -163,13 +163,14 @@ function GamePage() {
   createCodeLanguageArray();
 
   const setOneCurrentLanguage = (newLanguage: Language) => {
-    languageList[currentProblem] = newLanguage;
+    languageList[currentProblemIndex] = newLanguage;
   };
 
   const setOneCurrentCode = (newCode: string) => {
-    codeList[currentProblem] = newCode;
+    codeList[currentProblemIndex] = newCode;
   };
 
+  // Returns the most recent submission made for problem of index curr.
   const getSubmission = (curr: number, playerSubmissions: Submission[]) => {
     for (let i = playerSubmissions.length - 1; i >= 0; i -= 1) {
       if (playerSubmissions[i].problemIndex === curr) {
@@ -197,8 +198,8 @@ function GamePage() {
       const codeMap = result;
 
       for (let i = 0; i < result.length; i += 1) {
-        newCodeList.push(result[i][Language.Python]);
-        newLanguageList.push(Language.Python);
+        newCodeList.push(result[i][Language.Java]);
+        newLanguageList.push(Language.Java);
       }
 
       // If previous code and language specified, save those as defaults
@@ -358,9 +359,9 @@ function GamePage() {
     const request = {
       initiator: currentUser!,
       input,
-      code: codeList[currentProblem],
-      language: languageList[currentProblem],
-      problemIndex: currentProblem,
+      code: codeList[currentProblemIndex],
+      language: languageList[currentProblemIndex],
+      problemIndex: currentProblemIndex,
     };
 
     runSolution(roomId, request)
@@ -370,7 +371,7 @@ function GamePage() {
         // Set the 'test' submission type to correctly display result.
         res.submissionType = SubmissionType.Test;
         submissions.push(res);
-        setCurrentSubmission(getSubmission(currentProblem, submissions));
+        setCurrentSubmission(getSubmission(currentProblemIndex, submissions));
         checkSendTestCorrectNotification(res);
       })
       .catch((err) => {
@@ -385,9 +386,9 @@ function GamePage() {
     setError('');
     const request = {
       initiator: currentUser!,
-      code: codeList[currentProblem],
-      language: languageList[currentProblem],
-      problemIndex: currentProblem,
+      code: codeList[currentProblemIndex],
+      language: languageList[currentProblemIndex],
+      problemIndex: currentProblemIndex,
     };
 
     submitSolution(roomId, request)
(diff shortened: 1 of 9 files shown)


frontend/src/views/Game.tsx
```diff
@@ -150,27 +151,38 @@ function GamePage() {
...
-    while (currentLanguage.length < problems.length) {
-      currentLanguage.push(Language.Python);
+    while (languageList.length < problems.length) {
+      languageList.push(Language.Python);
     }
 
-    while (currentCode.length < problems.length) {
-      currentCode.push('');
+    while (codeList.length < problems.length) {
+      codeList.push('');
```
zpChris, line 159: This later changes to default code, correct?

src/test/java/com/codejoust/main/api/GameTests.java
```diff
@@ -341,7 +341,7 @@ public void runCodeSuccess() throws Exception {
         // Confirm that running the code does not create a submission.
         assertEquals(1, gameDto.getPlayers().size());
         PlayerDto player = gameDto.getPlayers().get(0);
-        assertEquals(0, player.getSubmissions().size());
+        // assertEquals(0, player.getSubmissions().size());
```
zpChris, line 344: Either un-comment or delete the line.

frontend/src/views/Game.tsx
```diff
@@ -150,27 +151,38 @@ function GamePage() {
...
-    currentLanguage[currentProblem.valueOf()] = newLanguage;
+    languageList[currentProblem] = newLanguage;
   };
 
   const setOneCurrentCode = (newCode: string) => {
-    currentCode[currentProblem.valueOf()] = newCode;
+    codeList[currentProblem] = newCode;
+  };
+
+  const getSubmission = (curr: number, playerSubmissions: Submission[]) => {
```
zpChris, line 173: nit: Add a comment above this method to explain what this does.

frontend/src/views/Game.tsx
```diff
@@ -180,26 +192,35 @@ function GamePage() {
...
-      const codeList = [];
+      const newCodeList = [];
+      const newLanguageList = [];
+      const codeMap = result;
 
       for (let i = 0; i < result.length; i += 1) {
-        codeList.push(result[i][language[i]]);
+        newCodeList.push(result[i][Language.Python]);
+        newLanguageList.push(Language.Python);
       }
```
zpChris, line 202: What is happening here - why is only `Language.Python` being used?

frontend/src/views/Game.tsx
```diff
@@ -180,26 +192,35 @@ function GamePage() {
...
+      for (let i = 0; i < result.length; i += 1) {
+        const temp = getSubmission(i, playerSubmissions);
+
+        if (temp != null) {
+          newCodeList[i] = temp.code;
+          codeMap[i][temp.language as Language] = temp.code;
+          newLanguageList[i] = temp.language as Language;
+          setCurrentProblem(i);
+        }
       }
```
zpChris, line 214: So is this looping through all default code setups, and if there is a submission with that language and problem, store that in the local variables?

frontend/src/views/Game.tsx
```diff
@@ -180,26 +192,35 @@ function GamePage() {
...
 
-      // Set this user's current code and language
-      setCurrentCode(codeList);
-      setCurrentLanguage(language);
-
-      setDefaultCodeList(result);
+      // Set this user's current code
+      setCodeList(newCodeList);
+      setLanguageList(newLanguageList);
+      setDefaultCodeList(codeMap);
```
zpChris, line 219: Does `setDefaultCodeList` need to be added to the dependencies?

frontend/src/views/Game.tsx
```diff
@@ -275,15 +296,15 @@ function GamePage() {
 
           // If this user refreshed and has already submitted code, load and save their latest code
           res.players.forEach((player) => {
-            if (player.user.userId === location.state.currentUser.userId && player.code) {
-              setDefaultCodeFromProblems(res.problems, player.code, [player.language as Language]);
+            if (player.user.userId === location.state.currentUser.userId) {
+              setDefaultCodeFromProblems(res.problems, player.submissions);
```
zpChris, line 300: Can you explain the change that was made here?

zpChris (changes requested):
The changes work perfectly when I run your code and try it out, so awesome job!

I'm having trouble understanding what changes were made and the reasoning behind those updates - especially the changes that were done to the backend, as well as those done to `Game.tsx`. Do you think you could walk through them on a Loom screencast, or explain them on a call at some point?

I left a few comments, but that's really the only reason I'm requesting changes, just so you can help me better understand the code.

frontend/src/views/Game.tsx
jasonz6688: Yes. This is just so that the code can load in the default code into the array without causing an IndexOutOfBounds exception.

frontend/src/views/Game.tsx
jasonz6688: This creating the starting code list and language list for the frontend page. So this essentially just grabs default code from the Python code for each problem from the backend request, adds it to the code list, and adds python to the language list. I'll switch this to Java.

frontend/src/views/Game.tsx
jasonz6688: This loops through the problems, and using getSubmission() it gets the latest submission made for that problem. From the submission, it grabs the code and language submitted, and sets the languageList, codeList, and codeMap accordingly. So what this really does is get the most recent submitted code for each problem from the submissions the backend passes back on reload, preserving code across reloads.

frontend/src/views/Game.tsx
jasonz6688: So, whenever the page was revisited by someone who left, this callback is called. Previously, when this happened, setDefaultCodeFromProblems() took a field for code and language, as their was only one problem to save for. However, since there are multiple problems, the method has been changed to take the submissions from which the code is now taken.

frontend/src/views/Game.tsx
jasonz6688: I checked, and it seems like it doesn't, but I don't really understand why. So should I remove `setDefaultCodeList` from the dependency list?

frontend/src/views/Game.tsx
zpChris: Ah got it - thanks for the explanation! 👍

zpChris (approved):
LGTM! ✅ 

Sorry it took so long to review this PR! I will note that the `Game.tsx` page and the saving submission process is getting fairly complicated; however, I think it will be simplified in the future as we can use localstorage to save submissions, rather than load them from the backend. (This would prevent having to loop through past submissions for each problem, which is fairly inefficient.)

I had one other clarifying question: if there are multiple submissions for the same problem that use different languages, only the latest submission language will be loaded, correct?

Going forward, here are some changes I think would be useful before merging `stagingNumProblems` into `main`:
- Revise the frontend UI for the "Prev" and "Next" buttons.
- Do not have the "Prev" and "Next" buttons wrap-around to the other problems, and/or show what problem number this is out of the list (i.e., 1/5).

I may also be interested in removing the "run code" from the submissions list - curious about @alankbi's thoughts on that too. Lastly, this will likely be _after_ merging, but eventually I think we will simply allow however many problems are selected on the lobby page and remove the "random selection" process.

All of the above is just notes for the future, so you can go ahead and merge this into `stagingNumProblems`.

jasonz6688:
[Change storage system to use Submissions](https://trello.com/c/u3ggSy2J/387-change-storage-system-to-use-submissions)

PR #153, opened 2021-05-11
Write tests for new account behavior

alankbi:
### List of Changes:

* Update existing tests to provide authorization header and continue working properly
* Write unit and API tests to cover newly introduced behavior
  * Accounts, authorization tokens, Firebase service, etc. 

### Notes:

* Because of not being able to easily mock the Firebase library, I introduced a config value in `application.properties` to always return a dummy UID in `FirebaseService.java` if in a testing environment (similar to what's done for `SubmitService.java`

### DIFF
--- a/frontend/src/components/config/App.tsx
+++ b/frontend/src/components/config/App.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import ReactGA from 'react-ga';
+import { hotjar } from 'react-hotjar';
 import { Switch, useLocation } from 'react-router-dom';
 import MainLayout from '../layout/Main';
 import LandingPage from '../../views/Landing';
@@ -28,6 +29,8 @@ import Loading from '../core/Loading';
 
 // Set up Google Analytics
 ReactGA.initialize('UA-192641172-2');
+// Set up Hotjar
+hotjar.initialize(2398506, 6);
 
 function App() {
   const location = useLocation();
--- a/src/test/java/com/codejoust/main/mapper/ProblemMapperTests.java
+++ b/src/test/java/com/codejoust/main/mapper/ProblemMapperTests.java
@@ -1,5 +1,6 @@
 package com.codejoust.main.mapper;
 
+import com.codejoust.main.model.Account;
 import com.codejoust.main.util.TestFields;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.test.context.SpringBootTest;
@@ -28,6 +29,10 @@ public class ProblemMapperTests {
         expected.setName(TestFields.PROBLEM_NAME);
         expected.setDescription(TestFields.PROBLEM_DESCRIPTION);
         expected.setDifficulty(ProblemDifficulty.HARD);
+
+        Account account = new Account();
+        account.setUid(TestFields.UID);
+        expected.setOwner(account);
 
         ProblemTestCase testCase = new ProblemTestCase();
         testCase.setInput(TestFields.INPUT);
@@ -62,6 +67,7 @@ public class ProblemMapperTests {
         assertEquals(expectedProblemInputs, actual.getProblemInputs());
 
         assertEquals(expected.getOutputType(), actual.getOutputType());
+        assertEquals(expected.getOwner().getUid(), actual.getOwner().getUid());
     }
 
     @Test
--- a/src/test/java/com/codejoust/main/socket/RoomSocketTests.java
+++ b/src/test/java/com/codejoust/main/socket/RoomSocketTests.java
@@ -29,6 +29,7 @@ import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.lang.reflect.Type;
+import java.util.Collections;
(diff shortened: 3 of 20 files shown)


src/test/java/com/codejoust/main/util/MockHelper.java
```diff
@@ -15,7 +16,8 @@
 public class MockHelper {
 
     public static <T> T getRequest(MockMvc mockMvc, String url, Class<T> c, HttpStatus status) throws Exception {
-        MvcResult result = mockMvc.perform(get(url))
+        MvcResult result = mockMvc.perform(get(url)
+                .header(HttpHeaders.AUTHORIZATION, TestFields.TOKEN))
```
zpChris, line 20: Are there other `GET` requests in which the Header is not necessary, but is simply added to all the mocks and tests because if it isn't necessary, it doesn't hurt to include it?

frontend/src/components/config/App.tsx
```diff
@@ -28,6 +29,8 @@ import Loading from '../core/Loading';
 
 // Set up Google Analytics
 ReactGA.initialize('UA-192641172-2');
+// Set up Hotjar
+hotjar.initialize(2398506, 6);
```
zpChris, line 33: On our Hotjar account, it is recording sessions (interesting to look at), but I don't see heatmaps - I think I remember you mentioned that you weren't sure why, do you know if more data is needed there or not sure?

src/test/java/com/codejoust/main/mapper/ProblemMapperTests.java
```diff
@@ -62,6 +67,7 @@ public void entityToDto() {
         assertEquals(expectedProblemInputs, actual.getProblemInputs());
 
         assertEquals(expected.getOutputType(), actual.getOutputType());
+        assertEquals(expected.getOwner().getUid(), actual.getOwner().getUid());
```
zpChris, line 70: Can the accounts themselves have compared equality? Could you do `assertEquals(expected.getOwner(), actual.getOwner());` and expect the same result?

src/test/java/com/codejoust/main/socket/RoomSocketTests.java
```diff
@@ -62,6 +63,14 @@
 
     @BeforeEach
     public void setup() throws Exception {
+        // Include the test authorization token in each request
+        template.getRestTemplate().setInterceptors(
+                Collections.singletonList((request, body, execution) -> {
+                    request.getHeaders()
+                            .add("Authorization", TestFields.TOKEN);
+                    return execution.execute(request, body);
+                }));
```
zpChris, line 72: Can you explain what's going on here - was this code partially copied from elsewhere? Is this needed for all socket tests since `template` is used rather than the normal REST testing method?

zpChris (approved):
LGTM! ✅

Thanks for adding these tests, this looks good - very few comments and nits.

src/test/java/com/codejoust/main/util/MockHelper.java
alankbi: Yeah pretty much - lmk if you prefer an alternative at all though

frontend/src/components/config/App.tsx
alankbi: I saw the same issue and [found this](https://help.hotjar.com/hc/en-us/articles/360047604633-Why-aren-t-clicks-being-recorded), so I think it has to due with the dynamic HTML. However there's also this option on the Hotjar admin page: 

Option to fetch new screenshot of sorts

After doing this it worked for the home page. No clicks showed up for the other pages though, but if you go to the "Move" tab, then it does show a heatmap, so either there's still something wrong with the clicks or we simply don't have enough traffic with people clicking things?

src/test/java/com/codejoust/main/mapper/ProblemMapperTests.java
alankbi: I think they can if we have the EqualsAndHashCode and they're of the same type - for this one specifically it's comparing Account and AccountUidDto so idk if that would work but in general though I think that should work

src/test/java/com/codejoust/main/socket/RoomSocketTests.java
alankbi: Yeah it was from a StackOverflow post for how to add the authorization headers to TestRestTemplate (since it's different from MockMvc) - I think it's only necessary for RoomSocketTests and GameSocketTests though, whereas RoomHostTests doesn't require any auth so there's none of this code in that one

frontend/src/components/config/App.tsx
zpChris: Ok gotcha - either way no problem, I don't think this is a priority, thanks for troubleshooting the issue anyway 👍

src/test/java/com/codejoust/main/util/MockHelper.java
zpChris: I think that's fine, just wanted to confirm that I understood the setup 👍

alankbi:
I'm gonna merge this into #151 first before merging #151 into main - reason being that #151 has failing tests and test compilation errors that this PR address

PR #154, opened 2021-05-18
Add problem tags on the frontend.

zpChris:
### List of Changes:

- ...

### Notes:

- ...

### Screencast and Images:

...


zpChris:
[Add problem tags to problem on frontend](https://trello.com/c/9xsDxsAK/372-add-problem-tags-to-problem-on-frontend)

frontend/src/components/problem/SelectedDisplay.tsx
```diff
@@ -0,0 +1,135 @@
...
+    </Content>
+  );
+}
+
+type TagProps = {
+  tags: ProblemTag[],
+};
+
+export function FilterAllTagsDisplay(props: TagProps) {
+  const { tags } = props;
```
alankbi, line 105: Should this be in the selector file? Although I guess it's not really selecting things so more of a question of whether filtering aligns more with displaying or selecting

frontend/src/components/problem/Selector.tsx
```diff
@@ -146,4 +152,66 @@ function ProblemSelector(props: ProblemSelectorProps) {
   );
 }
 
-export default ProblemSelector;
+export function TagSelector(props: TagSelectorProps) {
+  const { tags, selectedTags, onSelect } = props;
```
alankbi, line 156: I realized this is tough since there's small differences here and there (such as with the type of problem vs. tag) but are there any options to reduce the amount of code duplication here? One thing off the top of my head is something like: passing the type in the list of items + a function taking in an object of that type and returning the desired html/component (is this similar at all to the createButtonLink thing we have?), maybe also a function that takes in a search query + a problem and returns a boolean to indicate whether it's filtered or not

Maybe another thing could also be to extract out the ref/useEffect logic (possibly into a custom hook) and other non-html logic

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -176,6 +187,19 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
+
+  useEffect(() => {
+    getAllProblemTags()
+      .then((res) => {
+        setAllTags(res);
+      })
+      .catch((err) => {
+        setError(err.message);
+      });
+  }, []);
```
alankbi, line 202: Is it possible to include the majority of this tag logic into a separate component within the same problem folder? Main reason being that this file is already really large and I'm thinking in a future PR where we improve the creation flow, it might break it into multiple steps/pages

alankbi (approved):
Approved! Left a few comments to consider before merging, but otherwise looks great

alankbi:
> Currently, the lobby page problem tag selection does not actually do anything. This is because I anticipate that we will be moving more towards having the room owner select all of the problems for the room, with no random settings. I think eventually we should have the problem tags filter for problems to be selected, but the room is not set up well for that right now. Thoughts, @alankbi?

Yeah I agree, I think the primary focus will def be selecting problems (and maybe random can be used to fill in any remaining problems)


>   * Furthermore, this means right now that when the room host selects problem tags, the other users cannot see those (as they're not connected to the backend).

Yeah this should be fine - also like I guess it's not even too crucial that non-hosts be able to see all these options cause it's not too important for students to see that info

Another completely random thought I had is it could be cool for hosts to name their rooms, like "Mr. Bob AP CS: Loops"

> * I implemented a way to create new problem tags on the frontend problem page, but not delete tags. I am not sure who should have the settings to delete what - this gets fairly complicated in my mind.

Once the account functionality is merged I think this could be a lot simpler - problem deletion is authorized by the owner of the problem, so something similar could be true for tags

What might be more complicated is what we do about public/private problems and tags? We'll somehow have to deal with lots of problems/tags that aren't quality controlled, and filtering through those on the lobby page (or we just have everything but ours be private, but that still might require some filtering on the lobby page)

My main concern is there eventually being 50+ or 100+ tags that makes it hard to filter through them (unless we hard coded the standard ones or sorted by popularity and capped the number displayed

> * I did not add any indication of the problem tags on the game page - do you think I should add this @alankbi?

Yeah I think that's totally fine - it could be pretty simple to add a tag icon in the description pane, but that's very low priority 

> [Problem Tag Frontend Setup](https://www.loom.com/share/830ed835c43e43148b9b96b3c72b95ef)
> 
> **Edit Problem Tag Selection**
> Edit Problem Tag Selection
> 
> **Edit Problem Create and Filter Tag Modal**
> Edit Problem Create and Filter Tag Modal
> 
> **Lobby Page Tag Selection**
> Lobby Page Tag Selection

frontend/src/components/problem/SelectedDisplay.tsx
zpChris: Yeah I'm happy to put it in the selector file, I think it could go in either since it's kind of a combination. My initial intuition was the `SelectedDisplay.tsx` file here, but let me know which you prefer and I can move it @alankbi.

frontend/src/components/problem/Selector.tsx
zpChris: Given that the `ProblemSelector` is a bit different (mostly given the inner difficulty widget), I'm thinking that these could be separate functions, but any future selectors could use the `TagSelector` as a general solution (with `ProblemSelector` still separate). Thoughts @alankbi?

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Done! ✅ 

I think separating the creation flow is an interesting idea for sure, let me know what thoughts you have! At least separating the components in the `ProblemDisplay` folder would be good too. 👍

zpChris:
Should I keep the tag selector widget on the lobby page, or no because that does not actually work? (Or, I could add a note that the part mentioned isn't fully implemented @alankbi?)

I've responded to some of your other comments below.

> Yeah I agree, I think the primary focus will def be selecting problems (and maybe random can be used to fill in any remaining problems)

Teacher feedback would definitely be helpful for this. 👍 

> Another completely random thought I had is it could be cool for hosts to name their rooms, like "Mr. Bob AP CS: Loops"

That's a cool idea! I wonder if teachers would appreciate that or wouldn't pay too much attention to it.

> Once the account functionality is merged I think this could be a lot simpler - problem deletion is authorized by the owner of the problem, so something similar could be true for tags
>
> What might be more complicated is what we do about public/private problems and tags? We'll somehow have to deal with lots of problems/tags that aren't quality controlled, and filtering through those on the lobby page (or we just have everything but ours be private, but that still might require some filtering on the lobby page)
>
> My main concern is there eventually being 50+ or 100+ tags that makes it hard to filter through them (unless we hard coded the standard ones or sorted by popularity and capped the number displayed

This definitely gets complicated, because I agree that if a user creates a "public" tag, which is then used by lots of people , then decides to delete it, there is a big issue. It'd be interesting to see what other sites do (since this is probably a solved problem somewhere), but in my mind eventually people can only create private tags (those that are just specific to them, others can't use them but can view problems with that tag), while everybody can select from a set number of predefined tags (that we create). I agree sorted by popularity (which is just usage) would be a good solution.

frontend/src/components/problem/SelectedDisplay.tsx
alankbi: Yeah it's no big deal tbh so up to you if you lean one way or the other

frontend/src/components/problem/Selector.tsx
alankbi: Sounds good - my only other suggestion would be if lines 164-173 in Selector.tsx could be made into a custom useX hook (since I think all the logic other than setShowTags/setShowProblems is reused) - that should hopefully be very easy/quick to do tho

alankbi:
> Should I keep the tag selector widget on the lobby page, or no because that does not actually work? (Or, I could add a note that the part mentioned isn't fully implemented @alankbi?)

I think it's prob fine to keep it as is (since soon enough we'll probably implement that feature), but up to you if you lean one way or another (also feel free to merge after whatever you decide)

frontend/src/components/problem/Selector.tsx
zpChris: Done! ✅ 

This implementation was a bit different than the others because the `useClickOutside` hook did not return anything, but this seemed to work fine. 😃

PR #155, opened 2021-05-19
Allow host to manually end game 

alankbi:
### List of Changes:

* Create POST endpoint for host to end game before time is up or all players have finished
* Add button on frontend for host to send end game request
* Add relevant test cases for new endpoint and behavior

### Notes:

* There's a small UI bug where after refreshing the page (thus losing host status) the end game button still shows (clicking it will give a permission denied error) - I _think_ this should be resolved once merged with the Redux stuff (which refactors some of the game fetching) but I'm not sure
* At some point in the future if there's a fourth variable to check for game status (right now I think it's gameEnded, allSolved, and timeUp), it might be worth considering refactoring that setup to use a single variable and create a (customizable based on game mode) method/service of sorts that checks when to end the game based on the game mode and stuff sent in

### Screencast and Images:

New end game button


alankbi:
[Allow host to end game immediately](https://trello.com/c/nJVAIQg7/352-allow-host-to-end-game-immediately)

src/main/java/com/codejoust/main/game_object/Game.java
```diff
@@ -28,4 +28,7 @@
 
     // Boolean to hold whether all users have solved the problem.
     private Boolean allSolved = false;
+
+    // Boolean to hold whether the host prematurely ended the game
```
zpChris, line 32: nit, personal preference: "Prematurely" makes it sound like the host did it on accident, so I made a small suggested change to clarify the wording.
```suggestion
    // Boolean to hold whether the host ended the game early
```

src/main/java/com/codejoust/main/service/GameManagementService.java
```diff
@@ -265,7 +266,24 @@ public void updateCode(String roomId, String userId, PlayerCode playerCode) {
...
+        }
+
+        game.setGameEnded(true);
+        handleEndGame(game);
+
+        GameDto gameDto = GameMapper.toDto(game);
+        socketService.sendSocketUpdate(gameDto);
+
+        return gameDto;
+    }
```
zpChris, line 284: Does there have to be any handling on if the game is already ended and this method is called?

Also, I will note that I thought there could arise an issue where the game ends by timer and by manual end game at the same time, but when I tried testing this out by ending the game at around one second, no adverse consequences seemed to occur. 👍

zpChris (approved):
LGTM! ✅ 

I left a few comments and nits, but nothing at all major. I noted down on another comment for this PR that we may want to wait until the Redux PRs are merged, so we can ensure that we resolve the issue you mentioned where the host still has the "End Game" button even after refreshing.

src/main/java/com/codejoust/main/service/GameManagementService.java
alankbi: I don't believe that would cause any issues, since it's just setting that variable to true and canceling the timers - rn I think the one verification we do is if trying to play again when the game hasn't ended yet

However this brings up a good point cause (unless I'm completely missing it) I don't think we do any validation on starting a game or submitting solutions, so that could probably totally screw things up lol. Probably a lower priority since it's an edge case but something to consider

frontend/src/views/Game.tsx
```diff
@@ -397,6 +412,11 @@ function GamePage() {
           <GameTimerContainer gameTimer={gameTimer || null} />
         </FlexCenter>
         <FlexRight>
+          {currentUser?.userId === host?.userId ? (
+            <TextButton onClick={endGameAction}>
+              End Game
+            </TextButton>
+          ) : null}
           <TextButton onClick={() => leaveRoom(history, roomId, currentUser)}>Exit Game</TextButton>
```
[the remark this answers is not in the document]
alankbi: I think that makes sense, I like the first option - to make it more like our current shadowy design, how would you feel about something like this? 

Exit game button

alankbi:
> Got it - in that case, thoughts on waiting to merge this until the Redux PRs are merged? I'll still review this PR, but since there's this bug it may be worth it to fix that before merging.

Sounds good - I think #151 is the last one to be approved so once that goes through I'll merge those before updating this one and merging

> I think that makes sense - would the game status variable hold an enum or something that has all the different game status possibilities?

I was thinking if the game object holds an enum for the game mode, then it could just be a boolean for gameOver and whatever service or method that handles the game status could use the game mode/other details to determine when the game ends. Unless there's ever a game mode that has more than just game active and game over, then yeah an enum probably makes sense

frontend/src/views/Game.tsx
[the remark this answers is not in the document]
zpChris: That redesign looks fantastic! Much better than my short attempts. 👍

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: > I don't believe that would cause any issues, since it's just setting that variable to true and canceling the timers - rn I think the one verification we do is if trying to play again when the game hasn't ended yet

That makes sense - if we ever have statistics or show on the UI how the game ended (out of time, ended game early, everybody submitted), then we may need to update this method - otherwise this probably isn't an issue.

alankbi: I think like preventing users from submitting after the game is over, or from starting a game when it's already active, etc.

alankbi:
Just tested locally and the bug should be resolved - I'll merge this tomorrow in case @zpChris you want to re-review at all

PR #156, opened 2021-05-20
Set spectator on the frontend.

zpChris:
### List of Changes:
- Enable all users to view action cards on hover for the room page.
- Add the spectator status (and update ability) on the frontend for the room.
- Add socket update and test for the spectator status update.
- Add a terminology help modal for the action card and user attributes.

### Notes:
- The new action cards are relatively complicated as they depend on a broad number of factors (the current user, the user in question, the host, the spectator status, and more). Let me know if I should avoid using the tertiary statements in the way I've incorporated them.
- I made the backend update so that users who try to join an active room will be included in the room as a spectator, and no users in that room can change their spectator status to `false` while the game is ongoing.
- I updated the frontend `createRoom` call so that the host will by default be a spectator. (I did not do this before I made the screencast below, which is why that did not show that change.)

### Screencast and Images:

[Set Spectator Settings on the Frontend](https://www.loom.com/share/27c39f48172e40b2a07fa2f47d306431)

**Spectator Icon on the Room Page**
Spectator Icon on the Room Page

**Revised Action Card on the Room Page**
Revised Action Card on the Room Page

**Terminology Modal for Action Card Attributes**
Terminology Modal for Action Card Attributes


zpChris:
[Spectator view setup](https://trello.com/c/Eh4PKYWS/373-spectator-view-setup)

frontend/src/views/Join.tsx
```diff
@@ -73,6 +73,8 @@ function JoinGamePage() {
       getRoom(roomIdParam)
         .then((res) => {
           setLoading(false);
+
+          // TODO: Check to be removed to allow spectator view.
```
alankbi, line 77: Is this TODO still relevant?

frontend/src/views/Lobby.tsx
```diff
@@ -582,6 +636,48 @@ function LobbyPage() {
   // Render the lobby.
   return (
     <>
+      <Modal
+        show={actionCardHelp}
+        onExit={() => setActionCardHelp(false)}
+        fullScreen
```
alankbi, line 642: nit: Move this modal into a separate component

src/main/java/com/codejoust/main/service/RoomService.java
```diff
@@ -363,6 +363,11 @@ public RoomDto setSpectator(String roomId, SetSpectatorRequest request) {
             throw new ApiException(RoomError.NOT_FOUND);
         }
 
+        // Return error if user changes status to non-spectator in active game.
+        if (room.getActive() && !request.getSpectator()) {
+            throw new ApiException(RoomError.ACTIVE_GAME);
+        }
+
```
alankbi, line 370: Are there any situations where we'd allow people currently competing to switch themselves into spectators?

alankbi (approved):
Looks amazing! No major comments or changes. 

I do have a couple thoughts/ideas (not necessarily all for this PR); let me know what you think about the following: 

1. Thoughts on moving the question mark/info button outside of the action card and somewhere automatically visible? Since a lot of those icons are visible without hovering, so it might help in case anyone's confused (and maybe it can include a section at the bottom explaining you can hover over a user to view actions)
2. Now that we have spectators as well, how do we want to sort these when displaying on the page? Right now we have it by active vs. inactive, but wondering if we should include spectators into the mix (if things get too complicated with the lists, we could introduce sorting in the `RoomMapper.toDto` method similar to with the game leaderboard, thus on the frontend all we'd have to do is display the `users` list)

src/main/java/com/codejoust/main/service/RoomService.java
zpChris: I couldn't really think of any, but thought the handicap would only occur when a spectator switches themselves to a non-spectator as they may have seen other people's code, which is why I prevented that here.

I could prevent the other switch too (non-spectator to spectator), but I figured I wouldn't overcomplicate the method without having a clear reason to prevent this functionality.

frontend/src/views/Join.tsx
zpChris: Yes it is; for the next PR, this will need to be removed so people who are joining active games are automatically created as spectators, and are thus actually able to join the active game.

zpChris:
> Looks amazing! No major comments or changes.
> 
> I do have a couple thoughts/ideas (not necessarily all for this PR); let me know what you think about the following:
> 
> 1. Thoughts on moving the question mark/info button outside of the action card and somewhere automatically visible? Since a lot of those icons are visible without hovering, so it might help in case anyone's confused (and maybe it can include a section at the bottom explaining you can hover over a user to view actions)

That makes sense - I'll move it next to the refresh button at the top of this container, and I can include some more sections. This could potentially be written out a bit more in the future, or even include outside links or videos for how this all works.

> 2. Now that we have spectators as well, how do we want to sort these when displaying on the page? Right now we have it by active vs. inactive, but wondering if we should include spectators into the mix (if things get too complicated with the lists, we could introduce sorting in the `RoomMapper.toDto` method similar to with the game leaderboard, thus on the frontend all we'd have to do is display the `users` list)

I agree that active vs. inactive is not the best sorting mechanism, as that may not be a very relevant statistic for teachers. I was thinking that the design could later have the Players at a top section, and then the Spectators in a below section? The only potential issue with that is if people want the host to be more visible, and they would be at the bottom by default.

We could also just display alphabetically, which may simplify the process of finding users, perhaps with the exception of putting the host first. I'm open to suggestions when I further implement all this on future PRs.

PR #158, opened 2021-05-26
Implement dashboard page

alankbi:
### List of Changes:

* Implement dashboard page according to Figma design
* Update problem card design for dashboard and all problems pages
* Update header to show dropdown for account-related actions
* Small bug fix and custom hook refactoring
* Move dashboard to home page if logged in

### Notes:

* I moved the dashboard to the home page as I think was mentioned before - lmk your thoughts on this or whether we should keep it at `/dashboard` (I think some sites like GitHub change the home page based on auth status, others like Kahoot don't, and meanwhile sites like Heroku have a separate page for dashboard but redirect you from the home page if you're logged in
* Getting the dashboard sidebar to look decent was kind of a hassle - right now it relies a lot on hard-coded px values to get it all to work (it handles fine on most screen sizes other than super small)
* The dashboard is currently absolutely positioned (rather than fixed, so it will not scroll with you), and the height is set to fill whatever remaining view height of the browser, so on small heights it might look funky after scrolling down - lmk any thoughts on this
* Showing the description in the problem card looks bad if it starts with markdown (e.g. `###`) - maybe a workaround for this is to consider changing the format of our problem descriptions? Especially if we're going down the simpler problems route, maybe shorter descriptions with fewer sections/heavy markdown formatting could help solve this issue

### Screencast and Images:

No screencast this time :'(

New dashboard page:
New dashboard page

All problems page with new problem card:
All problems page with new problem card


alankbi:
[Implement new dashboard design](https://trello.com/c/oPCifVBu/375-implement-new-dashboard-design)

zpChris:
> * Small bug fix and custom hook refactoring

What was the small bug fix?

> * I moved the dashboard to the home page as I think was mentioned before - lmk your thoughts on this or whether we should keep it at `/dashboard` (I think some sites like GitHub change the home page based on auth status, others like Kahoot don't, and meanwhile sites like Heroku have a separate page for dashboard but redirect you from the home page if you're logged in

I think this setting sounds good! 👍 

> * Getting the dashboard sidebar to look decent was kind of a hassle - right now it relies a lot on hard-coded px values to get it all to work (it handles fine on most screen sizes other than super small)
> * The dashboard is currently absolutely positioned (rather than fixed, so it will not scroll with you), and the height is set to fill whatever remaining view height of the browser, so on small heights it might look funky after scrolling down - lmk any thoughts on this

Thanks for these notes - I'll check this out and include any comments in the review.

> * Showing the description in the problem card looks bad if it starts with markdown (e.g. `###`) - maybe a workaround for this is to consider changing the format of our problem descriptions? Especially if we're going down the simpler problems route, maybe shorter descriptions with fewer sections/heavy markdown formatting could help solve this issue

I think we can remove the starting "### Description" on the actual problem pages, since it looks ok without that - I removed that heading in the template on the PR #160. I think that should largely fix this issue, but lmk!

frontend/src/views/dashboard/MyProblems.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+type MyProblemsProps = {
+  loading: boolean,
+};
+
+const Content = styled.div`
+  text-align: left;
+`;
+
+const TopText = styled.div`
```
zpChris, line 26:
```suggestion
const TopTextContainer = styled.div`
```
nit (personal preference): Thought this was text at first, this would just clear it up a tad.

frontend/src/components/config/App.tsx
```diff
@@ -37,20 +38,22 @@ function App() {
...
   // Track page view on every change in location and clear errors when switching pages
   useEffect(() => {
     ReactGA.pageview(location.pathname);
   }, [location]);
 
   // Set authentication status when Firebase auth status changes
   useEffect(() => {
-    app.auth().onAuthStateChanged((firebaseUser) => {
-      dispatch(setFirebaseUser(firebaseUser?.toJSON() as FirebaseUserType || null));
+    app.auth().onAuthStateChanged((user) => {
```
zpChris, line 50: What was your reasoning for changing this to `user`? I got confused between `firebaseUser` as the account object, and `user` as the `User` object, which is why I'm asking.

frontend/src/components/layout/FullLayout.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import styled from 'styled-components';
+import { MinimalHeader } from '../navigation/Header';
+import { FullContainer } from '../core/Container';
+
+const Content = styled.div`
+  width: 100%;
+  min-height: 100vh;
+  //display: flex;
+  //flex-direction: column;
```
zpChris, line 10: nit: Remove these commented-out lines.

frontend/src/components/core/Grid.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
```
zpChris, line 1: I didn't see this file and its contents used anywhere - is this going to be incorporated someplace in the future, or no? If not, maybe remove this file?

zpChris (changes requested):
This looks awesome! I'm only requesting review to ask some questions about PR design, and to see if flexbox is possible for the Dashboard page design (if not, no problem, just want to know the reasoning behind this implementation).

I included a number of questions / comments below:

- Why did you decide to use hard-coded pixel values and margins instead of flexbox with two columns, like the [Problem Page](https://codejoust.co/problem/791d34a1-5525-4a33-95c3-d604fef6c94d)?
- I also saw you made sub-folders within the `views` folder - should we do the same thing for the problem page? I didn't realize you did this with the `Login` and `Register` pages until now, but either way that made more sense as I thought each file corresponded to a unique path.
- For the Start New Game, Create (Problem), and large Problem buttons, thoughts on making those actual links instead of buttons that push the next link onto history? I now we used the push on the `/problems/all` page, but I realized that if someone tries to open it on a new tab like I did (command + click), it won't do so.
- Also, thoughts on adding a "Join Room" button somewhere on the Dashboard? I know teachers shouldn't really need that, but they might want to just test out that functionality. This could be a text button with arrow, or perhaps the "Start New Game" button could be "Play Game", which on hover shows two options above it that are "Start New Game" or "Join Game".  Thoughts on all that, @alankbi?

frontend/src/views/dashboard/MyProblems.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { useHistory } from 'react-router-dom';
+import { MainHeaderText, SecondaryHeaderText } from '../../components/core/Text';
+import ProblemCard from '../../components/card/ProblemCard';
+import { TextLink } from '../../components/core/Link';
+import { useAppSelector } from '../../util/Hook';
+import {
+  CenteredContainer,
+  FlexHorizontalContainer,
+  FlexLeft,
+  RelativeContainer
```
zpChris, line 12:
```suggestion
  RelativeContainer,
```
nit: Add ending comment to remove the "missing comma dangle" warning. (Sorry, I only saw this after I finished my review, which is why I'm adding this comment after.)

frontend/src/views/dashboard/DashboardSidebar.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <TabItem
+          active={tab === DashboardTab.SUGGEST_FEATURE}
+          onClick={() => window.open('https://airtable.com/shrGkEhC6RhAxRCxG', '_blank')}
+        >
+          <NoMarginSubtitleText>Suggest a Feature</NoMarginSubtitleText>
+        </TabItem>
+      </InnerContent>
+
+      <BottomContent>
+        <PrimaryButton onClick={() => history.push('/game/create')} width="80%">Start New Game</PrimaryButton>
```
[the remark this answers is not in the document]
alankbi: Sure np, sounds good 👍

frontend/src/views/dashboard/MyProblems.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </TopText>
+
+      <FilterContainer>
+        <FilterInput
+          value={filterText}
+          onChange={(e) => setFilterText(e.target.value)}
+          placeholder="Filter (by name, difficulty, tag, etc.)"
+        />
+        <WhiteTextButton onClick={() => setFilterText('')}>✕</WhiteTextButton>
+      </FilterContainer>
```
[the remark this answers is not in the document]
alankbi: Sure, done! Also the second thing is a good point - I changed the implementation to allow you to separate queries by comma. There are probably better more intuitive implementations, but for now I guess that should sort of solve the issue

frontend/src/views/dashboard/MyProblems.tsx
```diff
@@ -0,0 +1,141 @@
...
+        </FlexHorizontalContainer>
+      </TopText>
+
+      <FilterContainer>
+        <FilterInput
+          value={filterText}
+          onChange={(e) => setFilterText(e.target.value)}
+          placeholder="Filter (by name, difficulty, tag, etc.)"
+        />
+        <WhiteTextButton onClick={() => setFilterText('')}>✕</WhiteTextButton>
```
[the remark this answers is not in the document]
alankbi: Sounds good - went with option 1

frontend/src/views/dashboard/MyProblems.tsx
alankbi: Sg - will implement this locally (since the component needs to be refactored as well)

frontend/src/components/config/App.tsx
alankbi: The main reason was after importing `firebaseUser` from the Redux store, this name overshadowed that variable, causing an ESLint warning. Therefore I went with just renaming it to `user` instead

frontend/src/components/core/Grid.tsx
alankbi: Just deleted it, nice catch - this was a remnant of my initial attempt at the dashboard two column layout

frontend/src/views/dashboard/MyProblems.tsx
alankbi: Sg - I think this is outdated now since I included this change in an earlier commit, so I'll resolve now

frontend/src/views/dashboard/DashboardSidebar.tsx
```diff
@@ -0,0 +1,100 @@
...
+const Content = styled.div`
+  position: absolute;
+  width: 300px;
+  top: 100px;
+  left: 30px;
+  bottom: 30px;
+  text-align: left;
+  display: flex;
+  flex-direction: column;
+`;
```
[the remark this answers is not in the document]
alankbi: Thanks for bringing this up, especially the flexbox solution - I just gave it another try and was able to get it to work with your tips :) There's still a bit of hard coding required to get the height of the sidebar to fit the screen size (`height: calc(100vh - 130px`), but overall I think this is an improvement, so feel free to take a look

alankbi:
> * Why did you decide to use hard-coded pixel values and margins instead of flexbox with two columns, like the [Problem Page](https://codejoust.co/problem/791d34a1-5525-4a33-95c3-d604fef6c94d)?

Mentioned this in a comment, but I changed to flexbox using the solution you proposed :) 

> * I also saw you made sub-folders within the `views` folder - should we do the same thing for the problem page? I didn't realize you did this with the `Login` and `Register` pages until now, but either way that made more sense as I thought each file corresponded to a unique path.

Hm that's a good point, hadn't thought of that - I'm kind of indifferent either way, but maybe once we build those out in the future we can deal with the refactoring at the point or something

> * For the Start New Game, Create (Problem), and large Problem buttons, thoughts on making those actual links instead of buttons that push the next link onto history? I now we used the push on the `/problems/all` page, but I realized that if someone tries to open it on a new tab like I did (command + click), it won't do so.

Good catch - done! `createButtonLink` coming in handy here

> * Also, thoughts on adding a "Join Room" button somewhere on the Dashboard? I know teachers shouldn't really need that, but they might want to just test out that functionality. This could be a text button with arrow, or perhaps the "Start New Game" button could be "Play Game", which on hover shows two options above it that are "Start New Game" or "Join Game".  Thoughts on all that, @alankbi?

Sure, added 👍

frontend/src/views/dashboard/DashboardSidebar.tsx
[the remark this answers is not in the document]
zpChris: Awesome, thanks for making that change! I think that hard coding is for sure ok - we may change it in the future, but it's no big deal now.

frontend/src/views/dashboard/MyProblems.tsx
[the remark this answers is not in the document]
zpChris: The comma setup looks great! I like this solution, I think it could stay here for the long-run. 😃

frontend/src/components/config/App.tsx
zpChris: Ah gotcha, sounds good 👍

frontend/src/components/problem/FilteredProblemList.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <FilterInput
+          value={filterText}
+          onChange={(e) => setFilterText(e.target.value)}
+          placeholder="Filter by name, difficulty, or tag (separate queries by comma)"
+        />
+        {filterText ? <WhiteTextButton onClick={() => setFilterText('')}>✕</WhiteTextButton> : null}
+      </FilterContainer>
+
+      {problems.map((problem: Problem) => {
+        const texts = filterText.toLowerCase().split(',');
```
zpChris, line 61: Given this, in the future it'd be smart to disallow commas from problem titles and tags; however, this likely won't arise anyway, so this isn't the biggest concern.

frontend/src/components/card/ProblemCard.tsx
```diff
@@ -1,50 +1,66 @@
...
-            <Text>{problem.difficulty}</Text>
-          </CenteredContainer>
-        </FlexRight>
-      </FlexHorizontalContainer>
+    <Content>
+      <DivLink to={`/problem/${problem.problemId}`}>
+        <InnerContent>
+          <TitleText>{problem.name}</TitleText>
+          {getDifficultyDisplayButton(problem.difficulty, true)}
+          <Text>{`${problem.description.substring(0, 80)}...`}</Text>
```
zpChris, line 55: Note: I changed all of the problems under the support@codejoust.co account to remove the "Description" header in the description.

zpChris (approved):
LGTM! ✅ 

Thanks for making the earlier changes to flexbox, adding links, and more.

I wanted to note one last potential change, but just if you have a strong preference for style (both are fine by me). The dashboard page I think is the only one that isn't centered with the screen - it leans more to the left. We could keep as-is or change to center it more (see below table).

<table>
  <tr>
    <th>Existing Design</th>
    <th>Centered Design</th>
  </tr>
  <tr>
    <td>Existing Design</td>
    <td>Centered Design
</td>
  </tr>
</table>

The change is small, just making the overall container `width: 90%`, adding `margin: 0 auto`, and then changing the problems list container to `width: 100%` (or `90%`, something larger). However, I think both work, this was just me playing around with inspect element.

I'll review #162 soon!

frontend/src/views/dashboard/DashboardSidebar.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type TabItemProps = {
+  active: boolean,
+  onClick: (tab: DashboardTab) => void,
+  enabled?: boolean,
+};
+
+const Content = styled.div`
+  position: relative;
+  width: 300px;
```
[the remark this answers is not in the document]
alankbi: Sounds good - will prob stick with fixed width for now to avoid having to deal with odd proportions, but that's good to know if we wanted to refactor the design in the future at any point

frontend/src/views/dashboard/DashboardSidebar.tsx
```diff
@@ -0,0 +1,107 @@
...
+  text-align: left;
+  display: flex;
+  flex-direction: column;
+  
+  // [Header: height + padding] + [Dashboard: margin-top] + desired padding-bottom
+  height: calc(100vh - 130px);
```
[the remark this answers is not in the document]
alankbi: Good call :)

PR #160, opened 2021-05-27
Refine the problem page.

### LINKED ISSUE #157: Problem page problem input issue
On the create problem and edit problem pages, the problem input does not stay focused; on every letter typed, the user loses the input focus to continue typing. See the screencast below.

[Problem Inputs Issue](https://www.loom.com/share/e5f55dd7f849451eabdec3f76908d362)

zpChris:
### List of Changes:
- Fix #157 by adding the index as a key, muting warning (see notes).
- Refactor `ProblemInputDisplay` as a separate component.
- Add the ability to use a template description when writing the problem description.
- Add a help modal to the problem page.

### Notes:
- The #157 issue was fixed when removing the `key` attribute on the `InputTypeContainer` in the original codebase (it was since changed in this PR). However, when removing that, React gave a warning that each child must have a unique "key" prop. For some reason, when using `generateRandomId()`, the same issue arose; however, when using `index` as the key, it worked fine, only creating the [no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) warning.
Thus, my final solution was to add the `index` as the key and suppress that line warning; despite sub-optimal performance due to this choice, this suppresses warnings and fixes #157, so I viewed it as the best option. @alankbi, if you know how to include `generateRandomId()` while still fixing that bug, I'm happy to update this (this [SO post](https://stackoverflow.com/questions/22573494/react-js-input-losing-focus-when-rerendering) may be relevant).

<details><summary>React no key warning (click here to view screenshot)</summary>
React No Key Warning
</details>

- The "Add a template description" button was implemented by checking if the description was empty (surprisingly, this occurs when the description is an empty String _or_ if it equals `\\\n`, which occurs if you delete a currently-present description). Then, in order to get the markdown editor to display this new description, I had to force a re-render of the markdown editor by attaching a new key to the editor (see this [SO post](https://stackoverflow.com/questions/22573494/react-js-input-losing-focus-when-rerendering) for background).
- There are now help modals on the Problem Page and the Lobby Page; when making relevant updates on these pages, these should be updated. We can look to other solutions if the future if we wish (an external site like [Gimkit Help](https://help.gimkit.com/en/) or some other solution).
- I referenced the new problem [Find Longest String](http://localhost:3000/problem/791d34a1-5525-4a33-95c3-d604fef6c94d) as an example CodeJoust problem in the help modal - let me know what you think about this @alankbi.
- Additional ideas to simplify the problem creation process:
  1. **Problem Templates**: Near the top of the create problem page, there could be a "Use Template" button that let's you look through your problems, or verified public problems, and select them to create a default template for the problem.
  2. **Screencast Walkthrough**: A screencast walkthrough that pops up for a help modal, where we simply walk through the process of creating a new problem.
  3. **Preview Problem**: Similar to the process of writing the code to solve the problem, there could be a preview page to view and test out the problem.

### Screencast and Images:

[Refine the Problem Page](https://www.loom.com/share/1f40799a35a14b01821c602aa9e52cfb)

**Create Problem Page**
Create Problem Page

**Create Problem Page with Template Description**
Create Problem Page with Template Description

**Problem Page Help Modal**
Problem Page Help Modal


alankbi:
> Problem Templates: Near the top of the create problem page, there could be a "Use Template" button that let's you look through your problems, or verified public problems, and select them to create a default template for the problem.

Maybe by default too we could include the desired template pre-filled when first loading the create page? Like some hard-coded string with all the sections and line breaks that we could input for the default `description = useState(defaultDescription)`

zpChris:
[Refine edit problem page](https://trello.com/c/STpVE618/377-refine-edit-problem-page)

zpChris:
> > Problem Templates: Near the top of the create problem page, there could be a "Use Template" button that let's you look through your problems, or verified public problems, and select them to create a default template for the problem.
> 
> Maybe by default too we could include the desired template pre-filled when first loading the create page? Like some hard-coded string with all the sections and line breaks that we could input for the default `description = useState(defaultDescription)`
> 
> Might also make things easier for teachers if it says stuff like “description goes here” or “example input: ...” similar to our PR templates

Hey, that's what I implemented before I saw this comment! I did that but with one difference - it isn't present by default; instead, if the description is empty, there's a button to add that new template description.

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -362,11 +393,25 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
             defaultValue={newProblem.description}
             onChange={(getNewValue) => handleDescriptionChange(getNewValue())}
             readOnly={!problemEditable}
           />
+          {
+            (newProblem.description.length === 0 || newProblem.description === '\\\n') ? (
+              <TextButtonLink
+                onClick={() => {
+                  // Add default description and force refresh editor.
+                  newProblem.description = defaultDescription;
```
alankbi, line 407: nit: Change this to use `handleDescriptionChange` to avoid directly mutating state variables (also not sure where the refresh editor comes into play but I feel like doing this change might solve that issue)

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -138,21 +153,24 @@ const NoMarginTopText = styled(Text)`
...
   color: ${({ theme }) => theme.colors.red2};
   background: ${({ theme }) => theme.colors.white};
 `;
 
+const TextButtonLink = styled(TextButton)`
+  padding: 0;
+  margin-top: 10px;
+  text-decoration: underline;
+  text-align: left;
+`;
```
alankbi, line 167: Small styling suggestions: maybe move this down a few lines (like so it's shown a little farther down when no description exists), and/or also maybe make this text gray?

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -528,47 +573,17 @@ function ProblemDisplay(props: ProblemDisplayParams) {
...
-                  onClick={() => deleteProblemInput(index)}
-                >
-                  ✕
-                </CancelTextButton>
-              ) : null}
-            </InputTypeContainer>
+            <ProblemInputDisplay
+              // eslint-disable-next-line react/no-array-index-key
+              key={index}
+              input={input}
```
alankbi, line 579: Was refactoring this code into ProblemInputDisplay part of the solution to the bug fix you mentioned? If not I feel like the amount of props needed to pass this component makes the code feel a bit rigid - maybe instead we could either figure out some way to reduce the number of props, or even instead move the entirety of the options sidebar into its own component? 

For example if we move the options sidebar into its own component I think it'd only require 4 props (problemEditable, editMode, newProblem, setNewProblem - tho in the near future I'm thinking we can move newProblem to the Redux store which would cut this to 2) while also cutting the size of the ProblemDisplay file almost in half

alankbi (approved):
Sorry for the delay in reviewing - this looks good 👍 

Left a small number of comments about things here and there, but otherwise looks great :)

frontend/src/components/problem/ProblemDisplay.tsx
```diff
@@ -40,11 +35,31 @@ import {
...
+  '\\',
+  'Replace this with an explanation of the output, if necessary.',
+  '### Example Test Case',
+  '**Input**',
+  '`name`: `5`',
+  '**Output**',
+  '`1`',
+  '**Explanation**',
+  'Replace this line with an explanation of this example test case.',
+].join('\n\n');
```
alankbi, line 62: Another thing I was thinking as well was thoughts on simplifying this template a bit? Looking at LeetCode and Binary Search (and to a lesser extent Practice It), generally they include the parameters in the description (given X and Y parameters, write a method Z that returns W) and then simply have a section for example test cases (as we do now), which I think might be easier for teachers to understand if they end up writing simpler problems

alankbi:
Random thought, but what do you think about renaming the Problem Inputs and Problem Output titles to be Parameters and Return Type? Wondering if that would help clear things up at all

zpChris:
> Random thought, but what do you think about renaming the Problem Inputs and Problem Output titles to be Parameters and Return Type? Wondering if that would help clear things up at all

I think that's a great idea - I'll do that in this PR! I think that will help clarify that problem setup when teachers are creating new problems.

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: The refactoring was not part of the solution - and that's a fair point. I might lean towards just undoing the creation of this new component right now, since you suggested we may make even more changes to this whole page. I'm thinking we could do the refactoring / separating components there as necessary, but let me know if I should do it here instead.

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: I agree - I'll play around with this and remove some of the unnecessary lines.

frontend/src/components/problem/ProblemDisplay.tsx
alankbi: Yeah I think that sounds good - we can do that once we get around to that refactoring PR

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Done! ✅ 

That change alone doesn't fix the refresh editor issue, so I kept in the previous solution of increasing the key as well.

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: I made this text the same color as the text placeholder, and moved it `20px` farther down; however, I don't think it's super visible. Any thoughts or advice here @alankbi? If you have a color or margin in mind that'd help, no problem if not.

**New Template Description Text**
New Template Description Text

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Ok, I reverted this change but kept the small fix that resolves #157. 👍

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: What do you think about the following change? I removed everything mentioned except the "Parameters" list - I could remove that too, I think that can be helpful though, let me know.

**New Template Description**
New Template Description

frontend/src/components/problem/ProblemDisplay.tsx
alankbi: Is this the default gray or a slightly different color? If the default gray doesn't look good, then I'm fine with reverting it to the original color, either or

frontend/src/components/problem/ProblemDisplay.tsx
alankbi: Sure, I think this looks good - I guess depending on teacher feedback if people have any confusions we can continue to refine this once we get to that step

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: This is the placeholder gray that is in the above Markdown setup. I'll try switching around on different grays, and seeing what looks best 👍

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: Ok cool, sounds great! 😃

frontend/src/components/problem/ProblemDisplay.tsx
zpChris: I switched to the default gray as you said, that looked good!

PR #167, opened 2021-06-08
Changing Scoring Criteria and End Game Criteria

jasonz6688:
### List of Changes:

- Score is now calculated as the number of problems which you correctly solve
- Games now only after everyone solves every problem correctly

### Notes:

I removed run code requests from submissions, which means I can add back the test concerning run code and submissions. Also, on my computer, after refreshing, the results of past submissions are no longer displayed in the console; the code and "score" are still saved. Additionally, something is broken with the whole socket system; everyone once in a while, java will throw an exception that the client has closed their connection. It doesn't have any effect on the game or the processes, but a long error message is printed out.

### Screencast and Images:
https://www.loom.com/share/40f862b9089849caa63c98360bf14a15

### DIFF
--- a/frontend/src/components/results/PlayerResultsItem.tsx
+++ b/frontend/src/components/results/PlayerResultsItem.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import { Player, Submission } from '../../api/Game';
 import { LowMarginText, Text } from '../core/Text';
 import { Color } from '../../api/Color';
-import useGetScore, { useBestSubmission } from '../../util/Hook';
+import useGetScore, { useBestSubmission, useGetSubmissionTime } from '../../util/Hook';
 import Language, { displayNameFromLanguage } from '../../api/Language';
 import { TextButton } from '../core/Button';
 
@@ -82,15 +82,17 @@ type PlayerResultsCardProps = {
   isCurrentPlayer: boolean,
   gameStartTime: string,
   color: Color,
+  numProblems: number,
   onViewCode: () => void,
 };
 
 function PlayerResultsItem(props: PlayerResultsCardProps) {
   const {
-    player, place, isCurrentPlayer, color, gameStartTime, onViewCode,
+    player, place, isCurrentPlayer, color, gameStartTime, numProblems, onViewCode,
   } = props;
 
   const score = useGetScore(player);
+  const time = useGetSubmissionTime(player);
   const bestSubmission : Submission | null = useBestSubmission(player);
 
   const getDisplayNickname = () => {
@@ -103,21 +105,19 @@ function PlayerResultsItem(props: PlayerResultsCardProps) {
       return '0';
     }
 
-    const percent = Math.round((score / 12321) * 100);
+    const percent = Math.round((score / numProblems) * 100);
     return `${percent}%`;
   };
 
   const getSubmissionTime = () => {
-    if (!bestSubmission) {
-      return 'N/A';
+    if (!time) {
+      return 'Never';
     }
 
-    // Calculate time from start of game till best submission
-    const startTime = new Date(gameStartTime).getTime();
-    const diffMilliseconds = new Date(bestSubmission.startTime).getTime() - startTime;
+    const currentTime = new Date().getTime();
+    const diffMilliseconds = currentTime - new Date(time).getTime();
     const diffMinutes = Math.floor(diffMilliseconds / (60 * 1000));
-
-    return ` ${diffMinutes} min`;
+    return `${diffMinutes}m ago`;
   };
 
   const getSubmissionCount = () => player.submissions.length || '0';
--- a/frontend/src/components/results/PlayerResultsItem.tsx
+++ b/frontend/src/components/results/PlayerResultsItem.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import { Player, Submission } from '../../api/Game';
 import { LowMarginText, Text } from '../core/Text';
 import { Color } from '../../api/Color';
-import useGetScore, { useBestSubmission, useGetSubmissionTime } from '../../util/Hook';
+import { useBestSubmission, useGetScore, useGetSubmissionTime } from '../../util/Hook';
 import Language, { displayNameFromLanguage } from '../../api/Language';
 import { TextButton } from '../core/Button';
 
@@ -111,11 +111,11 @@ function PlayerResultsItem(props: PlayerResultsCardProps) {
 
   const getSubmissionTime = () => {
     if (!time) {
-      return 'Never';
+      return 'N/A';
     }
 
-    const currentTime = new Date().getTime();
-    const diffMilliseconds = currentTime - new Date(time).getTime();
+    const startTime = new Date(gameStartTime).getTime();
+    const diffMilliseconds = new Date(time).getTime() - startTime;
     const diffMinutes = Math.floor(diffMilliseconds / (60 * 1000));
     return `${diffMinutes}m ago`;
   };
--- a/frontend/src/util/Hook.tsx
+++ b/frontend/src/util/Hook.tsx
@@ -27,7 +27,8 @@ export const useBestSubmission = (player?: Player) => {
 };
 
 const useGetScore = (player?: Player) => {
-  const counted = new Set<Number>();
+  const counted = new Set<number>();
+  const [score, setScore] = useState<number>(0);
 
   useEffect(() => {
     if (player) {
@@ -37,37 +38,39 @@ const useGetScore = (player?: Player) => {
           counted.add(player.submissions[i].problemIndex);
         }
       }
+
+      setScore(counted.size);
     }
-  }, [player]);
+    
+  }, [player, setScore]);
 
   if (player == null || player.submissions.length === 0) {
     return null;
   }
-
-  return counted.size;
+  
+  return score;
 };
 
 export default useGetScore;
 
 export const useGetSubmissionTime = (player?: Player) => {
-  const counted = new Set<Number>();
-  let time;
+  const counted = new Set<number>();
+  const [time, setTime] = useState<string>();
 
   useEffect(() => {
     if (player) {
       for (let i = 0; i < player.submissions.length; i += 1) {
-        if (player.submissions[i].submissionType === SubmissionType.Submit &&
-          player.submissions[i].numCorrect === player.submissions[i].numTestCases &&
+        if (player.submissions[i].numCorrect === player.submissions[i].numTestCases &&
           !counted.has(player.submissions[i].problemIndex)) {
           counted.add(player.submissions[i].problemIndex);
-          time = player.submissions[i].startTime;
+          setTime(player.submissions[i].startTime);
         }
       }
     }
   }, [player]);
 
   if (!time && player && player.submissions.length > 0) {
-    time = player.submissions[player.submissions.length - 1].startTime;
+    setTime(player.submissions[player.submissions.length - 1].startTime);
   }
 
   return time;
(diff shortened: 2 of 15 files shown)


alankbi (approved):
I took a brief skim through and I think it seems tentatively good for approval - will note that I'm not too familiar with the other changes in stagingNumProblems so would prob leave a more thorough review once we open up a PR for that branch. Also would prob need to fix any bugs that come up if they still exist

zpChris:
I just followed the same setup you did on the screencast and got the same error at the end, where the player did not move to first place despite getting the highest score. I can investigate more tomorrow to see when this error occurs.

zpChris:
Based on some further testing, it seems like this happens when there are three problems, and if there are two players both at 67% and one is behind the other, than if the one in second place solves the last problem they don't move to first. I'll look at the code to see why this might be.

frontend/src/views/Results.tsx
```diff
@@ -356,6 +356,7 @@ function GameResultsPage() {
           inviteContent={inviteContent()}
           loading={loading}
           isCurrentPlayer={players[1]?.user.userId === currentUser?.userId}
+          numProblems={game?.problems.length || 1}
```
zpChris, line 359: Why is the alternative `1`? Is that just a placeholder because something has to be there when `game?.problems.length` is `undefined`?

src/main/java/com/codejoust/main/dto/game/GameMapper.java
```diff
@@ -69,7 +69,7 @@ public static Game fromRoom(Room room) {
         for (User user : room.getUsers()) {
             Player player = PlayerMapper.playerFromUser(user);
             player.setColor(colorList.get(index));
-            player.setSolved(new Boolean[room.getNumProblems()]);
+            player.setSolved(new boolean[room.getNumProblems()]);
```
zpChris, line 72: What does this line do?

src/main/java/com/codejoust/main/dto/game/PlayerDto.java
```diff
@@ -19,6 +19,6 @@
     private String code;
     private CodeLanguage language;
     private List<SubmissionDto> submissions = new ArrayList<>();
-    private Boolean[] solved;
+    private boolean[] solved;
```
zpChris, line 22: What was your thinking for changing these all to primitives? I'm not really sure of the pros / cons here of using primitives vs. objects, thoughts @jasonz6688 and @alankbi?

frontend/src/components/results/PlayerResultsItem.tsx
```diff
@@ -82,15 +82,17 @@ type PlayerResultsCardProps = {
...
   onViewCode: () => void,
 };
 
 function PlayerResultsItem(props: PlayerResultsCardProps) {
   const {
-    player, place, isCurrentPlayer, color, gameStartTime, onViewCode,
+    player, place, isCurrentPlayer, color, gameStartTime, numProblems, onViewCode,
   } = props;
 
   const score = useGetScore(player);
```
zpChris, line 94: Since score and `numProblems` are used in conjunction to calculate one value, I feel like there'd be a simpler way of doing this such that `numProblems` doesn't have to be passed in to the `ResultsTable`, `Podium`, and `PlayerResultsItem`.

frontend/src/components/results/PlayerResultsItem.tsx
```diff
@@ -103,21 +105,19 @@ function PlayerResultsItem(props: PlayerResultsCardProps) {
...
 
-    // Calculate time from start of game till best submission
-    const startTime = new Date(gameStartTime).getTime();
-    const diffMilliseconds = new Date(bestSubmission.startTime).getTime() - startTime;
+    const currentTime = new Date().getTime();
+    const diffMilliseconds = currentTime - new Date(time).getTime();
     const diffMinutes = Math.floor(diffMilliseconds / (60 * 1000));
-
-    return ` ${diffMinutes} min`;
+    return `${diffMinutes}m ago`;
```
zpChris, line 120: I see this was changed from how long from the game it took to submit, to how long ago they submitted. I'm open to discussion on this - I feel like whereas the "ago" might be more intuitive in the short-term, it's not quite as useful and how long it took to find the correct submission may actually be a better metric. Thoughts @jasonz6688 and @alankbi?

zpChris (commented):
Great job @jasonz6688! 😃 

I have lots of comments on this PR, but talking with @alankbi, I'm wondering if we should merge this and then let him work on finishing this feature, given relative strengths on frontend vs. backend work and since Alan and I have more time nowadays. That's why I didn't leave an explicit Request changes or Approve - what do both of you think?

zpChris:
> Based on some further testing, it seems like this happens when there are three problems, and if there are two players both at 67% and one is behind the other, than if the one in second place solves the last problem they don't move to first. I'll look at the code to see why this might be.

Also, I didn't figure out why this was, but whoever works on this PR next may want to look into it, since it's definitely a recurring bug.

zpChris:
[We're in End Game Now, End Game for multiple problems PR](https://trello.com/c/aUa3vYzD/392-were-in-end-game-now-end-game-for-multiple-problems-pr)

alankbi:
> I have lots of comments on this PR, but talking with @alankbi, I'm wondering if we should merge this and then let him work on finishing this feature, given relative strengths on frontend vs. backend work and since Alan and I have more time nowadays. That's why I didn't leave an explicit Request changes or Approve - what do both of you think?

I think that might be helpful actually - especially since merge conflicts are a slight blocker atm and it would be helpful if the only branch that needs updates is the main `stagingNumProblems`

alankbi:
Just a heads up @jasonz6688, I'll probably push commits onto this branch to address @zpChris's comments, that way we can get it all merged in and wrap up this multiple-problem story hopefully in the next two weeks or so

frontend/src/components/results/PlayerResultsItem.tsx
```diff
@@ -103,21 +105,19 @@ function PlayerResultsItem(props: PlayerResultsCardProps) {
...
-    const percent = Math.round((score / 12321) * 100);
+    const percent = Math.round((score / numProblems) * 100);
     return `${percent}%`;
   };
 
   const getSubmissionTime = () => {
-    if (!bestSubmission) {
-      return 'N/A';
+    if (!time) {
+      return 'Never';
```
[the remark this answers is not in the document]
alankbi: I think for consistency this makes sense as well (matches the code column that has N/A), so might revert this back

frontend/src/util/Hook.tsx
```diff
@@ -37,37 +38,39 @@ const useGetScore = (player?: Player) => {
           counted.add(player.submissions[i].problemIndex);
         }
       }
+
+      setScore(counted.size);
```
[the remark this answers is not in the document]
alankbi: From the looks of it I think so

frontend/src/util/Hook.tsx
```diff
@@ -37,37 +38,39 @@ const useGetScore = (player?: Player) => {
...
+    
+  }, [player, setScore]);
 
   if (player == null || player.submissions.length === 0) {
     return null;
   }
-
-  return counted.size;
+  
+  return score;
```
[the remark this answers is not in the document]
alankbi: I'm not sure if this is the reasoning for the change, but I wonder if it was to fix a bug or at least to maintain the practice of using state variables with custom hooks? Might prefer to leave this in just cause with state variables I believe once the value changes then it's guaranteed to re-render, and might just be safer than using a calculated variable

frontend/src/util/Hook.tsx
```diff
@@ -37,37 +38,39 @@ const useGetScore = (player?: Player) => {
...
   useEffect(() => {
     if (player) {
       for (let i = 0; i < player.submissions.length; i += 1) {
-        if (player.submissions[i].submissionType === SubmissionType.Submit &&
-          player.submissions[i].numCorrect === player.submissions[i].numTestCases &&
+        if (player.submissions[i].numCorrect === player.submissions[i].numTestCases &&
           !counted.has(player.submissions[i].problemIndex)) {
           counted.add(player.submissions[i].problemIndex);
-          time = player.submissions[i].startTime;
+          setTime(player.submissions[i].startTime);
```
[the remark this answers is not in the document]
alankbi: I'm sort of confused - is there a difference between correct and best submissions? Idk how it plays into effect when there are multiple problems though, since from this code I get the impression that it finds the time of the latest correct submission of a problem that hasn't been solved yet (idk if this is actually true though esp. with a non-state set variable). Maybe @jasonz6688 can clarify after this is merged?

src/main/java/com/codejoust/main/dto/game/PlayerDto.java
alankbi: Not sure either - I guess it simplifies things by making them default false rather than null? With single boolean values ik it changes the getter name, but idk if that's relevant here when it's an array

frontend/src/components/results/PlayerResultsItem.tsx
alankbi: Yeah I think this should prob remain as how long it took to submit it - main reason being that on the results page (and in reports) it kinda loses meaning if the time keeps increasing after the game is already over.

frontend/src/components/results/PlayerResultsItem.tsx
alankbi: I can look into this after merging/pulling from main in the next series of PRs

alankbi:
Merging this into the staging branch, but @jasonz6688 feel free to continue the conversation here once you're back and we can address those in future PRs

frontend/src/util/Hook.tsx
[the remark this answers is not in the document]
jasonz6688: Yeah, so counted.size didn't quite work in my testing. I think it is because counted is not a useState variable so it doesn't update properly.

frontend/src/views/Game.tsx
```diff
@@ -396,8 +390,8 @@ function GamePage() {
         setLoading(false);
 
         // Set the 'test' submission type to correctly display result.
-        submissions[submissions.length - 1].submissionType = SubmissionType.Test;
-        setCurrentSubmission(getSubmission(currentProblemIndex, submissions));
+        res.submissionType = SubmissionType.Test;
+        setCurrentSubmission(res);
```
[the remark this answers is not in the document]
jasonz6688: Yeah, those two lines specifically set the console to display the run code result. I had deleted the code before because I changed to the other system of storing runCodes.

frontend/src/views/Results.tsx
jasonz6688: Yeah, I don't know when such a case would be, but VSCode read that as a error without the `|| 1`.

src/main/java/com/codejoust/main/dto/game/GameMapper.java
jasonz6688: This line sets every player to not having solved any of the problems.

src/main/java/com/codejoust/main/dto/game/PlayerDto.java
jasonz6688: Yeah, the primary reason was that initialization now makes an array of falses because otherwise, we would have to iterate through the array and put initialize a false boolean object.

frontend/src/util/Hook.tsx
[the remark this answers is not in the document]
jasonz6688: The code should find the time of the last new correct submission. I can explain it in call, but I'm pretty confident in what it does.

Best submission is just legacy code I left in there so that we can modify it for the end game code. But best submission theoretically finds the submission that solved the most test cases. Correct just means that a submission solved all the test cases for the problem.

PR #170, opened 2021-06-16
Connect users with accounts, create report data model.

zpChris:
### List of Changes:
- Connect users with accounts.
- Create the data objects in the backend for the report setup.
- Add tests for the new and updated backend methods.

### Notes:
- The overall GameReport model works as follows:
  - An Account will hold a list of GameReports (`ManyToMany` association because multiple Accounts can be associated with the same GameReport, as looking to the future Users may have Accounts).
  - Each GameReport holds a list of Problems and a list of Users. The list of Problems is fairly self-evident (note: some work may be done identifying problems that have been updated since that game), and the list of Users has a list of SubmissionGroups.
  - Each SubmissionGroup holds a `gameId` and a list of Submissions. This is necessary because each User is tied to a Room, not to an individual Game, so I had to break down the association and find only the Submissions for that particular game. The SubmissionGroup may hold more information for that User's performance in the Game - this is to be determined.
  - Each Submission holds the code, language, number of correct test cases, and more about the User's submission. (When @jasonz6688's multiple problems work gets merged, this will likely include a Problem or `problemIndex`.)
- The `User` object can now be seen more as a `UserSession` object, where a new session is generated for every room. This may change in the future, but it was a simple solution that doesn't require much refactoring with nicknames and whatnot that allows us to connect Accounts with GameReports.
- When constructing the relationships between objects, I initially created `OneToMany` associations without the paired `ManyToOne` as well, but ran into issues as it seems that wasn't allowed. There may be a way to get this to work (see [this SO post](https://stackoverflow.com/questions/2095998/onetomany-without-inverse-relationship-and-without-a-join-table), but I ended up just creating the associated `ManyToOne` pairings as well.
- On the testing side, I made the new `postRequestNoToken` and `putRequestNoToken` methods so that I could create a request with no token. This is also achievable by creating a new parameter on the original methods, but I initially avoided that since they are used basically everywhere and didn't want to make them more complicated. However, I'm open to adding that or another solution - lmk @alankbi.

### Screencast and Images:

[Screencast of Connect Users with Accounts](https://www.loom.com/share/c5dd748ca92a4db0820bef1831c1a014)

There are no images attached as no changes were made to the frontend .


zpChris:
[Connect users with accounts, create report data model](https://trello.com/c/CC74Y4Td/405-connect-users-with-accounts-create-report-data-model)

src/main/java/com/codejoust/main/model/User.java
```diff
@@ -37,8 +45,23 @@
...
+
     // This column holds the primary key of the room (not the roomId variable)
     @ManyToOne(fetch = FetchType.EAGER)
     @JoinColumn(name = "room_table_id")
     private Room room;
+
+    // This column holds the primary key of the game report
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "game_report_table_id")
+    private GameReport gameReport;
```
alankbi, line 61: Does each user only have up to one GameReport? I thought multiple game reports could belong to each room and user session

src/main/java/com/codejoust/main/model/User.java
```diff
@@ -37,8 +45,23 @@
...
+
+    // Thie list holds the submission group associated with the current room
+    @OneToMany(mappedBy = "user", fetch = FetchType.EAGER)
+    @Setter(AccessLevel.PRIVATE)
+    private List<SubmissionGroupReport> submissionGroupReports = new ArrayList<>();
```
alankbi, line 66: Is my thinking correct here (just wanting to confirm)? From understanding, when a GameReport is fetched, you'll have the list of users just fine. Then, to see what each player submitted, you would loop through this field and find the one index whose SubmissionGroupReport's gameId belongs to that GameReport?

src/main/java/com/codejoust/main/model/report/CodeLanguage.java
```diff
@@ -0,0 +1,32 @@
...
+import com.codejoust.main.exception.GameError;
+import com.codejoust.main.exception.api.ApiException;
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import lombok.Getter;
+
+@Getter
+public enum CodeLanguage {
+
+    PYTHON,
```
alankbi, line 12: Any reason for duplicating this file? We could move the other one into a higher up directory that way it's more suited for reuse

src/main/java/com/codejoust/main/model/User.java
alankbi: If so, maybe we could manually implement a getter method along the lines of getSubmissions (either here or in GameReport) that would filter through and find the correct report. 

Along a similar line, it could be useful when mapping the GameReport to a GameReportDto, to automatically do that (meaning GameReportDto's user DTO would simply hold a single SubmissionGroupReport)

src/main/java/com/codejoust/main/model/report/GameReport.java
```diff
@@ -0,0 +1,55 @@
...
+@Entity
+@Getter
+@Setter
+public class GameReport {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Integer id;
+
+    private String gameReportId;
```
alankbi, line 35: Should this be auto generated? Wondering if we're able to use a UUID here similar to in other entities

src/main/java/com/codejoust/main/model/report/GameReport.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @OneToMany(mappedBy = "gameReport", fetch = FetchType.EAGER)
+    @Setter(AccessLevel.PRIVATE)
+    @Fetch(value = FetchMode.SUBSELECT)
+    private List<Problem> problems = new ArrayList<>();
+
+    @OneToMany(mappedBy = "gameReport", fetch = FetchType.EAGER)
+    @Setter(AccessLevel.PRIVATE)
```
alankbi, line 43: Same question repeated here - should these users and problems fields be ManyToMany?

src/main/java/com/codejoust/main/model/report/SubmissionGroupReport.java
```diff
@@ -0,0 +1,42 @@
...
+    private String gameReportId;
+
+    @OneToMany(mappedBy = "submissionGroupReport", fetch = FetchType.EAGER)
+    @Setter(AccessLevel.PRIVATE)
+    private List<SubmissionReport> submissionReports = new ArrayList<>();
+
+    // This column holds the primary key of the user who made these submissions
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "user_table_id")
+    private User user;
```
alankbi, line 41: I _very briefly_ looked through that SO post so this may be way off base, but would it work to simply move the JoinColumn to the OneToMany side (as in the first answer)?

src/main/java/com/codejoust/main/model/report/SubmissionReport.java
```diff
@@ -0,0 +1,44 @@
...
+    private Integer numCorrect;
+
+    private Integer numTestCases;
+
+    private Double runtime;
+
+    // This column holds the primary key of the associated submission group
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "submission_group_table_id")
+    private SubmissionGroupReport submissionGroupReport;
```
alankbi, line 43: Same deal here - wondering if that SO solution would cause us to not have to have this here (the main reason why this could be very helpful is that way we don't need to worry about bidirectional relationships and all the bugs that would ensue (e.g. adding a SubmissionReport but not calling `setSubmissionGroupReport(this)` here)

src/main/java/com/codejoust/main/service/RoomService.java
```diff
@@ -90,6 +97,14 @@ public RoomDto joinRoom(String roomId, JoinRoomRequest request) {
             user.setUserId(utility.generateUniqueId(UserService.USER_ID_LENGTH, Utility.USER_ID_KEY));
         }
 
+        // Set the account associated with the host, if any.
+        if (token.length() == 0) {
```
alankbi, line 101: if you do end up setting required = false on those request headers, I think this (and createRoom/updateUserAccount) will need to be updated to check for the null condition as well

src/main/java/com/codejoust/main/model/User.java
```diff
@@ -37,8 +45,23 @@
      */
     private String sessionId;
 
+    // This column holds the primary key of the account
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "account_table_id")
+    private Account account;
```
alankbi, line 51: I left a few comments about the SO post and moving JoinColumn to the OneToMany side to get it to work, but then I just realized that this seems to work fine? There isn't an associated field in Account and there's no extra generated table in the database - lmk if I'm missing something

src/test/java/com/codejoust/main/util/MockHelper.java
```diff
@@ -49,6 +49,18 @@
         return UtilityTestMethods.toObject(jsonResponse, c);
     }
 
+    public static <T> T postRequestNoToken(MockMvc mockMvc, String url, Object body, Class<T> c, HttpStatus status) throws Exception {
+        MvcResult result = mockMvc.perform(post(url)
+                .header(HttpHeaders.AUTHORIZATION, "")
```
alankbi, line 54: if making these requeset headers optional, this line could be removed (and in RoomTestMethods)

alankbi (commented):
This looks awesome, very excited for this feature :) I haven't approved yet mainly cause of questions I had about the various database stuff, but should be good for approval after clarifying all that. 

Two more notes:

1. Similar to somehow identifying problems that have been updated, we might need some changes to handle deleted problems, since I think the current design stores the list of problems and a problemIndex for each submission, which would cause things to become out of sync if one is removed from that list

2. Eventually, it might be good to allow users to "delete" a report from their history, most likely just be disassociating their account from the User (that way the report still exists for other users, and if all accounts have been dissociated, can delete the report for real).

zpChris:
> 1. Similar to somehow identifying problems that have been updated, we might need some changes to handle deleted problems, since I think the current design stores the list of problems and a problemIndex for each submission, which would cause things to become out of sync if one is removed from that list

I was thinking about this exact same thing! I see two potential solutions right now, but I'm sure there are many more:

1. All references to the problems are replaced by `problemId` instead of `problemIndex`. Thus, if the `problemId` that is referenced does not exist in that list, then you know that that problem must have been deleted. You could then reflect this update on the frontend as well, showing a widget that a certain problem has been deleted by showing a reference to that problem.
  - The slight issue with this is that if there are no submissions for a problem that ends up deleted, you will not identify that there used to be another problem there that was deleted.
2. Replace the `List<Problem> problems` field within the game report with something like a `List<ProblemContainer>`, which contains the `Problem` as a field (as well as other information if necessary, like perhaps the number of test cases at the time the game was played). Thus, if the problem is deleted, the `ProblemContainer` is still present and we know that there was a deleted problem there. This also helps solve the issue of if problems were updated, meaning that their references are out-of-date - some of that crucial material (like number of test cases) could remain within the `ProblemContainer`.
  - This idea could also be used in conjunction with the one above, or you could maintain the `problemIndex` field - both should work fine.

> 2. Eventually, it might be good to allow users to "delete" a report from their history, most likely just be disassociating their account from the User (that way the report still exists for other users, and if all accounts have been dissociated, can delete the report for real).

I agree. I'd also say similar to conversations we've had in the past, it'd be helpful to be able to organize reports into a folder type structure, so teachers can group similar games or those from different classes together. That is all post-MVP in my mind, but I'd be open to implementing report deletion from history, especially because teachers may often just be testing new games. (We could even add settings after the game is finished to have teachers confirm they want that game in reports, or something similar.)

src/main/java/com/codejoust/main/dto/user/UserMapper.java
```diff
@@ -15,7 +16,18 @@ public static UserDto toDto(User entity) {
...
+        UserDto userDto = mapper.map(entity, UserDto.class);
+
+        // Attach the account to the user, if it exists.
+        if (entity.getAccount() != null) {
+            AccountUidDto accountUidDto = new AccountUidDto();
+            accountUidDto.setUid(entity.getAccount().getUid());
+            userDto.setAccountUid(accountUidDto);
+        } else {
+            userDto.setAccountUid(null);
+        }
```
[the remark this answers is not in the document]
zpChris: I was surprised too, but it is necessary - I'm not sure how the `mapper.map` algorithm works, but the screenshot of the debugging process below shows that the `accountUid` is not yet set.

**Debugging the toDto Method**
Debugging the toDto Method

src/main/java/com/codejoust/main/exception/RoomError.java
```diff
@@ -19,6 +19,7 @@
     USER_NOT_FOUND(HttpStatus.NOT_FOUND, "The specified user could not be found in the room."),
     NOT_FOUND(HttpStatus.NOT_FOUND, "A room could not be found with the given id."),
     DUPLICATE_USERNAME(HttpStatus.CONFLICT, "A user with the nickname provided has already joined the room."),
+    DUPLICATE_ACCOUNT(HttpStatus.CONFLICT, "A user with the account provided has already joined the room."),
```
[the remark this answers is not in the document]
zpChris: Done! ✅

src/main/java/com/codejoust/main/model/report/CodeLanguage.java
zpChris: That's a good point - I removed the duplicate file and moved the old one to the `main/model/report` so that it's placed in the database.

src/main/java/com/codejoust/main/model/report/GameReport.java
zpChris: Yes it should - I made this change in the next branch `GenerateReport`, but updated it here as well. 👍

src/main/java/com/codejoust/main/model/report/SubmissionGroupReport.java
zpChris: It turns out this works great - the reason it wasn't working originally is because I didn't include `@Fetch(value = FetchMode.SUBSELECT)` on the `OneToMany` side, so I mistakenly thought that this wasn't possible. I see that it works great now though, so I'll update the data model associations to remove the unnecessary `ManyToOne` annotations and fields.

src/main/java/com/codejoust/main/model/User.java
zpChris: So I realized both work just fine, but that the `OneToMany` associations need the `@Fetch(value = FetchMode.SUBSELECT)`annotation, whereas the `ManyToOne` association doesn't need that. That's why this works fine, but I thought that the `OneToMany` associations didn't work initially.

src/main/java/com/codejoust/main/model/problem/Problem.java
```diff
@@ -75,6 +77,11 @@
     @Fetch(value = FetchMode.SUBSELECT)
     private List<ProblemTag> problemTags = new ArrayList<>();
 
+    // This column holds the primary key of the game report
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "game_report_table_id")
+    private GameReport gameReport;
```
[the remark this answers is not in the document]
zpChris: Agreed - I mistakenly didn't think this was achievable, but found the issue I made in [this discussion](https://github.com/CodeJoustHQ/main/pull/170/files#r654115910). I'll remove this field! ✅

src/main/java/com/codejoust/main/model/report/SubmissionReport.java
zpChris: Agreed  removed! ✅

src/main/java/com/codejoust/main/model/report/GameReport.java
zpChris: You're right, they absolutely should. Once I update the `problems` to the `ProblemContainer` type though, that should have a `OneToMany` mapping because there is a new `ProblemContainer` for each new game.

zpChris:
Note for future development: The [Kahoot statistics](https://www.google.com/search?q=kahoot+statistics+quiz&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjS9cTnwaTxAhVKj54KHasYBbMQ_AUoAnoECAEQBA&biw=1278&bih=624#imgrc=jDM4-z9Kh-gNaM&imgdii=qLfeT8kWeGjGhM) are a very good example in my mind of how to extract useful information from each game. I like the idea of highlighting the users who need help, emphasizing the difficult problems, and more.

Ideally, too, I think having a leaderboard that spans across all problems works the best imo - but I'm open to discussion on that, since it's not a big problem.

src/main/java/com/codejoust/main/model/report/GameReport.java
zpChris: Yes this should! I actually did that in the [next branch](https://github.com/CodeJoustHQ/main/blob/GenerateReport/src/main/java/com/codejoust/main/model/report/GameReport.java), but I'll update it here as well. 👍

src/main/java/com/codejoust/main/controller/v1/RoomController.java
```diff
@@ -37,13 +38,13 @@ public RoomController(RoomService service) {
...
-        return new ResponseEntity<>(service.joinRoom(roomId, request), HttpStatus.OK);
+    public ResponseEntity<RoomDto> joinRoom(@PathVariable String roomId,@RequestBody JoinRoomRequest request, @RequestHeader(name="Authorization") String token) {
+        return new ResponseEntity<>(service.joinRoom(roomId, request, token), HttpStatus.OK);
     }
 
     @PostMapping("/rooms")
-    public ResponseEntity<RoomDto> createRoom(@RequestBody CreateRoomRequest request) {
-        return new ResponseEntity<>(service.createRoom(request), HttpStatus.CREATED);
+    public ResponseEntity<RoomDto> createRoom(@RequestBody CreateRoomRequest request, @RequestHeader(name="Authorization") String token) {
+        return new ResponseEntity<>(service.createRoom(request, token), HttpStatus.CREATED);
```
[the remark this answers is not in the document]
zpChris: That's a good idea, I didn't realize that allowed `token` to be simply shown as `null`. I've updated the corresponding service methods and tests as well. 👍

src/main/java/com/codejoust/main/model/User.java
zpChris: You're right, I made a mistake here - I've removed this field and updated the `List<User>` field within the `GameReport` class.

src/main/java/com/codejoust/main/model/report/CodeLanguage.java
zpChris: No reason, good point - I deleted this file and moved the original to the `main/model/report` folder.

src/main/java/com/codejoust/main/service/RoomService.java
zpChris: Done! ✅

src/test/java/com/codejoust/main/util/MockHelper.java
zpChris: Done ✅

src/main/java/com/codejoust/main/model/User.java
zpChris: I think all this is correct - the implementation and details / whatnot will be on the next PR, but if I understand correctly that's what I'm thinking too 👍

src/main/java/com/codejoust/main/dto/user/UserMapper.java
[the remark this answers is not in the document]
alankbi: I see - looking at it I think it's actually cause the account field is named accountUid instead of account in the UserDto so it can't map automatically. This isn't a major deal either way tho if you prefer to leave this as is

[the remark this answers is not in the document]
zpChris: Ahh ok - I'll try to make that change, as this would simplify things quite a bit 👍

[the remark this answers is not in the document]
zpChris: Done! ✅

src/main/java/com/codejoust/main/model/problem/ProblemContainer.java
```diff
@@ -0,0 +1,42 @@
...
+    private Integer userSolvedCount;
+
+    // The number of test cases.
+    private Integer testCaseCount;
+
+    // The average (mean) number of test cases passed.
+    private Double averageTestCasesPassed;
+
+    // The average (mean) number of submission attempts.
+    private Double averageAttemptCount;
```
zpChris, line 40: These values are set to be stored in the database; @alankbi, thoughts on that? We will want to store some things like `testCaseCount` because the number of `testCases` on the problem could change in the future.

The reason I stored the other statistics too was because I figured it was better to store the information than re-calculate it every time, as this is a relatively large number of calculations. However, I'm open to other arguments, especially if it is unorthodox to store statistic information like this in a database - let me know your thoughts!

src/main/java/com/codejoust/main/model/problem/ProblemContainer.java
```diff
@@ -0,0 +1,42 @@
...
+public class ProblemContainer {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Integer id;
+    
+    // The problem associated with this game report
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "problem_table_id")
+    private Problem problem;
```
alankbi, line 28: One thing to keep in mind for the future - if deleting a problem, will it automatically make this field be null? Cause IIRC there were some issues in the past with deleting problems because tags were still linked to them, so wondering if the same would be true here or if there's a setting to automatically remove those associations

alankbi (approved):
Looks good to me, thanks for making those changes/answering those questions! I left one additional question, but otherwise it seems ready to be merged

src/main/java/com/codejoust/main/model/problem/ProblemContainer.java
alankbi: I think that sounds good - based on our convo in standup a few days ago, I think that makes sense to avoid having to constantly recalcuate those

src/main/java/com/codejoust/main/model/problem/ProblemContainer.java
zpChris: This is a good point - I'm not sure, so this is definitely something I'll test before finalizing this feature. 👍

PR #172, opened 2021-06-18
Staging Num Problems

jasonz6688:
### List of Changes:

- Added slider to the lobby page to set the number of problems for a game
- Added numProblems value to the room, lobby, and updateSettingsRequest
- Frontend now accepts and stores multiple problems on the PlayerGameView.tsx page, PlayerGameView.tsx and Editor.tsx also now accept a 2D array for code and an array for languages, PlayerGameView.tsx page stores the languages and code as two arrays, whereas the Editor stores the code and languages as is.
- PlayerGameView.tsx stores currentProblemIndex, which determines which problem is being worked on, the code and language which are displayed and modified in the Editor depend on this value as well as the displayed description and console results/testcases.
- currentProblemIndex is added as a field to submission request to tell the backend for which problem the submission is made.
- currentProblemIndex can be incrememted or decremented by the small buttons added to PlayerGameView.tsx, it will wrap around from 0 and problems.length (a small number needs to be added somewhere to display the currentProblemIndex).
- Code are submissions are now stored in a array of submissions in PlayerGameView.tsx, which is used to reload code after refreshes.
- Removed code and language field in PlayerDTO and frontend player object
- Score is now calculated as the number of problems which you solve completely correctly.
- Game ends only after everyone solves every problem correctly.

### Notes:

Alan has patched all the bugs that we know of, but there may be more bugs, specifically I think that there might be weird edge cases with reloading the page and loading saved code. Also you can technically set the numProblems slider below the number of selected problems, but the site will set the error and not progress when you try to start the game; this works but differs from how we do errors for other setting requirements. Also, I think that we should reconsider the choose problem modal to show up before the user selects the number of problem, as the user will first select a problem, close the modal, select a number, and re-open the modal, and choose problems.

Additionally, a UI PR will probably be needed to make the buttons more in place (especially because the time limit modal appears over the next and previous buttons) and to fix the font issue (if Alan hasn't already fixed that). Additionally, additional PRs can deal with viewing the winning code. Another UI change that we may want to add later is some text that displays which problem is being worked on (i.e. 1/2, 2/4, ...).

### Screencast and Images:
Selecting multiple problems:
Shot1

Shot8

Shot9

Shot10

Selecting to many problems:
Shot11

Selecting problems then putting the slider down and pressing start game:
Shot12

Selecting one problem:
Shot2

Starting the game:
Shot3

Submitting correctly the first of three problems:
SHot4

Moving to the second problem:
Shot5

Submitting a partially correct (some test cases are wrong) the second problem:
Shot6

After submitting third problem correctly:
Shot7


frontend/src/components/game/Editor.tsx
```diff
@@ -8,9 +8,11 @@ import { DefaultCodeType } from '../../api/Problem';
 type EditorProps = {
   onLanguageChange: ((language: Language) => void) | null,
   onCodeChange: ((code: string) => void) | null,
-  codeMap: DefaultCodeType | null,
+  getCurrentLanguage: (() => Language) | null,
+  defaultCodeMap: DefaultCodeType[] | null,
   defaultLanguage: Language,
   defaultCode: string | null,
+  currentProblem: number,
   liveCode: string | null,
```
alankbi, line 16: Could we leave a multi-line comment above this editor props carefully detailing what each prop does? On initial glance it can be hard to remember what each of these things does

Also, why is a getCurrentLanguage,  defaultLanguage, and currentLanguage all needed? Could we just pass in the current language as a prop and use that to replace the state variable, or does that cause any bugs?

frontend/src/components/game/PlayerGameView.tsx
```diff
@@ -182,17 +189,52 @@ function PlayerGameView(props: PlayerGameViewProps) {
...
+  // Returns the most recent submission made for problem of index curr.
+  const getSubmission = (curr: number, playerSubmissions: Submission[]) => {
+    for (let i = playerSubmissions.length - 1; i >= 0; i -= 1) {
+      if (playerSubmissions[i].problemIndex === curr) {
+        return playerSubmissions[i];
+      }
+    }
+
+    return null;
+  };
```
alankbi, line 222: nit: maybe make this into a useCallback? Not sure if there's any major advantage (speed maybe) to doing that or not tho, but could be worth looking into the docs about that

alankbi: Actually, maybe this could even be moved into one of the utils files, as I can imagine this being useful on the results and other pages

frontend/src/components/game/PlayerGameView.tsx
```diff
@@ -357,22 +418,40 @@ function PlayerGameView(props: PlayerGameViewProps) {
...
         setError(err.message);
       });
   };
 
+  const nextProblem = () => {
+    setCurrentProblemIndex((currentProblemIndex + 1) % problems?.length);
+    setCurrentSubmission(getSubmission((currentProblemIndex + 1) % problems?.length, submissions));
+  };
+
+  const previousProblem = () => {
```
alankbi, line 435: Note: these functions are updated in #174 to no longer loop between the first and last problem

frontend/src/components/game/PlayerGameView.tsx
```diff
@@ -510,6 +596,10 @@ function PlayerGameView(props: PlayerGameViewProps) {
           }
         </SplitterLayout>
       </SplitterContainer>
+
+      <SmallButton onClick={previousProblem}>Previous</SmallButton>
+      <SmallButton onClick={nextProblem}>Next</SmallButton>
```
alankbi, line 601: Note: these are also updated in #174

frontend/src/components/results/PlayerResultsItem.tsx
```diff
@@ -1,12 +1,12 @@
...
+import { Player, Submission } from '../../api/Game';
 import { LowMarginText, Text } from '../core/Text';
 import { Color } from '../../api/Color';
-import { Player } from '../../api/Game';
-import { useBestSubmission } from '../../util/Hook';
+import { useBestSubmission, useGetScore, useGetSubmissionTime } from '../../util/Hook';
 import Language, { displayNameFromLanguage } from '../../api/Language';
 import { TextButton } from '../core/Button';
-import { getScore, getSubmissionCount, getSubmissionTime } from '../../util/Utility';
+// import { getScore, getSubmissionCount, getSubmissionTime } from '../../util/Utility';
```
alankbi, line 9: I think I was the one who commented this out and forgot to delete it lol, but could you remove this? Also, could you check if these are still be used (and if so, whether than can be replaced by the new hooks you made) and delete them if not?

alankbi: Actually, looking at the rest of this file, those three methods are duplicated in this component as well, so something to keep in mind - maybe that means the methods can be deleted from this file and you can reuse those util ones, or maybe both can be deleted, etc.

frontend/src/util/Hook.tsx
```diff
@@ -30,6 +30,52 @@ export const useBestSubmission = (player?: Player | null) => {
...
+      for (let i = 0; i < player.submissions.length; i += 1) {
+        if (player.submissions[i].numCorrect === player.submissions[i].numTestCases
+          && !counted.has(player.submissions[i].problemIndex)) {
+          counted.add(player.submissions[i].problemIndex);
+        }
+      }
+
+      setScore(counted.size);
+    }
+  }, [player, setScore, counted]);
```
alankbi, line 48: Just to double check, do sets count as "state" in some form? As in just making sure that the counted variable isn't like cleared/reset on every re-render (or something similar that could cause bugs) given that it's not part of core React

frontend/src/util/Hook.tsx
```diff
@@ -30,6 +30,52 @@ export const useBestSubmission = (player?: Player | null) => {
...
+        }
+      }
+    }
+  }, [player, counted]);
+
+  if (!time && player && player.submissions.length > 0) {
+    setTime(player.submissions[player.submissions.length - 1].startTime);
+  }
+
+  return time;
```
alankbi, line 76: What does this return if there's no submission/time? Is it possible to add a good default so that it doesn't mess things up in that edge case?

alankbi (commented):
Looks great, glad this is finally getting merged! Since I also worked on this branch a bit, I'll hold off on approving or requesting changes for now, in the meantime just left a few comments/questions

alankbi:
> Also, I think that we should reconsider the choose problem modal to show up before the user selects the number of problem, as the user will first select a problem, close the modal, select a number, and re-open the modal, and choose problems.

If I understand correctly, yeah this should be updated in #174 - basically numProblems will only be relevant if the user doesn't select a specific problem (if they do, then numProblems will automatically be incremented to match the count)

> Additionally, a UI PR will probably be needed to make the buttons more in place (especially because the time limit modal appears over the next and previous buttons) and to fix the font issue (if Alan hasn't already fixed that). Additionally, additional PRs can deal with viewing the winning code. Another UI change that we may want to add later is some text that displays which problem is being worked on (i.e. 1/2, 2/4, ...).

+1 on the UI changes (should be handled in the mentioned PR or if not, the one after that). The font issue I pushed a change to _hopefully_ address it, but since I'm not using Windows maybe @jasonz6688 you can run that branch locally and verify? 

The one thing I didn't address is the scrollbar design on Windows, which to me looks a bit offputting and am wondering if we should customize that at all. However, idk if that's a standard thing to see on Windows, so maybe it isn't a huge deal - thoughts @jasonz6688 @zpChris?

zpChris:
> +1 on the UI changes (should be handled in the mentioned PR or if not, the one after that). The font issue I pushed a change to _hopefully_ address it, but since I'm not using Windows maybe @jasonz6688 you can run that branch locally and verify?
> 
> The one thing I didn't address is the scrollbar design on Windows, which to me looks a bit offputting and am wondering if we should customize that at all. However, idk if that's a standard thing to see on Windows, so maybe it isn't a huge deal - thoughts @jasonz6688 @zpChris?

My thoughts could change on this in the future, but in terms of our current goals I don't think a scrollbar Windows design fix is an important enough issue to put on our current docket to address.

zpChris:
I found a small bug relating to multiple problems output for which I've made a [screencast](https://www.loom.com/share/8b83bcb864c84a879038173d221d7f16); when the problem is ran or submitted, and then before completing the load the problem is switched, the output and/or results will pop up under the wrong problem.

After going back and forth the problem results will correct themselves, but they are initially off and that's likely a bug we should address, imo.

frontend/src/util/Hook.tsx
jasonz6688: It returns null, and anywhere this hook is used, checks to see if it's null, and if so, uses never as the time. I think I used this system because it was in place - never was extrapolated from a null best submission. This can probably just be changed to return Never.

zpChris (approved):
LGTM! ✅ 

I left a good number of comments, but since @alankbi is working on #174 which will update this PR's work, I'm ok if some of those fixes or updates fall through to the next PR.

So excited to merge this in soon (long time coming) - let's gooo @jasonz6688!!! 🎉 🥳

src/main/java/com/codejoust/main/dto/game/SubmissionDto.java
```diff
@@ -19,6 +19,7 @@
 public class SubmissionDto {
     private CodeLanguage language;
     private String code;
+    private int problemIndex;
```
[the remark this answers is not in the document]
jasonz6688: In this case, I think I just defaulted to my habit. I can change this to an Integer, but I'm not sure why that's preferable.

frontend/src/components/results/PlayerResultsItem.tsx
```diff
@@ -129,13 +153,13 @@ function PlayerResultsItem(props: PlayerResultsCardProps) {
...
-        <Text>{getScore(bestSubmission)}</Text>
+        <Text>{getScore()}</Text>
       </td>
       <td>
-        <Text>{getSubmissionTime(bestSubmission, gameStartTime)}</Text>
+        <Text>{getSubmissionTime()}</Text>
       </td>
       <td>
-        <Text>{getSubmissionCount(player)}</Text>
+        <Text>{getSubmissionCount()}</Text>
```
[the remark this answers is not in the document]
jasonz6688: @alankbi, I think you made this change. Correct me if I'm wrong.

frontend/src/util/Hook.tsx
```diff
@@ -30,6 +30,52 @@ export const useBestSubmission = (player?: Player | null) => {
...
+    }
+  }, [player, setScore, counted]);
+
+  if (player == null || player.submissions.length === 0) {
+    return null;
+  }
+  return score;
+};
+
+export const useGetSubmissionTime = (player?: Player) => {
```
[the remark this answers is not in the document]
jasonz6688: Yeah, so the reason I didn't go backwards was because if the user submitted the same correct code twice, useGetSubmissionTIme would get the time of the second one. I believe we agreed that we want the first submission that solves the last solved problem.

frontend/src/util/Hook.tsx
jasonz6688: I'm not so sure what that means, but if set is cleared, the next time that the player object changes it will be refilled. We can talk about this in the call.

frontend/src/components/game/Editor.tsx
```diff
@@ -139,9 +149,31 @@ function ResizableMonacoEditor(props: EditorProps) {
 
   useEffect(() => {
     if (codeMap != null && codeEditor != null) {
-      codeEditor.setValue(codeMap[currentLanguage]);
+      if (codeMap[currentProblem] != null) {
+        const codeMapTemp = codeMap;
+
+        if (codeEditor.getValue() !== 'Loading...') {
+          codeMapTemp[previousProblem][currentLanguage] = codeEditor.getValue();
+        }
```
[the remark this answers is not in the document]
jasonz6688: This check is actually crucial, else Loading... will be saved as the code for the first problem.

frontend/src/components/game/PlayerGameView.tsx
```diff
@@ -135,6 +135,7 @@ type StateRefType = {
   currentUser: User | null,
   currentCode: string,
   currentLanguage: string,
+  currentIndex: number,
```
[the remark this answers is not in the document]
zpChris: You resolved this but didn't change it, or leave another comment - was there some thinking here or no?

frontend/src/components/game/Editor.tsx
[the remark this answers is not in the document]
zpChris: Ah you're right - thanks 👍

frontend/src/components/results/PlayerResultsItem.tsx
```diff
@@ -83,23 +83,47 @@ type PlayerResultsCardProps = {
...
+  const getSubmissionTime = () => {
+    if (!time) {
+      return 'N/A';
+    }
+
+    const startTime = new Date(gameStartTime).getTime();
+    const diffMilliseconds = new Date(time).getTime() - startTime;
+    const diffMinutes = Math.floor(diffMilliseconds / (60 * 1000));
+    return `${diffMinutes}m ago`;
+  };
```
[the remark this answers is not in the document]
zpChris: What changes did you make @jasonz6688? I couldn't find this function in the original method so I'm curious how you resolved it - adding comments to this discussion should help.

frontend/src/util/Hook.tsx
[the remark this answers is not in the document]
zpChris: You're right - I realized this later but forgot to remove this comment. Thanks for thinking of that! 💯

src/main/java/com/codejoust/main/dto/game/SubmissionDto.java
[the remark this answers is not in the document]
zpChris: We had the discussion on the call - this sounds good, I'll also try to do this going forward.

frontend/src/components/results/PlayerResultsItem.tsx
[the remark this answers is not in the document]
alankbi: I don't remember making this change, but looking through the codebase I think it's because these methods are newly added, as opposed to before when using the exported functions from Utility

frontend/src/components/game/PlayerGameView.tsx
```diff
@@ -182,17 +189,52 @@ function PlayerGameView(props: PlayerGameViewProps) {
...
+  };
+
+  const setOneCurrentCode = (newCode: string) => {
+    setCodeList(codeList.map((current, index) => {
+      if (index === currentProblemIndex) {
+        return newCode;
+      }
+      return current;
+    }));
+  };
```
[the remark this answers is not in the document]
jasonz6688: I don't remember writing that code, which makes me think it's one of the optimizations or bug fixes that Alan wrote. So @alankbi, would it be better to use the filter function here?

frontend/src/api/Game.ts
```diff
@@ -8,8 +8,6 @@ import { Color } from './Color';
 
 export type Player = {
   user: User,
-  code: string,
-  language: string,
```
alankbi, line 12: nit: change solved to be `boolean[]`

frontend/src/util/Hook.tsx
```diff
@@ -30,6 +30,68 @@ export const useBestSubmission = (player?: Player | null) => {
   return bestSubmission;
 };
 
+export const useGetScore = (player?: Player) => {
+  const counted = new Set<number>();
+  const [score, setScore] = useState<number>(0);
```
alankbi, line 35: If things are working on the backend, I actually don't think we need this at all - I think it can be replaced with something as simple as `player.solved.filter((s) => s)`

frontend/src/components/results/PlayerResultsItem.tsx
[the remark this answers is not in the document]
jasonz6688: Sorry, I totally forgot to explain this when resolving these comments. Alan commented on the exact same three methods, saying that they were redundant, and that I should delete them and refactor to use the equivalent methods in the hook.tsx and utility.tsx.

frontend/src/api/Game.ts
jasonz6688: Wait, why should allSolved be a boolean array? The backend still uses a Boolean to represent if all players have solved all problems.

frontend/src/util/Hook.tsx
jasonz6688: I'm not quite sure what you mean by that. We can talk about it in the standup.

frontend/src/components/game/PlayerGameView.tsx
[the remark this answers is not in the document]
zpChris: As we discussed last time, there's no need to change anything here, so I'll resolve this. 👍

frontend/src/components/results/PlayerResultsItem.tsx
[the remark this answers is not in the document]
zpChris: Resolving this because we discussed it on the call.

frontend/src/util/Hook.tsx
alankbi: After talking about it, I'll implement this change in my upcoming PR 👍

zpChris:
@jasonz6688 Is this PR ready to merge?

jasonz6688:
[Staging Num Problems, finally](https://trello.com/c/PHV83RKb/411-staging-num-problems-finally)

jasonz6688:
[Staging Num Problems, finally](https://trello.com/c/PHV83RKb/411-staging-num-problems-finally)

PR #175, opened 2021-06-27
Fix link bugs and add style improvements.

zpChris:
### List of Changes:
- Update the "Create a problem" link on the "Verified Problems" page to the correct url: `/problem/create`.
- Update problem to only be set initially once.
- Update Back button to return to previous page only if one exists on the same site; otherwise, go to the dashboard.
- Restrict the problem description line length to one line.

### Notes:
- The problem must be set only initially once because every time it is set, a new id is given for each test case; this means that the `JSON.stringify` check on the Back button will sometimes return that the problem was updated, even when it was not.
- I found the Back button "POP" check at [this Stack Overflow post](https://stackoverflow.com/questions/37385570/how-to-know-if-react-router-can-go-back-to-display-back-button-in-react-app).
- An explanation / demonstration of the line length clamp is shown at [this screencast](https://www.loom.com/share/e942f9872be449a8a6bc158eac99bfb3) (albeit on another page, but still relevant).

### Screencast and Images:

[Screencast of the Fix Link Bugs and Update Site Design](https://www.loom.com/share/f6ff3d943fef4953bd04927af5ee6630)

**Update Description Length on Verified Problems**
_This image just shows one small portion of the PR._
Update Description Length on Verified Problems

### DIFF
--- a/frontend/src/components/card/ProblemCard.tsx
+++ b/frontend/src/components/card/ProblemCard.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import { Problem } from '../../api/Problem';
-import { LargeText, SelectedItemText, Text } from '../core/Text';
+import { LargeText, SelectedItemText, SingleLineText } from '../core/Text';
 import { getDifficultyDisplayButton } from '../core/Button';
 import { SelectedItemContainer } from '../core/Container';
 import { DivLink } from '../core/Link';
@@ -43,14 +43,6 @@ const TitleText = styled(LargeText)`
   vertical-align: middle;
 `;
 
-const DescriptionText = styled(Text)`
-  display: -webkit-box;
-  -webkit-line-clamp: 1;
-  overflow: hidden;
-  -webkit-box-orient: vertical;
-  word-break: break-all;
-`;
-
 function ProblemCard(props: ProblemCardProps) {
   const { problem } = props;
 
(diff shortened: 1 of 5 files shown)


zpChris:
[Fix link bugs and add style improvements](https://trello.com/c/tv7YLGOY/409-fix-link-bugs-and-add-style-improvements)

frontend/src/components/card/ProblemCard.tsx
```diff
@@ -43,6 +43,14 @@ const TitleText = styled(LargeText)`
...
 `;
 
+const DescriptionText = styled(Text)`
+  display: -webkit-box;
+  -webkit-line-clamp: 1;
+  overflow: hidden;
+  -webkit-box-orient: vertical;
+  word-break: break-all;
+`;
+
```
alankbi, line 53: nit: maybe rename and move this to the Text file so it can be reused? Could potentially have the font size as inherit so that it automatically adjusts to whatever wraps it - problem with this is that I don't think you're allowed/supposed to nest p tags (maybe changing this to span will help with that, but idk, if that doesn't work this should be fine as is)

alankbi (approved):
Looks good to me! Left one comment about potentially reusing the line camp component (nice find btw 🙂), but otherwise should be good to merge whenever ready

alankbi:
Also, lmk if you planned on setting the min-width on this or another PR and I can take a look at that as well (it's a small change though so prob would be fine with just merging that without additional review)

frontend/src/components/card/ProblemCard.tsx
zpChris: I added a `SingleLineText` component to the `Text.tsx` file for this - I saw you already implemented your own implementation in #174, so whether you use this or not is fine by me. 👍

zpChris:
> Also, lmk if you planned on setting the min-width on this or another PR and I can take a look at that as well (it's a small change though so prob would be fine with just merging that without additional review)

I'll likely do this on another PR with a list of other small changes 👍

PR #176, opened 2021-06-30
Generate game report after the game ends.

zpChris:
### List of Changes:

Add bullet points here.

### Notes:

Add additional bullet points here.

### Screencast and Images:

Add your images and/or screencasts here.


alankbi:
Looks like there are some commit issues similar to what was on #174 so I'll hold off on reviewing for now - however, after #172 was merged those commit issues when away for #174 so I'd imagine this will get resolved pretty quickly (as soon as #174 is merged)

zpChris:
> Looks like there are some commit issues similar to what was on #174 so I'll hold off on reviewing for now - however, after #172 was merged those commit issues went away for #174 so I'd imagine this will get resolved pretty quickly (as soon as #174 is merged)

I'll resolve all the merge conflicts at once after #174 is merged, since I just approved that and #178.

alankbi:
Looks like the commits are fixed, but maybe not the line numbers 😮 hopefully a merge will fix that tho, in which case will try to review this quickly (sorry for the delay)

src/main/java/com/codejoust/main/service/GameManagementService.java
```diff
@@ -283,14 +307,131 @@ public GameDto manuallyEndGame(String roomId, EndGameRequest request) {
...
         for (Timer timer : gameTimer.getNotificationTimers()) {
             timer.cancel();
         }
+
+        CreateGameReportTask createGameReportTask = new CreateGameReportTask(this, game);
+        Timer createGameReportTimer = new Timer();
+        createGameReportTimer.schedule(createGameReportTask, GameTimer.DURATION_1 * 1000);
+    }
+
+    public GameReport createGameReport(Game game) {
```
alankbi, line 324: nit: move this + other report logic into a new ReportService

src/main/java/com/codejoust/main/service/GameManagementService.java
```diff
@@ -283,14 +307,131 @@ public GameDto manuallyEndGame(String roomId, EndGameRequest request) {
...
+        GameReport gameReport = new GameReport();
+        int numProblems = game.getProblems().size();
+        int numPlayers = game.getPlayers().size();
+
+        // Initialize the statistic variables for each problem.
+        int[] userSolved = new int[numProblems];
```
alankbi, line 330: In order to make this more readable, maybe split it into separate methods? Something like `generateGameStatistics`, `generateProblemStats`, `populateAccounts`, etc. that way it's more clear what each section is doing

src/main/java/com/codejoust/main/service/GameManagementService.java
```diff
@@ -283,14 +307,131 @@ public GameDto manuallyEndGame(String roomId, EndGameRequest request) {
...
         GameTimer gameTimer = game.getGameTimer();
         gameTimer.getTimer().cancel();
 
         for (Timer timer : gameTimer.getNotificationTimers()) {
             timer.cancel();
         }
+
+        CreateGameReportTask createGameReportTask = new CreateGameReportTask(this, game);
+        Timer createGameReportTimer = new Timer();
+        createGameReportTimer.schedule(createGameReportTask, GameTimer.DURATION_1 * 1000);
```
alankbi, line 321: I'm sort of conflicted on this - I feel like timers always add extra complexity which could be a hassle down the line and also if teachers ever wanted to view their reports immediately after, it might be annoying to wait 1+ minutes or confusing if we don't have a banner informing them. 

I'm wondering if it makes sense to cut off submissions that go through after the deadline, that way we can immediately generate this in `handleEndGame`? The main downside I think is that students who submit with only a few seconds left might lose that submission, but imo that's an edge case and a risk they knew they were taking (and currently we also don't have a good way to show results that came in post-game over other than just recalculating the leaderboard, which might be confusing). Thoughts?

src/main/java/com/codejoust/main/service/GameManagementService.java
```diff
@@ -283,14 +307,131 @@ public GameDto manuallyEndGame(String roomId, EndGameRequest request) {
...
+
+        // Set problem container variables.
+        int numTestCases = 0;
+        List<Problem> problems = game.getProblems();
+        for (int i = 0; i < numProblems; i++) {
+            Problem problem = problems.get(i);
+            numTestCases += problem.getTestCases().size();
+
+            ProblemContainer problemContainer = new ProblemContainer();
+            problemContainer.setProblem(problem);
```
alankbi, line 382: What will happen if a problem is deleted in the middle of a game? I believe the problem will still exist in the game object since it's in the backend, but what will happen here?

alankbi (commented):
From a quick look, this looks great! Just leaving some preliminary comments since the bulk of changes seem to be in the GameManagementService file, and that way once you fix the merge conflicts/lines changed hopefully you won't be completely blocked waiting on a review

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: Sounds good - done! ✅

zpChris: When I created the new `ReportService` class, I also created the new `ReportServiceTests` class; in the original tests, I also used a `Spy` for `GameManagementService` not just for the `createGameReport` method, but also to add submissions and add the game from the room. Is it ok that I have `Spy` objects for both `GameManagementService` and `ReportService` within that file, or should I find some workaround?

(I ask because this change makes the `ReportServiceTests` dependent not only on `ReportService`, but also on the `GameManagementService` - so it's not entirely isolated, but the work to get there may be large.) Thoughts @alankbi?

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: I'm cool with that actually, just removing the `CreateGameReportTask` class and tests entirely and making that change mentioned. I also agree because I feel it'd be great to directly add a link to the game report - this won't happen instantaneously, but it could at least be calculated in a few seconds and it'd be easy to mock a "loading" symbol until that time. (In this case, I may also add a socket update to send once the game report is created, so we can display that update.)

I'll plan on doing that by only adding the submission if the game is not over - that way, the submission may be calculated, but if it doesn't complete before the game finishes, then it is not added and the game report would not change.

I will note, there's also the option in the future of letting the teacher choose. We may not want to bombard them with too many options, but if it becomes a bit issue that submissions are getting cut off from the game end, then this functionality could vary either way based on a room setting. For now though, I'm ok with this new implementation.

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: This is a good point! You're right - I believe the problem will still exist in the game object, but it will not exist in the database. The following error thus exists.

**Problem Deleted During Game**
Problem Deleted During Game

I therefore want to make the change to check if the problem does not exist in the database, then I want to set the problem within the `ProblemContainer` as null. I'm adding this, and I'm also going to do this right before I save the game report.

There is still a chance, I believe, that while looping through the problems, one of the problems is deleted after it is already visited. I'm not sure what to do about conflicts like this - I think that relates to the issue we've had elsewhere of simultaneous SQL updates.

zpChris: I will note that after I implemented this change, it worked, but still produced the below error in the Terminal; I'm not sure whether this is something that I should be concerned about or not.

**Terminal Problem Not Found Error**
Terminal Problem Not Found Error

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: I also added a boolean field to the `Game` object titled `createGameReportStarted` - this is true once the process of creating a game report for that specific game has begun. This means that if multiple of the end game conditions occurred at very close times, only one of the game reports should be saved on the database.

This is more reliable than checking if a game report with the game ID already exists in the database as this saving process could take a while, and this lag may end up accidentally saving two games. By doing the check locally, one of the end game conditions should set this field to true, preventing the other one from also creating a game report as that field's value is updated very quickly.

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: I split it up into `computeGameStatisticsAndAddSubmissions`, setGameReportStatistics, `createProblemContainers, and `addGameReportAccounts`. The first method is clearly complicated (given the `And` in the name), but it fits fairly naturally because in order to calculate those game statistics, I have to iterate through each user's submissions.

Let me know if there's another system that you think would be better.

alankbi (approved):
Looks good! Mainly just left a few questions here and there. Also, another potential suggestion would be to add a new README under `model/report` diagraming/explaining how all the report model classes fit together (since it can be confusing), but not a huge deal

src/main/java/com/codejoust/main/service/GameManagementService.java
alankbi: Does this cause any issues? Never tried that before, but I guess if it works then it hopefully shouldn't be an issue

src/main/java/com/codejoust/main/service/GameManagementService.java
alankbi: Does this error still happen? If so, I wonder if it has to do with one of the timer tasks not canceling properly? If it doesn't affect anything then should hopefully be fine, assuming that it only shows up in the edge case of deleting a problem

src/main/java/com/codejoust/main/service/GameManagementService.java
zpChris: It doesn't cause any issues, just makes the code styling arguably a bit worse. In the interest of keeping progress moving given how long this PR has been out, I'll merge this change in though, if that's alright. 👍

src/main/java/com/codejoust/main/model/Account.java
```diff
@@ -52,12 +52,16 @@
...
     @Setter(AccessLevel.PRIVATE)
     @JoinColumn(name = "game_report_id")
     @Fetch(value = FetchMode.SUBSELECT)
     private List<GameReport> gameReports = new ArrayList<>();
 
     @Enumerated(EnumType.STRING)
     private AccountRole role = AccountRole.TEACHER;
+
+    public void addGameReport(GameReport gameReport) {
+        gameReports.add(0, gameReport);
```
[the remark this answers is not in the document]
zpChris: That's a good question - it might not make a difference, I haven't seen a different ordering system pop up under Sequel Pro. If that's the case, I may revert this to `gameReports.add(gameReport);` in the future for simplicity, but I'll check this when I make the relevant frontend PR.

src/main/java/com/codejoust/main/model/User.java
```diff
@@ -58,9 +59,13 @@
...
     // Thie list holds the submission group associated with the current room
-    @OneToMany(fetch = FetchType.EAGER)
+    @OneToMany(fetch = FetchType.EAGER, cascade = CascadeType.ALL)
     @Setter(AccessLevel.PRIVATE)
     @Fetch(value = FetchMode.SUBSELECT)
     @JoinColumn(name = "submission_group_reports_table_id")
     private List<SubmissionGroupReport> submissionGroupReports = new ArrayList<>();
+
+    public void addSubmissionGroupReport(SubmissionGroupReport submissionGroupReport) {
+        submissionGroupReports.add(submissionGroupReport);
```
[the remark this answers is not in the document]
zpChris: That's right - this is not included in the standard `UserDto` class, nor will I add it there. I don't yet have a `GameReportDto` class as I've been working in the backend thus far, but I will add that at some point - I'll make another DTO class for the `User` object specifically for the report purposes, so it is only passed to the frontend when necessary.

src/main/java/com/codejoust/main/service/ProblemService.java
```diff
@@ -243,6 +253,25 @@ public ProblemDto deleteProblem(String problemId, String token) {
...
+        /**
+         * Before the problem can be deleted, set all foreign key references
+         * to the Problem within all ProblemContainers to null.
+         * See https://stackoverflow.com/a/10030873/7517518.
+         */
+        List<ProblemContainer> problemContainers = problemContainerRepository.findAllByProblem(problem);
+        for (ProblemContainer problemContainer : problemContainers) {
+            problemContainer.setProblem(null);
+            problemContainerRepository.save(problemContainer);
+        }
```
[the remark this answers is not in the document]
zpChris: You're right, this can take a very long time - I agree, some method to optimize this would be very useful. This is especially a problem because after clicking the "Delete Problem" button, the page does not redirect until the REST call returns. I think we could address this issue in a future PR, though I'll leave a few ideas here:
1. Add a `boolean deletionInProgress` field that allows us to show a notice that the problem is currently being deleted on the frontend.
2. Redirect immediately after the call is made rather than waiting for the return.
3. Have the function return success if it parses through everything, even if it has not finished deleting the problem - this may require having the for loop above continue in a separate thread or something, which I'm fairly certain is possible though I don't know how that would really work.
4. Somehow speed up the actual deletion process.
5. Remove the `problem` relation and replace it with a `problemId`, then simply replace those with `null` whenever a reference is attempted and not found.

PR #184, opened 2023-02-17
Create the local database and tester service

zpChris:
### List of Changes:

- Update the reference URLs to build CodeJoust completely locally - Tester service and database included.
- Address linter warnings by removing unused imports and adding null checks.
  - The null checks were superfluous in these cases, so I added comments for additional clarification.
- Remove the deployment pipeline on `workflow.yml`.
  - This can be added back with relevant adjustments once this Tester repository is re-deployed.
- Fix [log4j](https://github.com/CodeJoustHQ/main/security/dependabot/80) and [gson](https://github.com/CodeJoustHQ/main/security/dependabot/65) warnings.
- Address issues on tests consistently failing - they should hopefully now always pass, whether run on Github Actions, Maven, or VSCode.

### Notes:

- I fixed the imports on testing files to `import org.junit.jupiter.api.Test;` as they were not being detected by Maven.
- Upgrading the lombok dependency version to remove `IllegalAccessError`, as referenced in [this Stack Overflow](https://stackoverflow.com/a/66981165/7517518) post.
- Note for running with environment variables on new Macbook Air setup: environment variables are stored in `~/.zshenv`, and you can run `source ~/.zshenv` for help.
- I also fixed Java runtime warnings on my local machine via [this process](https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#java.configuration.runtimes) which configures JDK runtimes.

### Screencast and Images:

None.

### DIFF
--- a/src/main/java/com/codejoust/main/Main.java
+++ b/src/main/java/com/codejoust/main/Main.java
@@ -29,6 +29,8 @@ public class Main {
 		if (keyfile == null) {
 			log.error("Could not find the firebase keyfile in an environment variable.");
 			log.error("Please set this value before launching the program.");
+
+			// Exit the program with unsuccessful status; return used to satisfy linter warning.
 			System.exit(1);
 			return;
 		}
--- a/src/main/java/com/codejoust/main/Main.java
+++ b/src/main/java/com/codejoust/main/Main.java
@@ -30,6 +30,7 @@ public class Main {
 			log.error("Could not find the firebase keyfile in an environment variable.");
 			log.error("Please set this value before launching the program.");
 			System.exit(1);
+			return;
 		}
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/codejoust/main/Main.java
```diff
@@ -30,6 +30,7 @@ public static void main(String[] args) {
 			log.error("Could not find the firebase keyfile in an environment variable.");
 			log.error("Please set this value before launching the program.");
 			System.exit(1);
+			return;
```
zpChris, line 33: Adding this `return` statement, which may be superfluous but removed the `Potential null pointer access: The variable keyfile may be null at this location` warning from `java.compile.nullAnalysis.mode`.

